PR #9, opened 2017-08-25
Core library implementation with new API

acetechnologist:


stats/aggregation.go
```diff
@@ -0,0 +1,69 @@
+// Copyright 2017 Google Inc.
```
bogdandrutu: We changed the Copyright. 
// Copyright 2017, OpenCensus Authors

PR #11, opened 2017-08-25
Add initial stats API.

acetechnologist:


README.md
```diff
@@ -8,4 +8,249 @@ are in the process of being changed and will certainly break your code.
...
+                                      .Build()
+    ctx2 := tags.ContextWithChanges(ctx, newTagSet)
+
+## Registering views and retrieving their collected data.
+
+### To create/register a measure a.k.a resource
+Create/load measures units:
+
+    // returns a *MeasureFloat64
+    mf := stats.NewMeasureFloat64("/my/float64/measureName", "some measure")
```
AlekseySanin: Should we have error check here?

README.md
```diff
@@ -8,4 +8,249 @@ are in the process of being changed and will certainly break your code.
...
+    mi := stats.NewMeasureInt64("/my/otherName", "some other measure")
+    ...
+
+Register measures:
+
+	if err := stats.RegisterMeasure(mf); err != nil {
```
AlekseySanin: I don't understand the difference between New and Register. Should the measures be similar to tags and just have Create methods?

README.md
```diff
@@ -8,4 +8,249 @@ are in the process of being changed and will certainly break your code.
...
+    precisionSubsets := 10
+    wnd2 := stats.NewWindowSlidingCount(lastNSamples, precisionSubsets)
+    
+    wn3 := stats.NewWindowCumulative()
+
+
+### To creater/register a view
+Create a view:
+
+    vw1 = stats.NewViewInt64("my/int64/viewName", "some description", []Keys{key1, key2}, mf, agg1, wnd1)
```
AlekseySanin: I would strongly prefer some kind of builder pattern here. We have too many parameters for this call.

stats/aggregation.go
```diff
@@ -0,0 +1,69 @@
...
+	// [bounds[i-1], bounds[i]) for 0 < i < len(Bounds)
+	// [bounds[i-1], +infinity) for i = len(Bounds)
+	//
+	// if len(Bounds) == 0 then there is no histogram associated with the
+	// distribution. There will be a single bucket with boundaries
+	// (-infinity, +infinity).
+	//
+	// if len(Bounds) == 1 then there is no finite buckets, and that single
+	// element is the common boundary of the overflow and underflow buckets.
+	bounds []float64
```
AlekseySanin: We should make sure this semantic is applied across all languages

stats/aggregation_value.go
```diff
@@ -0,0 +1,284 @@
...
+// AggregationValue is the interface for all types of aggregations values.
+type AggregationValue interface {
+	isAggregate() bool
+	addSample(v interface{})
+	multiplyByFraction(fraction float64) AggregationValue
+	addToIt(other AggregationValue)
+	clear()
+}
+
+func aggregationValueAreEqual(av1, av2 AggregationValue) bool {
```
AlekseySanin: Should we add equal() method to the aggregationvalue interface instead?

stats/collector.go
```diff
@@ -0,0 +1,83 @@
...
+	// window is the window under which the aggregation is performed.
+	w Window
+}
+
+func (c *collector) addSample(s string, v interface{}, now time.Time) {
+	aggregator, ok := c.signatures[s]
+	if !ok {
+		var newAggregationValue func() AggregationValue
+		switch agg := c.a.(type) {
+		case *AggregationCount:
```
AlekseySanin: Might be better to find a way to use inheritance here

stats/aggregation.go
acetechnologist: That is part of the spec that aveitch had at some point. I think everybody agreed on.

README.md
acetechnologist: Originally I had what you are suggesting for both measures and views. But then you suggested to separate create/new from register for views in one of your comments in "Census Stats API v0.1" :)
After thinking about it, it seems separating the 2 actions makes more sense as it makes it easier to understand what is going on. Anyone can create any view/measure they want. But registering it the library is what might return an error.
I think keeping measure/views interactions consistent makes it easier to get our head around it.

acetechnologist: That being said I do not have a strong opinion about it. It just seems better to be consistent.

README.md
acetechnologist: I don't see the benefits. 
The construction logic of a view is straightforward. Why try to encapsulate it?

The types of the arguments are all different (besides the name/description) and they are all required (besides the description). It seems we'll catch more errors this way at compile time vs using a builder pattern where users can forget to set arguments and failures won't occur until runtime.

stats/aggregation_value.go
acetechnologist: yes

stats/collector.go
acetechnologist: There is a way. Done.
Now, this code looks more concise and cleaner.

README.md
acetechnologist: The error is to be checked during registration.

README.md
AlekseySanin: It feels like tags and measures are the same -- kind of constants used by other things. I don't see a use case when Measure is looked up by name (same as tags). I think views are different.

README.md
AlekseySanin: I just don't like constructors with >3 arguments but I see your point.

stats/aggregation.go
AlekseySanin: yes I am just noting it.

README.md
acetechnologist: I modified measures to be automatically "registered" on creation

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+                                    .UpsertString( key2, "bar value")
+                                    .Build()
+
+
+
+To create a new tagsSet from an existing tag set oldTagSet:
+    oldTagSet := ...
+    tsb := &tags.TagSetBuilder{}
+    tsb.StartFromTagSet(oldTagSet)
+    tsb.InsertString( key1, "foo value")
```
rakyll: tsb.InsertString(key1, "foo value")

gofmt these snippets please.

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+    tsb := &tags.TagSetBuilder{}
+    tsb..StartFromEmpty()
+    tsb.InsertString( key1, "foo value")
+    tsb.UpdateString( key1, "foo value2")
+    tsb.UpsertString( key2, "bar value")
+    tagsSet := tsb.Build()
+
+    // A shorter way of achieving the same is:
+    tagsSet := &tags.TagSetBuilder{}.StartFromEmpty()
+                                    .InsertString( key1, "foo value")
```
rakyll: This snippet won't compile. In Go, try doing:

```
tagsSet := &tags.TagSetBuilder{}.StartFromEmpty().
                                     InsertString(key1, "foo value").
                                     UpdateString( key1, "foo value2").
etc
```

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+        // handle error
+    }
+    vw2, err := stats.GetViewByName("my/float64/viewName")
+    if err != nil {
+        // handle error
+    }
+    ...
+    
+
+UnRegister view:
```
rakyll: Unregister

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+    if err := stats.RegisterView(vw1); err != nil {
+        // handle error
+    }
+    if err := stats.RegisterView(vw2); err != nil {
+        // handle error
+    }
+
+Retrieve view by name:
+    vw1, err := stats.GetViewByName("my/int64/viewName")
+	if err != nil {
```
rakyll: Either user tabs or spaces, it is a mix here.

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+
+    if err := stats.RegisterView(vw1); err != nil {
+        // handle error
+    }
+    if err := stats.RegisterView(vw2); err != nil {
+        // handle error
+    }
+
+Retrieve view by name:
+    vw1, err := stats.GetViewByName("my/int64/viewName")
```
rakyll: I'd name this with a more representative name, e.g. countView := ...

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+        // handle error
+    }
+
+    c2 := make(c chan *ViewData)
+    if err := stats.SubscribeToView(vw2, c2); err != nil {
+        // handle error
+    }
+    ... 
+
+Unubscribe from a view:
```
rakyll: Unsubscribe

stats/aggregation.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// Package stats defines the stats collection API and its native Go
+// implementation.
```
rakyll: This is a Go package, you don't have to say it is a Go implementation. It is redundant.

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+    if err, key2 := tags.CreateKeyString("keyNameID2"); err != nil {
+        // handle error
+    }
+    ...
+
+### Create a set of tags associated with keys
+To create a new tag set from scratch using changes:
+
+    tsb := &tags.TagSetBuilder{}
+    tsb..StartFromEmpty()
```
rakyll: tsb.StartFromEmpty()

StartFromX pattern is a very uncommon pattern in Go. Users expect either to thing work with a zero value or a constructor is present.

Couldn't this package just export:
```
func NewFromTagSet(ts *TagSet) *TagSetBuilder
func (ts *TagSet) Empty() *TagSetBuilder
```

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+                                      .UpdateString( key1, "foo value2")
+                                      .UpsertString( key2, "bar value")
+                                      .Build()
+
+### Add new tagSet to a context / Modify tagSet in a context 
+Add tags to a context for propagation to downstream methods and downstream rpcs:
+To create a new context with the tags. This will create a new context where all the existing tags in the current context are deleted and replaced with the tags passed as argument.    
+    
+    newTagSet  := ...
+    ctx2 := tags.ContextWithNewTags(ctx, newTagSet)
```
rakyll: ContextWithNewTags sounds like a accessor. What about NewContext(context.Context, *TagSet) instead to self document it will derive from the given context and return a new context.

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+
+Create a new context keeping the old tag set and adding new tags to it, removing specific tags from it, or modifying the values fo some tags.
+    
+    oldTagSet := tags.FromContext(ctx)
+    newTagSet := &tags.TagSetBuilder{}.StartFromTagSet(oldTagSet)
+                                      .InsertString( key1, "foo value")
+                                      .UpdateString( key1, "foo value2")
+                                      .UpsertString( key2, "bar value")
+                                      .Build()
+    ctx2 := tags.ContextWithChanges(ctx, newTagSet)
```
rakyll: Same here.

README.md
acetechnologist: Replaced it with:
func EmptyTagSetBuilder() TagSetBuilder
func NewTagSetBuilderFromTagSet(ts *TagSet) TagSetBuilder

where TagSetBuilder is an interface now.

.travis.yml
```diff
@@ -1,18 +1,22 @@
 language: go
 
 go:
-  - 1.6.x
-  - 1.7.x
   - 1.8.x
 
-go_import_path: github.com/census-instrumentation/instrumentation-go
+go_import_path: github.com/census-instrumentation/opencensus-go
```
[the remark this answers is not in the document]
acetechnologist: No I am not. Removed.

README.md
acetechnologist: Done

stats/aggregation.go
acetechnologist: Done

README.md
acetechnologist: Done

README.md
acetechnologist: Done

README.md
acetechnologist: Just called it myView

README.md
acetechnologist: This is to mimic the standard: context.WithValue(parent Context, key, val interface{}). But we need to hide the key.
Do you still feel "NewContext(context.Context, *TagSet)" is a better choice?

README.md
acetechnologist: Same as previous comment.

README.md
acetechnologist: Done

stats/aggregation.go
acetechnologist: ack

stats/aggregation.go
```diff
@@ -0,0 +1,78 @@
...
+	// distribution. There will be a single bucket with boundaries
+	// (-infinity, +infinity).
+	//
+	// if len(Bounds) == 1 then there is no finite buckets, and that single
+	// element is the common boundary of the overflow and underflow buckets.
+	bounds []float64
+}
+
+// NewAggregationDistribution creates a new aggregation of type distribution
+// a.k.a histogram.
```
hyangah: More details about how bounds (boundaries) are used will be appreciated. (some of the comments included in the AggregationDistribution.bounds comment which is not displayed on godoc)

stats/worker.go
```diff
@@ -0,0 +1,375 @@
...
+		err: make(chan error),
+	}
+	defaultWorker.c <- req
+	return <-req.err
+}
+
+// UnsubscribeFromView unsubscribes a previously subscribed channel from the
+// View subscriptions. If no more subscriber for v exists and the the ad hoc
+// collection for this view isn't active, data stops being collected for this
+// view.
```
hyangah: Alternative approach is what google cloud pubsub api uses - it creates a new subscriber that invokes a user provided callback for each received message. And, for unsubscription, it utilizes the context cancellation. I think that allows users to avoid boilerplate for channel creation, loop around channel, etc. and also the API has flexibility to determine what kind of channel is appropriate. What do you think?

stats/worker.go
```diff
@@ -0,0 +1,375 @@
...
+		c:   c,
+		err: make(chan error),
+	}
+	defaultWorker.c <- req
+	return <-req.err
+}
+
+// StartCollectionForAdhoc starts data collection for this view even if no
+// listeners are subscribed to it.
+func StartCollectionForAdhoc(v View) error {
```
hyangah: Is there a StartCollection for non-adhoc? Why not just StartCollection? Or, ForceCollection.

stats/worker.go
acetechnologist: The issue with that is that if the callback takes too long to run then our worker thread will get stuck and starve the library from executing work.
I guess alternatively we could create a new go routine every time we report the usage but it seems seems wasteful.
I rather give more control to the user and let them decide how they want to process the returned value.

stats/worker.go
acetechnologist: ForceCollection and StopForcedCollection sounds good to me. Done.

stats/aggregation.go
acetechnologist: Done

README.md
```diff
@@ -8,4 +8,245 @@ are in the process of being changed and will certainly break your code.
...
+        // handle error
+    }
+    mi, err := stats.NewMeasureInt64("/my/otherName", "some other measure")
+    if err != nil {
+        // handle error
+    }    
+    ...
+
+Retrieve measure by name:
+    mf, err := stats.GetMeasureByName("my/float64/measureName")
```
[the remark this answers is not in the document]
acetechnologist: Done.

README.md
bogdandrutu: I completely agree that we don't literally need an EmptyTagSetBuilder. 

In regard to the Builder pattern I think we would like to have TagSet immutable to simplify and optimize the synchronization schema, and then another possible option would be to have Mutation and have a NewTagSet(Mutation ...) where a mutation is one of the 3.

README.md
```diff
@@ -8,4 +8,240 @@ are in the process of being changed and will certainly break your code.
...
+    precisionSubsets := 10
+    wnd2 := stats.NewWindowSlidingCount(lastNSamples, precisionSubsets)
+    
+    wn3 := stats.NewWindowCumulative()
+
+
+### To creater/register a view
+Create a view:
+
+    myView1 = stats.NewViewInt64("/my/int64/viewName", "some description", []Keys{key1, key2}, mf, agg1, wnd1)
```
bogdandrutu: As we discussed we actually need only one function NewView and extract the type from the measure.

bogdandrutu: Also it would be great to always do type switch on the measure.

stats/aggregation.go
acetechnologist: I would but it doesn't exist yet.

README.md
acetechnologist: Understood. But for now I prefer the safety of compile time error.

README.md
acetechnologist: Replaced it with a single NewTagSetBuilder(ts *TagSet) method.

To hyangah@ tte builder pattern is for efficiency purposes. Once the TagSet is built we want it to be immutable. To avoid making copies when adding 10 or 100 KeyVals I think a builder pattern is appropriate here.

README.md
acetechnologist: Fair enough. Done.

README.md
acetechnologist: None. That was an error in the documentation. Fixed now

README.md
acetechnologist: Done

README.md
acetechnologist: Eliminated the need for ViewInt64 and ViewFloat64. Now there is only View.

PR #12, opened 2017-09-15
gRPC integration

acetechnologist:
Added grpc handlers.
Added sumOfSquaredDeviation to aggregation distribution (to mimic stackdriver) and simplified its implementation to always include the oldest bucket.


plugins/grpc/handlers.go
```diff
@@ -0,0 +1,75 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package stats
+
+import (
+	"context"
```
menghanl: gRPC-go still imports `"golang.org/x/net/context"` (https://github.com/grpc/grpc-go/issues/711).
It will be a massive incompatible API change for all users using <=go1.8 if we change that.

go treats `func("context".Context)` and `func("golang.org/x/net/context".Context)` as two different types, which means the stats handler here won't work (it doesn't compile, with <=go1.8).

go1.9 introduced a new feature called type alias, which makes `"golang.org/x/net/context"` and `"context"` the same type. So, this actually works fine with >=go1.9.

The solution I can think of right now is to import `"golang.org/x/net/context"` in the stats package.

Or, if we decide that the stats package only supports users with >=go1.9, the current code will be fine.

plugins/grpc/stats/client_handler.go
```diff
@@ -0,0 +1,176 @@
...
+
+	ts := tags.FromContext(ctx)
+	encoded := tags.EncodeToFullSignature(ts)
+
+	// Join old metadata with new one, so the old metadata information is not lost.
+	md, _ := metadata.FromOutgoingContext(ctx)
+	ctx = metadata.NewOutgoingContext(ctx, metadata.Join(md, metadata.Pairs(tagsKey, string(encoded))))
+
+	tsb := tags.NewTagSetBuilder(ts)
+	tsb.UpsertString(keyService, serviceName)
```
menghanl: Why does this function return a `TagSetBuilder` if it's called to update the builder?

plugins/grpc/handlers.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2017, OpenCensus Authors
```
menghanl: Some files are `Copyright 2017 Google Inc.`...

plugins/grpc/stats/server_handler.go
```diff
@@ -0,0 +1,215 @@
...
+	"strings"
+	"sync/atomic"
+	"time"
+
+	istats "github.com/census-instrumentation/opencensus-go/stats"
+	"github.com/census-instrumentation/opencensus-go/tags"
+	"github.com/golang/glog"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/stats"
+	/*
```
menghanl: Remove this

plugins/grpc/stats/types.go
```diff
@@ -0,0 +1,98 @@
...
+	"time"
+
+	istats "github.com/census-instrumentation/opencensus-go/stats"
+	"github.com/census-instrumentation/opencensus-go/tags"
+)
+
+// tagsKey is the metadata key used to identify both the stats tags in the GRPC
+// context metadata, as well as the RpcServerStats info sent back from the
+// server to the client in the GRPC context metadata.
+const tagsKey = "grpc-tags-bin"
```
menghanl: Metadata keys start with `grpc-` are actually reserved, so we should avoid having a `grpc-` prefixed key explicitly in the code...

To make this work, we added functions to set/get trace info: https://github.com/grpc/grpc-go/blob/master/stats/stats.go#L269.

Please use the new functions instead.

stats/aggregation_value.go
```diff
@@ -96,14 +112,15 @@ type AggregationDistributionValue struct {
...
+func NewTestingAggregationDistributionValue(bounds []float64, countPerBucket []int64, count int64, min, max, mean, sumOfSquaredDeviation float64) *AggregationDistributionValue {
 	return &AggregationDistributionValue{
 		countPerBucket: countPerBucket,
 		bounds:         bounds,
 		count:          count,
 		min:            min,
 		max:            max,
-		sum:            sum,
+		mean:           mean,
+		sumOfSquaredDeviation: sumOfSquaredDeviation,
```
menghanl: This is not formatted correctly.

stats/aggregation_value.go
```diff
@@ -188,69 +224,21 @@ func (a *AggregationDistributionValue) incrementBucketCount(f float64) {
...
+	// for simplicity, we include the oldest partial bucket when the aggregation
+	// is a distribution. We do not try to have a partial fraction.
 
-	if ret.count == 1 {
-		ret.min = (a.min + a.max) / 2
-		ret.max = ret.min
-		ret.sum = ret.min
-		ret.incrementBucketCount(ret.min)
-		return ret
+	ret := newAggregationDistributionValue(a.bounds)
```
menghanl: Did you miss anything? Parameter `fraction` is not used in this function.

plugins/grpc/stats/client_handler.go
acetechnologist: To make it easier to use the shortcut version when modifying multiple tags:
tags.NewTagSetBuilder(ts).
        UpsertString(keyService, serviceName).
        UpsertString(key2, value2).
        AddString(key3, value3)...

plugins/grpc/stats/client_handler.go
```diff
@@ -0,0 +1,176 @@
...
+	// connection related data into the context.
+	grpcClientConnKey *grpcInstrumentationKey
+	// grpcClientRPCKey is the key used to store client instrumentation RPC
+	// related data into the context.
+	grpcClientRPCKey *grpcInstrumentationKey
+)
+
+// ClientHandler is the type implementing the "google.golang.org/grpc/stats.Handler"
+// interface to process lifecycle events from the GRPC client.
+type ClientHandler struct{}
```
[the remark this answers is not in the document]
acetechnologist: Not sure I understand your comment. You mean adding an exported "constructor" for it?

plugins/grpc/handlers.go
acetechnologist: Since this is our first release, I think we should avoid creating unnecessary confusion and say that we support >= go1.9 if using stats with grpc.
Wondering if @rakyll have a different opinion.

plugins/grpc/handlers.go
acetechnologist: Good catch. Thank you. Fixed now.

plugins/grpc/stats/server_handler.go
acetechnologist: Done

stats/aggregation_value.go
acetechnologist: Done

stats/aggregation_value.go
acetechnologist: This method is part of the interface signature: "type AggregationValue interface".
We decided that for this specific aggregation we will not use the "fraction" for simplicity

plugins/grpc/stats/client_handler.go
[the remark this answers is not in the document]
menghanl: Yes..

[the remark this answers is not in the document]
acetechnologist: Done

README.md
```diff
@@ -20,7 +20,8 @@ To install this package, you need to install Go and setup your Go workspace on y
 $ go get -u github.com/census-instrumentation/opencensus-go
 
 ## Prerequisites
-This requires Go 1.8 or later as it uses the convenience function sort.Slice(...) in few places.
+This requires Go 1.8 or later as it uses the convenience function sort.Slice(...) introduced in Go 1.8.
+It also uses "golang.org/x/net/context".Context introduced in Go 1.7 heavily. Therefore, in order to integrate with any library still using "golang.org/x/net/context".Context, Go 1.9 or higher is required to take advantage of type aliasing.
```
menghanl: Do we want to keep this line?

plugins/grpc/stats/client_handler.go
```diff
@@ -0,0 +1,178 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package stats
+
+import (
+	"golang.org/x/net/context"
```
menghanl: [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports) would complain about the order of the imported packages.

We do those checks on travis: https://github.com/menghanl/grpc-go/blob/master/vet.sh#L51-L54

README.md
acetechnologist: No we don't. Deleted it.

plugins/grpc/stats/client_handler.go
acetechnologist: Ok. Fixed.

PR #14, opened 2017-09-27
Do not repeat package godoc.

rakyll:
Otherwise godoc will render multiple repetitions of the same package documentation.


rakyll:
You can see how it is rendered right now at https://godoc.org/github.com/census-instrumentation/opencensus-go/stats.

PR #25, opened 2017-10-03
Use simpler names for binary encoder/decoder

### LINKED ISSUE #23: Rename DecodeFromFullSignature and EncodeToFullSignature
Both function names are mouthful, Go libraries often just use Encode to encode a concrete type to []byte, and Decode to unmarshal from a []byte.

/cc @acetechnologist

rakyll:
Fixes #23.

### DIFF
--- a/tags/tag_set_codec.go
+++ b/tags/tag_set_codec.go
@@ -188,7 +188,7 @@ func Decode(bytes []byte) (*TagSet, error) {
 
 	version := eg.readByte()
 	if version > tagsVersionID {
-		return nil, fmt.Errorf("not supported version: %q; supports only up to: %q", version, tagsVersionID)
+		return nil, fmt.Errorf("unsupported version: %q; supports only up to: %q", version, tagsVersionID)
 	}
 
 	for !eg.readEnded() {
--- a/tags/tag_set_codec.go
+++ b/tags/tag_set_codec.go
@@ -198,7 +198,7 @@ func Decode(bytes []byte) (*TagSet, error) {
 		case keyTypeString:
 			break
 		default:
-			return nil, fmt.Errorf("key type invalid: %q", typ)
+			return nil, fmt.Errorf("invalid key type: %q", typ)
 		}
 
 		k, err := eg.readBytesWithVarintLen()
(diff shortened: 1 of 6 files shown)


tags/tag_set_codec.go
```diff
@@ -159,8 +159,8 @@ func (eg *encoderGRPC) bytes() []byte {
 	return eg.buf[:eg.writeIdx]
 }
 
-// EncodeToFullSignature will encode the tagSet to []byte.
-func EncodeToFullSignature(ts *TagSet) []byte {
+// Encode will encode the tag set into a []byte.
```
hyangah: Let's make the comment helpful (we know it takes a tag set and returns a []byte already); something like "Encode the tag set into the grpc wire format(?)."

BTW, I don't know what this tags package has to do with grpc.

tags/tag_set_codec.go
```diff
@@ -188,7 +188,7 @@ func DecodeFromFullSignature(bytes []byte) (*TagSet, error) {
 
 	version := eg.readByte()
 	if version > tagsVersionID {
-		return nil, fmt.Errorf("DecodeFromFullSignature doesn't support version %v. Supports only up to: %v", version, tagsVersionID)
+		return nil, fmt.Errorf("not supported version: %q; supports only up to: %q", version, tagsVersionID)
```
hyangah: s/not supported/unsupported

how about 'unsupported tag version'

tags/tag_set_codec.go
rakyll: I didn't want to give specific examples because this is a generic binary encoder/decoder. gRPC ends up using the binary encoding is an implementation detail for the gRPC propagation strategy.

tags/tag_set_codec.go
rakyll: Done

Yea, makes so much more sense :)

rakyll:
PTAL

tags/tag_set_codec.go
hyangah: What I meant is the comment in the current form doesn't add any value and should explain more about what the output is for. I also wanted to highlight that the appearance of grpc in this package doesn't look right (and it seems that succeeded. :-))

rakyll: Done

Explained further that it is useful to propagate the tag sets on write in binary format.

rakyll:
See https://github.com/census-instrumentation/opencensus-go/pull/29 instead.

PR #38, opened 2017-10-10
Rename GetMeasureByName and GetViewByName, simpler error messages

### LINKED ISSUE #31: Rename GetMeasureByName to MeasureByName
Accessors don't need a Get prefix and without the prefix it is consistent with tags.KeyStringByName.

rakyll:
Change also includes minor styling improvements and better error outputs from tests.

Fixes #31.

### DIFF
--- a/stats/worker_commands.go
+++ b/stats/worker_commands.go
@@ -39,16 +39,10 @@ type getMeasureByNameResp struct {
 
 func (cmd *getMeasureByNameReq) handleCommand(w *worker) {
 	if m, ok := w.measuresByName[cmd.name]; ok {
-		cmd.c <- &getMeasureByNameResp{
-			m,
-			nil,
-		}
+		cmd.c <- &getMeasureByNameResp{m, nil}
 		return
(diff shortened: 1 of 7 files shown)


stats/worker_commands.go
```diff
@@ -102,15 +96,12 @@ type getViewByNameResp struct {
 
 func (cmd *getViewByNameReq) handleCommand(w *worker) {
 	if v, ok := w.viewsByName[cmd.name]; ok {
-		cmd.c <- &getViewByNameResp{
-			v,
-			nil,
-		}
+		cmd.c <- &getViewByNameResp{v, nil}
```
bogdandrutu: Can we enforce golang format for the repository? Looks like a lot of this kind of changes. Maybe file an issue.

rakyll: The packages are already gofmt'ed and we have a CI rule to enforce it before merge. gofmt'ed is not opinionated on everything. Multiple lines are permitted but hard to read.

PR #39, opened 2017-10-10
Simplify the tags API

rakyll:
- Remove the builder, use mutators when constructing a TagSet.
- Rename KeyString to StringKey to prefer the natural language order.
- Stop exporting the the key IDs that are internal implementation details.
- Rename ValueAsString to StringValue which is a more common naming pattern in Go.
- Rename Tag.K to Tag.Key, Tag.Value to Tag.Value.


rakyll:
@bogdandrutu, you can review the new API by reading `example_test.go`.

tags/key.go
```diff
@@ -15,49 +15,94 @@
 
 package tags
 
-import "fmt"
-
-var keys []Key
+// Modifier functions modifies a tag set.
+type Modifier func(t *TagSet) *TagSet
```
zombiezen: What about:

```go
type Modifer struct {
  f func(t *TagSet) *TagSet
}
```

I'd like to prevent a user from write their own function to modify a `TagSet`, since it seems like `TagSet`s are intended to be immutable now.

Bikeshedding: I think I've seen the term "mutator" used to describe this kind of thing.  Doesn't really matter much to me, but I thought I'd mention.

tags/tag_set.go
```diff
@@ -43,17 +41,19 @@ type TagSet struct {
 	m map[Key][]byte
```
zombiezen: (on line 39 but GH won't let me add a comment since it isn't in the diff)

Remove `TagSetBuilder` from doc comment

tags/key.go
rakyll: Done

tags/tag_set.go
rakyll: Done

rakyll:
PTAL

tags/key.go
zombiezen: (See also my justification against interfaces with unexported methods over on #41.)

rakyll:
Not a fan of interfaces with unexported methods as well. I had tons of offline debates over this in the past weeks. Exported Mutate, will also improve the overall situation in these packages.

tags/tag_set.go
```diff
@@ -37,22 +36,24 @@ func (e ErrKeyNotFound) Error() string {
 	return fmt.Sprintf("key %q not found", e.Key)
 }
 
-// TagSet contains a set of tags. Use TagSetBuilder to build tag sets.
+// TagSet contains a set of tags. Use NewTagSet to build tag sets.
 type TagSet struct {
 	m map[Key][]byte
```
zombiezen: Is `TagSet` intended to be immutable?  If so, exporting Mutator seems to work in opposition to that.  If not, then this type seems simpler to describe as just `type TagSet map[Key][]byte`, much like `url.Values`.

rakyll: We want users not to care about encoding to []byte. Mutators are working against the types they know. UpdateInt64 will take a string and update the tag set by encoding the int64 into []byte. I'd prefer to hide the low-level representation from the user as much as possible.

tags/key.go
```diff
@@ -15,49 +15,112 @@
...
-
-var keys []Key
+// Mutator modifies a tag set.
+type Mutator interface {
+	Mutate(t *TagSet) *TagSet
+}
 
 // Key represents a tag key.
 type Key interface {
+	// Name returns the name of th key.
```
zombiezen: nit: "the"

tags/key.go
```diff
@@ -15,49 +15,112 @@
...
 
-import "fmt"
-
-var keys []Key
+// Mutator modifies a tag set.
+type Mutator interface {
+	Mutate(t *TagSet) *TagSet
+}
 
 // Key represents a tag key.
```
zombiezen: Worth documenting that a Key supports `==`.

tags/tag_set.go
zombiezen: Ah, so if you are trying to abstract the marshaling, what about this:

```go
type Set map[Key]Value  // since package is tags, this is tags.Set

func (set Set) Copy() Set

type Value struct {
  b []byte
}

func StringValue(s string) Value {
  return Value{[]byte(s)}
}

func Int64Value(i int64) Value {
  return Value{[]byte{/* encode the int */}}
}
```

Seems like it would be simpler to provide functions to copy the map and write the modifications in plain Go than creating an operation per operation x data type pair.  WDYT?

tags/key.go
rakyll: Done

tags/key.go
rakyll: I haven't seen this documented before, take a look at the new godoc. Does that look good?

rakyll:
> Ah, so if you are trying to abstract the marshaling, what about this:

We will remove the typed keys and there will likely to be only string keys. Let's not spend too much time on this right now, this API needs to be revisited entirely as soon as we finalize the data model.

I will send the s/TagSet/Set change in a different change.

rakyll:
Rebased, PTAL.

PR #41, opened 2017-10-11
Revise Measure and Measurement API

rakyll:
- Simplify the Measurement interface.
- Rename MeasureByName to FindMeasure, measures are always retrieved by name.
- Export recorder functions on measure.
- Rename (*MeasureX).Is to (*MeasureX.M).
- Add TODOs, make some style improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,11 +82,11 @@ if err != nil {
 Retrieve measure by name:
 
 ```go
-mf, err := stats.GetMeasureByName("/my/float64/measureName")
+mf, err := stats.FindMeasure("/my/float64/measureName")
 if err != nil {
     // handle error
 }
-mi, err := stats.GetMeasureByName("/my/otherName")
+mi, err := stats.FindMeasure("/my/otherName")
 if err != nil {
     // handle error
 }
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -230,15 +230,14 @@ if err := stats.StopForcedCollection(myView1); err != nil {
...
 ```go
-// mi is a *MeasureInt64 and v is an int64 .
-stats.RecordInt64(ctx, mi, v)
-// mf is a *RecordFloat64 and v is an float64 .
-stats.RecordFloat64(ctx, mf, v)
-// multiple measurements can be performed at once.
-stats.Record(ctx, mi.Is(4), mf.Is(10.5))
+mi.Record(ctx, 1)
+mf.Record(ctx, 5.6)
+stats.Record(ctx, mi.M(4), mf.M(10.5))
```
zombiezen: What are the use cases for the batch versus the single record modes here?  The sample doesn't illuminate why the user would want to use one or the other.

stats/measure.go
```diff
@@ -24,8 +24,32 @@ type Measure interface {
 	viewsCount() int
 }
 
-// Measurement is the interface for all measurement types. Measurements are
-// required when recording stats.
+// Measurement is the numeric value measured when recording stats. Each measure
+// provides methods to create measurements of their kind. For example, MeasureInt64
+// provides M to convert an int64 into a measurement.
 type Measurement interface {
```
zombiezen: So I've been kicking around an idea on how to displace interfaces with unexported types with a better alternative (since they can have confusing interactions with embedding and have the same API evolution issues as interfaces).  What if you defined Measurement in this way:

```go
type Measurement struct {
  m interface{} // one of *Float64Measure or *Int64Measure
  v interface{} // one of int64 or float64
}
```

Largely the same API surface for the application, but then it's more clear that only certain functions can create measurements.

rakyll: There will be more measurement types, I would revise this idea once we have all the types implemented.

README.md
rakyll: There are tons of cases where you want to record a set of measurements at certain points. For example, a response is returned from an outgoing RPC. The size, whether the RPC has been successful, latency, and similar measurements are all made available at the same time. Batch is useful for those cases, but I agree that we need to reconsider the APIs. I am going to add a TODO for it for now.

rakyll:
PTAL

README.md
zombiezen: SGTM.  My recommendation is (when the API is stable) to pick an example that illustrates when to use each.

PR #43, opened 2017-10-11
Convert View into a concrete type

rakyll:
There will be a single View implementation, hence
View is converted into a concrete type. This model also
is compatible with the Java client.

Renamed GetViewByName given views can only be retrieved
by name and the function performs a look-up to find
a view in the registry.

Moved some view related functions under View to be
methods.

Added some more godoc and made a few style improvements.

### DIFF
--- a/examples/stats/main.go
+++ b/examples/stats/main.go
@@ -94,13 +94,13 @@ func main() {
 		}
 	}(ch)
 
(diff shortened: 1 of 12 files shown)


rakyll:
@zombiezen, let's take a look at these changes together later today.

PR #44, opened 2017-10-12
Import context from the standard library where possible

rakyll:
Once https://github.com/grpc/grpc-go/pull/1492 is merged, we will entirely remove the dependency to golang.org/x/net.

### DIFF
--- a/examples/stats/main.go
+++ b/examples/stats/main.go
@@ -19,13 +19,13 @@
 package main
 
 import (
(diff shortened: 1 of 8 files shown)


dfawley:
Will rolling back grpc/grpc-go#1492 impact you?  AppEngine is still not actually on 1.8 yet, and they are requesting gRPC to support 1.6 for another X months.

PR #51, opened 2017-10-17
Rename TagSet and StringValue

### LINKED ISSUE #21: Remove ValueAsString?
```
func (k *KeyString) ValueAsString(b []byte) string
```
is not accessing any of the Key fields and is doing a redundant job of casting a []byte into string.

/cc @acetechnologist

### LINKED ISSUE #47: Rename TagSet to Set
tags.TagSet is repeating the "tag" twice and we don't like to stuttering in Go. Rename it to Set.

rakyll:
- Rename TagSet to Tag.
- StringValue is an encoder, it is better to have it in its name.
- Remove ErrKeyNotFound but rather return a bool from
  (*TagSet).StringValue. There are no other errors that can
  be returned other than existence/nonexistence.

Fixes #47.
Fixes #21.


tags/key.go
```diff
@@ -26,8 +26,8 @@ type Key interface {
 	// Name returns the name of the key.
 	Name() string
 
-	// StringValue encodes the given value represented in binary to string.
-	StringValue(b []byte) string
+	// EncodeValue encodes the given value represented in binary to string.
+	EncodeValue(b []byte) string
```
zombiezen: Doesn't this decode a value?

rakyll: This encodes the internal low-level representation into something else. For example, we may convert a float64 represented as a little endian byte slice into a hex encoded string.

README.md
```diff
@@ -30,7 +30,7 @@ OpenCensus libraries require Go 1.8 or later as it uses the convenience function
 
 Tags represent propagated key values. They can propagated using context.Context
 in the same process or can be encoded to be transmitted on wire and decoded back
-to a TagSet at the destination.
+to a Set at the destination.
```
bogdandrutu: I don't like this name for multiple reasons:
* It may be a Go thingy to shorter names but just Set does not make any sense (too generic).
* Consistency with other languages for the main classes names (this is probably one of the most important in Tags).
* Is not that bad that we duplicate 3 characters.

plugins/grpc/stats/client_handler.go
```diff
@@ -168,7 +168,7 @@ func (ch clientHandler) handleRPCEnd(ctx context.Context, s *stats.End) {
 	if s.Error != nil {
 		errorCode := s.Error.Error()
 		ts := tags.FromContext(ctx)
-		newTs := tags.NewTagSet(ts,
+		newTs := tags.NewSet(ts,
```
bogdandrutu: This for me (no Go person at all) does not make sense too much, I prefer the previous version.

tags/key.go
bogdandrutu: I am not sure I understand this API. BytesToString? Based on your example we will have a EncodeString(long). Also is better to read the name as a string EncodeByteToXXX or something similar.

README.md
rakyll: The reason for this change is not the shortness. The call sites will always include the package name to refer to this type as `tags.Set`. It is clear that it is a tag set from the full name. We prefer not to repeat when naming, tags.TagSet doesn't make the tag set more recognizable.

plugins/grpc/stats/client_handler.go
rakyll: Read my other argument.

tags/key.go
rakyll: We will always use EncodeValue. The int64 key will implement a EncodeValue that converts the []byte into string. I cannot follow your comment here.

BytesToString is a Java conventions. Go users look for encoders/decoders terminology.

rakyll:
@bogdandrutu  I responded to your last comment as well but it doesn't show on the UI, switch to the side-by-side view to read it.

README.md
jcd2: Maybe call it "tags.Set" in these top-level docs and it should read fine.

tags/key.go
jcd2: ISTM this method is used to take a value stored as a []byte in a Tag, and pretty-print it.  It's not used for serialization/deserialization.  Just call it "StringFromValue" or "ValueToString" or similar, and document that it's for pretty-printing.

Similarly, the documentation for tags.Set.StringValue says the return value is "encoded as a string" -- that could probably be made clearer since that's not used for serialization either.

rakyll:
@jcd2 All done.

An additional thought, we don't pluralize package names in Go and they serve as a scope/prefix. tag.Map sounds better than tags.Map. I will send another follow up change.

rakyll:
Also note that @bogdandrutu asked me rename the Set to Map given it is represented that way in other language implementations and it is a map itself.

jcd2:
> Also note that @bogdandrutu asked me rename the Set to Map given it is represented that way in other language implementations and it is a map itself.

That's fine.  We historically called these tag sets in the internal codebase, but keys weren't always unique!

Issue #53, opened 2017-10-17
Rename aggregations and windows


rakyll:
Aggregation implementations are using "Aggregation" as a prefix, Window implementations are using "Window" as a prefix. In Go, we prefer the natural language order.

Some example type names with the new order:

```
type CountAggregation struct{}
type DistributionAggregation struct{}

type SlidingTimeWindow struct {}
type SlidingCountWindow struct {}
```

Issue #17, opened 2017-10-02
Remove unnecessary constructors


rakyll:
There are lightweight types such as WindowSlidingTime, WindowSlidingCount, and WindowCumulative that are not complicated enough to require a constructor. Remove NewWindowCumulative and make the zero value useful, remove NewWindowSlidingCount and export fields on WindowSlidingCount, and remove NewWindowSlidingTime and export the fields of WindowSlidingTime.

/cc @hyangah @bogdandrutu

rakyll:
Actually the exported types are never useful, we should instead keep the constructors and unexport the types.

```
func NewWindowCumulative() Window
func NewWindowSlidingCount(count uint64, subSets int) Window
func NewWindowSlidingTime(duration time.Duration, subIntervals int) Window
```

acetechnologist:
The types will still be needed  by the exporters. That is how we can populate zpages and other exporters.

rakyll:
Can you give more info how the zpages and exporters need these types please? These types don't export anything. Will they be just type casted?

acetechnologist:
The data collected *ViewData returns the view along with the data collected (aka Rows).
The View have methods to return the interfaces: Window, Aggregation... The exporter needs to type switch on this interface to determine what Window or Aggregation it is and maybe extract or not data from them.
For example WindowSlidingTime needs to have a mehtod Duration() that returns the duration. This will allow the exporter to know that this data was for the last" 30s, 1mn, 5mn or whatever else.
We need Duration() instead of just exporting the field because we want them to be immutable once created.

This also applies to the interface AggregationValue returned in the Rows. It needs to be type switched to the right AggregationDistributionValue, AggregationCountValue...

rakyll:
Closed via #54 and #55.

PR #54, opened 2017-10-18
Remove the Window constructors

rakyll:
Instead allow users to init with a literal. Use "Window" as a prefix to follow the natural
language instead of the current name formatting.

Updates #17.
Updates #53.


rakyll:
I will also send out another PR for aggregation types.

PR #55, opened 2017-10-19
Rename aggregations and aggregation values

### LINKED ISSUE #53: Rename aggregations and windows

rakyll:
- Use Aggregation and AggregationValue as a suffix when naming.
- Remove unnecessary constructors.

Fixes #53.


rakyll:
This change is similar to https://github.com/census-instrumentation/opencensus-go/pull/54.

PR #63, opened 2017-10-23
Rename tags package to tag

rakyll:
In Go, we don't prefer pluralized package names given the full name always includes [package-name].[type-name]. Without pluralization package name represents the scope and acts as a prefix:

```
var m *tag.Map
```

This is a follow up from the conversation at https://github.com/census-instrumentation/opencensus-go/pull/51.

### DIFF
--- a/examples/stats/main.go
+++ b/examples/stats/main.go
@@ -25,17 +25,17 @@ import (
 	"time"
 
 	"github.com/census-instrumentation/opencensus-go/stats"
-	"github.com/census-instrumentation/opencensus-go/tags"
+	"github.com/census-instrumentation/opencensus-go/tag"
 )
 
 func main() {
 	// Creates keys. There will be two dimensions, device ID and OS version,
 	// that will be collected with the metrics we collect in the program.
-	deviceIDKey, err := tags.NewStringKey("/mycompany.com/key/deviceID")
+	deviceIDKey, err := tag.NewStringKey("/mycompany.com/key/deviceID")
 	if err != nil {
 		log.Fatalf("Device ID key not created: %v\n", err)
 	}
-	osVersionKey, err := tags.NewStringKey("/mycompany.com/key/osVersionKey")
+	osVersionKey, err := tag.NewStringKey("/mycompany.com/key/osVersionKey")
 	if err != nil {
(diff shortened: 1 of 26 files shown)


rakyll:
Please comment whether this is a yay or nay. I will fix the merge conflicts if we want to do this rename.

zombiezen:
I guess the only potential qualm is if `tag` is used commonly as a variable name while using this package, but it doesn't seem like it.

examples/stats/main.go
```diff
@@ -25,17 +25,17 @@ import (
 	"time"
 
 	"github.com/census-instrumentation/opencensus-go/stats"
-	"github.com/census-instrumentation/opencensus-go/tags"
+	"github.com/census-instrumentation/opencensus-go/tag"
```
bogdandrutu: I am not 100% convinced that this is a good thing to do. The reasons are the following:
1) There are packages using plural in the main golang (https://golang.org/pkg/) - bytes, errors, strings, etc.
2) All the other languages are using tags.
3) Using stat instead of "stats" is not a great choice.

rakyll: I am not planning to rename stats to stat. What's represented in stats is almost "stats collection" which sounds different when stats is singularized.

To be fair, you cannot name things after byte, error and string which are builtins in Go. I wouldn't use them as a reference.

I feel the second argument though. I am not convinced that I should merge this change.

rakyll: Given this a second thought, I am now supporting this change because the package name does not have the same importance in other languages the way it has in Go. In other languages, you (or your IDE automatically) import the symbol from a package and your call sites never have to deal with that name. We shouldn't care much about consistency among languages given they do not care about package organization and naming equally.

rakyll:
@bogdandrutu, can I have your final opinion on this?

bogdandrutu:
@rakyll I think I am good with this if this is a style in Go. Can we get someone from Go to also confirm this is the general style (double checking is better for this kind of decisions that we will have to live with them for long).

rakyll:
@zombiezen, can you confirm that this is the style in Go?

bogdandrutu:
@rakyll convinced me https://rakyll.org/style-packages/

PR #65, opened 2017-10-23
Enforce measures to return their descriptions and units

### LINKED ISSUE #64: Measures should return unit
All measure types need to be able to report their unit. Enforce measures to return a unit as string.

```
type Measure interface {
     Name() string
     Unit() string
}
```

rakyll:
The data model already expects all measures to have
descriptions and unit although it is not represented in
the Go interface.

Fixes #64.

### DIFF
--- a/stats/measure.go
+++ b/stats/measure.go
@@ -24,8 +24,12 @@ package stats
 // NewMeasureFloat64 automatically registers the measure
 // by the given name.
 // Each registered measure needs to be unique by name.
+// Measures also have units that represent the unit of the
+// metric they are representing.
 type Measure interface {
 	Name() string
+	Description() string
+	Unit() string
 
 	addView(v *View)
 	removeView(v *View)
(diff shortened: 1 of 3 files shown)


rakyll:
@dinooliva, can you take a look?

stats/measure.go
```diff
@@ -24,8 +24,12 @@ package stats
 // NewMeasureFloat64 automatically registers the measure
 // by the given name.
 // Each registered measure needs to be unique by name.
+// Measures also have units that represent the unit of the
+// metric they are representing.
 type Measure interface {
 	Name() string
+	Description() string
+	Unit() string
```
bogdandrutu: Oh, Unit was already defined as a public function but not added in the interface.

rakyll: Yes.

rakyll: To be more accurate, unit was already in the concrete types but not in the interface.

PR #69, opened 2017-10-24
Don't enforce aggregation values to provide pretty string

rakyll:


stats/view.go
```diff
@@ -182,7 +182,7 @@ func (r *Row) String() string {
 		buffer.WriteString(fmt.Sprintf("{%v %v}", t.Key.Name(), t.Key.ValueToString(t.Value)))
 	}
 	buffer.WriteString(" }")
-	buffer.WriteString(r.AggregationValue.String())
+	buffer.WriteString(fmt.Sprintf("%v", r.AggregationValue))
```
zombiezen: nit: `fmt.Fprint(&buffer, r.AggregationValue)`?

PR #72, opened 2017-10-25
Import the correct tag package

rakyll:
goimports is not going a good job again and picked up
a package I didn't intent to import. Replace it with
the actual tag package import path.

### DIFF
--- a/tag/example_test.go
+++ b/tag/example_test.go
@@ -19,7 +19,6 @@ import (
 	"log"
 
 	"github.com/census-instrumentation/opencensus-go/tag"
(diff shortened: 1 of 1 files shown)


tag/example_test.go
```diff
@@ -19,7 +19,6 @@ import (
 	"log"
 
 	"github.com/census-instrumentation/opencensus-go/tag"
-	"github.com/rakyll/census/tags"
```
bogdandrutu: Like this 👍

rakyll: I will add some linting to avoid this happening again.

PR #74, opened 2017-10-25
Introduce stats exporters

### LINKED ISSUE #73: Introduce exporters
Rather than allowing users access to the low-level buffer, introduce exporters and report the view data via the registered ones.

```
// Exporter exports the collected records as view data.
type Exporter interface {
	Export(viewData *ViewData)
}

// RegisterExporter registers an exporter.
func RegisterExporter(name string, e Exporter)

// UnregisterExporter unregisters an exporter.
func UnregisterExporter(name string)
```

As a follow up, we also need a rethinking of the current subscription model. In order to access to the view data of a view, you currently need to subscribe to a view after it is registered globally. Regardless of the registration status, view data can still be collected by force collection. Instead of having both methods available, we should either allow users to force collect or subscribe but not both.

rakyll:
Allow multiple exporters to be registered and report view data
on globally registered exporters.

Fixes #73.

### DIFF
--- a/stats/export.go
+++ b/stats/export.go
@@ -1,0 +1,45 @@
+// Copyright 2017, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package stats
+
+import "sync"
+
+var (
+	exportersMu sync.Mutex // guards exporters
+	exporters   = make(map[Exporter]struct{})
+)
+
+// Exporter exports the collected records as view data.
+type Exporter interface {
+	Export(viewData *ViewData)
+}
+
+// RegisterExporter registers an exporter.
+// Collected data will be reported via all the
+// registered exporters.
+func RegisterExporter(e Exporter) {
+	exportersMu.Lock()
+	defer exportersMu.Unlock()
+
+	exporters[e] = struct{}{}
+}
+
+// UnregisterExporter unregisters an exporter.
+func UnregisterExporter(e Exporter) {
+	exportersMu.Lock()
+	defer exportersMu.Unlock()
+
+	delete(exporters, e)
+}
--- a/stats/view.go
+++ b/stats/view.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"fmt"
 	"reflect"
+	"sync/atomic"
 	"time"
 
 	"github.com/census-instrumentation/opencensus-go/tag"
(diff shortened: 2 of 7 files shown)


rakyll:
/cc @bogdandrutu 

Currently a draft and but this is what I am thinking about for the exporters and the subscription model.

stats/export.go
```diff
@@ -0,0 +1,43 @@
...
+type Exporter interface {
+	Export(viewData *ViewData)
+}
+
+// RegisterExporter registers an exporter.
+func RegisterExporter(name string, e Exporter) {
+	exportersMu.Lock()
+	defer exportersMu.Unlock()
+
+	exporters[name] = e
```
bogdandrutu: Do you need to delete the previous value if already registered an exporter with that name? Maybe return an error?

stats/view.go
```diff
@@ -43,14 +43,14 @@ type View struct {
 
 	// ss are the channels through which the collected views data for this view
 	// are sent to the consumers of this view.
-	ss map[chan *ViewData]subscription
+	// ss map[chan *ViewData]subscription
```
bogdandrutu: Is the intention to remove this?

stats/worker.go
```diff
@@ -132,37 +131,28 @@ func (v *View) Unregister() error {
...
-		c:   c,
 		err: make(chan error),
 	}
 	defaultWorker.c <- req
 	return <-req.err
 }
 
-// Unsubscribe unsubscribes a previously subscribed channel from the
-// View subscriptions. If no more subscriber for v exists and the the ad hoc
+// Unsubscribe unsubscribes a previously subscribed channel.
```
bogdandrutu: No more channels correct?

stats/worker_commands.go
```diff
@@ -144,22 +144,19 @@ func (cmd *unregisterViewReq) handleCommand(w *worker) {
 // subscribeToViewReq is the command to subscribe to a view.
 type subscribeToViewReq struct {
 	v   *View
-	c   chan *ViewData
 	err chan error
```
bogdandrutu: Should this still be a chan error or just a normal error?

stats/view.go
```diff
@@ -43,14 +43,14 @@ type View struct {
...
 
 	// boolean to indicate if the the view should be collecting data even if no
 	// client is subscribed to it. This is necessary for supporting a pull
 	// model.
 	isForcedCollection bool
 
-	// TODO(jbd): Guard isForcedCollection.
```
bogdandrutu: It looks like you still need to protect the isForcedCollection, because this it is mutable correct?

stats/view.go
rakyll: Yea, will be removed. This is a draft, so I haven't cleaned it up yet.

stats/worker.go
rakyll: The external APIs won't work with channels anymore.

stats/worker_commands.go
rakyll: I explained you the worker model and in order to communicate with that worker, we use channels. The error channel is the communication channel for the return value.

stats/export.go
rakyll: Not sure if returning an error is helpful. How do the user handle that error?

Most registration patterns in the stdlib (e.g. [sql.Register](https://golang.org/pkg/database/sql/#Register) is panic'ing so the user can detect the problem during the develop time. Do we expect users to register exporters dynamically? If not, I am suggesting to panic. If not, we can replace and document it accordingly.

stats/view.go
rakyll: Correct, it is not mutated by the event loop. We need to guard it.

rakyll:
@bogdandrutu, can you take a look now?

stats/view.go
```diff
@@ -41,41 +42,23 @@ type View struct {
 	// start is time when view collection was started originally.
 	start time.Time
 
-	// ss are the channels through which the collected views data for this view
-	// are sent to the consumers of this view.
-	ss map[chan *ViewData]subscription
+	collectionMu sync.Mutex // guards forced and subscribed
+	forced       bool       // indicates whether the view should collect data if no one is subscribed
```
bogdandrutu: Can we use atomic variables instead of a mutex here?

rakyll: Done

PR #75, opened 2017-10-25
Remove dead code

rakyll:
recordReq is the general implementation that replaces
recordFloat64Req and recordInt64Req.

Remove slidingTimeSubuckets.

### DIFF
--- a/stats/worker_commands.go
+++ b/stats/worker_commands.go
@@ -257,40 +257,6 @@ func (cmd *retrieveDataReq) handleCommand(w *worker) {
 	}
 }
 
-// recordFloat64Req is the command to record data related to a measure.
-type recordFloat64Req struct {
-	now time.Time
-	tm  *tag.Map
-	mf  *MeasureFloat64
-	v   float64
-}
-
-func (cmd *recordFloat64Req) handleCommand(w *worker) {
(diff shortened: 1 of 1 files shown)


stats/worker_commands.go
```diff
@@ -263,40 +263,6 @@ func (cmd *retrieveDataReq) handleCommand(w *worker) {
 	}
 }
 
-// recordFloat64Req is the command to record data related to a measure.
-type recordFloat64Req struct {
```
bogdandrutu: Why is this dead code? Do we need it in the future?

rakyll: It is explained in the description. recordReq provides a multi-type solution. These types have been deprecated but not removed.

```
type recordReq struct {
	now time.Time
	tm  *tag.Map
	ms  []Measurement
}

func (cmd *recordReq) handleCommand(w *worker) {
	for _, m := range cmd.ms {
		switch measurement := m.(type) {
		case *measurementFloat64:
			for v := range measurement.m.views {
				v.addSample(cmd.tm, measurement.v, cmd.now)
			}
		case *measurementInt64:
			for v := range measurement.m.views {
				v.addSample(cmd.tm, measurement.v, cmd.now)
			}
		default:
		}
	}
}
```

PR #76, opened 2017-10-28
Add a Stackdriver Monitoring exporter

rakyll:
Introduce stats/exporter/stackdriver and enable exporting to
Stackdriver Monitoring.

Exporter batches the view data and periodically uploads
to the backend. It creates metric descriptors for exported
measure types.

The exporter currently only support count and distribution aggregation
and returns an error if another aggregation type is provided.

### DIFF
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -1,0 +1,266 @@
+// Copyright 2017, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package stackdriver contains the OpenCensus exporters for
+// Stackdriver Monitoring.
+//
+// Please note that the Stackdriver exporter is currently experimental.
+package stackdriver
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"path"
+	"sync"
+	"time"
+
+	"go.opencensus.io/stats"
+
+	monitoring "cloud.google.com/go/monitoring/apiv3"
+	timestamp "github.com/golang/protobuf/ptypes/timestamp"
+	"google.golang.org/api/option"
+	"google.golang.org/api/support/bundler"
+	distributionpb "google.golang.org/genproto/googleapis/api/distribution"
+	metricpb "google.golang.org/genproto/googleapis/api/metric"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+)
+
+// Exporter exports stats to the Stackdriver Monitoring.
+type Exporter struct {
+	// ProjectID is the identifier of the Stackdriver
+	// project the user is uploading the stats data to.
+	ProjectID string
+
+	// OnError is the hooked to be called when there is
+	// an error occured when uploading the stats data.
+	// If no custom hook is set, errors are logged.
+	// Optional.
+	OnError func(err error)
+
+	// ClientOptions are additional options to be passed
+	// to the underlying Stackdriver Monitoring API client.
+	// Optional.
+	ClientOptions []option.ClientOption
+
+	// ExportDelayThreshold determines the max amount of time
+	// the exporter can wait before uploading view data to
+	// the backend.
+	// Optional.
+	ExportDelayThreshold time.Duration
+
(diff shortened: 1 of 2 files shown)


stats/exporter/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,242 @@
...
+			Points: newPoints(vd),
+		}
+	}
+	return &monitoringpb.CreateTimeSeriesRequest{
+		Name:       monitoring.MetricProjectPath(e.ProjectID),
+		TimeSeries: timeSeries,
+	}
+}
+
+func (e *Exporter) createMeasure(ctx context.Context, vd *stats.ViewData) error {
```
zombiezen: Is `createMeasure` called concurrently?  I'm trying to reason through possible contention, since this makes an RPC while holding the mutex.

stats/exporter/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,242 @@
...
+		metricKind = metricpb.MetricDescriptor_CUMULATIVE
+		valueType = metricpb.MetricDescriptor_INT64
+	case stats.DistributionAggregation:
+		metricKind = metricpb.MetricDescriptor_GAUGE
+		valueType = metricpb.MetricDescriptor_DISTRIBUTION
+	default:
+		return fmt.Errorf("unsupported aggregation type: %T", agg)
+	}
+
+	if _, err := e.c.CreateMetricDescriptor(ctx, &monitoringpb.CreateMetricDescriptorRequest{
```
zombiezen: `CreateMetricDescriptor` is a very slow RPC IIRC from our testing.  I don't think you want to be calling it in the tighter update loop.

stats/exporter/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	if err := e.c.CreateTimeSeries(ctx, e.makeReq(vds)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (e *Exporter) makeReq(vds []*stats.ViewData) *monitoringpb.CreateTimeSeriesRequest {
+	timeSeries := make([]*monitoringpb.TimeSeries, len(vds))
+	for i, vd := range vds {
```
songy23: I think each `Row` in a `ViewData` should be converted to one `TimeSeries` (i.e one `ViewData` -> a list of `TimeSeries`), because one `TimeSeries` can only have one set of labels (tags).

exporter in Java: https://github.com/census-instrumentation/opencensus-java/pull/713/files#diff-d2886f5b11a5e3be7a00a8f5fff5a1f0R148

stats/exporter/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,242 @@
...
+
+	var metricKind metricpb.MetricDescriptor_MetricKind
+	var valueType metricpb.MetricDescriptor_ValueType
+
+	switch agg.(type) {
+	case stats.CountAggregation:
+		metricKind = metricpb.MetricDescriptor_CUMULATIVE
+		valueType = metricpb.MetricDescriptor_INT64
+	case stats.DistributionAggregation:
+		metricKind = metricpb.MetricDescriptor_GAUGE
```
songy23: Should this match `View.Window`? i.e CumulativeWindow -> CUMULATIVE, SlidingCountWindow -> DELTA

stats/exporter/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,242 @@
...
+		valueType = metricpb.MetricDescriptor_DISTRIBUTION
+	default:
+		return fmt.Errorf("unsupported aggregation type: %T", agg)
+	}
+
+	if _, err := e.c.CreateMetricDescriptor(ctx, &monitoringpb.CreateMetricDescriptorRequest{
+		Name: path.Join("projects", e.ProjectID, "metricDescriptors", "custom.googleapis.com", m.Name()),
```
songy23: I think `Name` should be `"projects/{project_id_or_number}"`, according to https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors/create.

stats/exporter/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,242 @@
...
+		MetricDescriptor: &metricpb.MetricDescriptor{
+			DisplayName: m.Name(),
```
songy23: In Java we chose `ViewName` instead of `MeasureName` for `MetricDescriptor`, because we believe `MetricDescriptor` is more align with our `View` data model since they both have columns (`LabelDescriptor/TagKey`).

stats/exporter/stackdriver/stackdriver.go
songy23: I just tried with the API Explorer (https://developers.google.com/apis-explorer/#p/monitoring.googleapis.com/v3/monitoring.projects.metricDescriptors.create), it seems DELTA is not supported for custom metric :( For Java we don't support exporting interval/sliding window views in this version.

rakyll:
@songy23, yes the worker calls the registered exporters' Export with the view data.

bogdandrutu:
Can we have a working demo app that does this?

bogdandrutu:
One minor thing I would have:
exporter/stats/stackdriver/stackdriver.go

rakyll:
@bogdandrutu, made the rename. I will send an example program in another PR.

PR #78, opened 2017-10-29
Redesign of the gRPC stats plugin

rakyll:
- Remove the composite stats handler utility. Tracing
will be handled by interceptors and we will provide a utility
configuration utility that sets both the stats handler and the
tracing interceptors in the future.

- Rename the package to grpcstats in order to avoid
name collusion with the OpenCensus stats and grpc/stats
package.

- Export instrumentation handlers for clients and servers.

- Several naming and organization improvements.


rakyll:
@adriancole, yay! We have one more round of reorg/redesign when the tracing package is available. Then, it will be in a better shape.

Issue #33, opened 2017-10-09
Setup vanity import paths


rakyll:
To reduce to dependencies to a specific git provider and to enhance the the current long and noisy import path, setup a vanity URL for the package and enforce the vanity URL everywhere. The vanity URL is either going to be "go.opencensus.io" or "opencensus.io/go".

rakyll:
We are going with go.opencensus.io.

/cc @tags07 @mtwo

rakyll:
go get go.opencensus.io finally works!

I will document how to deploy the vanity URL server before closing this issue.

bogdandrutu:
Thanks @rakyll.

bogdandrutu:
Probably also to track the update of the documentation and examples to use the vanity import in this issue.

PR #85, opened 2017-10-31
Enforce the new vanity import path

rakyll:
By enforcing the vanity URL, users won't be able to go get from the github.com URL.

Updates #33.


rakyll:
Sorry for the broken tests, all fixed now.

PR #98, opened 2017-11-10
Only allow string keys

### LINKED ISSUE #93: Finalize the case for typed keys
Tags allow typed keys such (e.g. StringKey) but there is interest to drop typed keys and have StringKey as the only option. Determine the future for the tag package and remove the unnecessary abstractions before freezing the APIs.

/cc @dinooliva

rakyll:
This change is due to a data model change. OpenCensus won't no longer
support keys other than string keys.

Discontinue typed keys, allow string-only keys. Remove Key interface
and implementations and make Key a concrete type that represents
string keys.

Also the representation of the values are now string given we don't have
any values other than string.

Fixes #93.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Other types will be supported in the future.
 [embedmd]:# (tags.go stringKey)
 ```go
 // Get a key to represent user OS.
-key, err := tag.NewStringKey("my.org/keys/user-os")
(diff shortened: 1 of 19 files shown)


rakyll:
(Fixing the test... You still can review the API, etc)

PR #112, opened 2017-11-14
NewMap should accept the incoming context

### LINKED ISSUE #110: NewMap should take a context
We would like to avoid users accidentally dropping the tags in the current context. Make NewMap working against a context.Context rather than an initial tag.Map.

```
func NewMap(ctx context.Context, m ...Mutator)
```

rakyll:
In order to encourage users not to drop the tags from the
incoming context, adjust APIs to accept the incoming context
for convience.

Fixes #110.

### DIFF
--- a/internal/readme/tags.go
+++ b/internal/readme/tags.go
@@ -43,7 +43,7 @@ func tagsExamples() {
 		log.Fatal(err)
 	}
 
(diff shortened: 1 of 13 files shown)


rakyll:
Depends on https://github.com/census-instrumentation/opencensus-go/pull/108.

PR #123, opened 2017-11-16
Rename WithSpan to NewContext

rakyll:
The Go ecosystem prefer either NewContext or WithX naming style as a short cut for NewContextWithX. Given we won't provide more NewContextWithX functions from this package, it is it is fair to rename WithSpan to NewContext.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -117,8 +117,8 @@ func FromContext(ctx context.Context) *Span {
 	return s
 }
 
-// WithSpan returns a new context with the given Span attached.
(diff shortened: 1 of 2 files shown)


jcd2:
I like foo.NewContext in general, but on reflection, I think trace.WithSpan is actually better than trace.NewContext in this case, because trace.WithSpan just makes it a bit clearer that it's a Span object we're adding to the context, not an object called Trace.

rakyll:
I agree with the last statement and am closing this PR.

PR #128, opened 2017-11-17
Provide composite trace and stats handler for gRPC

### LINKED ISSUE #124: Provide composite gRPC stats handler
Provide a stats handler that does enables both stats and tracing. Users who prefer not to have both can depend on grpcstats and grpctrace packages individually.

rakyll:
For most cases, users will likely to have them both enabled.
Provide a composite handler and document that grpctrace and
grpcstats handlers are available if users only want to
enable tracing or stats collection.

Fixes #124.

### DIFF
--- a/plugins/grpc/grpc.go
+++ b/plugins/grpc/grpc.go
@@ -1,0 +1,74 @@
+// Copyright 2017, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package grpc contains OpenCensus stats and trace
+// integrations with gRPC.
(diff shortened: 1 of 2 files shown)


plugins/grpc/grpc_test.go
```diff
@@ -0,0 +1,127 @@
...
+	te := &traceExporter{}
+	trace.RegisterExporter(te)
+	if err := grpcstats.RPCClientRequestCountView.Subscribe(); err != nil {
+		t.Fatal(err)
+	}
+
+	span := trace.NewSpan("/foo", trace.StartSpanOptions{
+		Sampler: trace.AlwaysSample(),
+	})
+	ctx = trace.WithSpan(ctx, span)
```
jba: FWIW, `WithSpan` reads fine to me here.

plugins/grpc/grpc.go
```diff
@@ -0,0 +1,74 @@
...
+// on, see grpcstats and grpctrace packages.
+func NewClientStatsHandler() stats.Handler {
+	return handler{
+		grpcstats.NewClientStatsHandler(),
+		grpctrace.NewClientStatsHandler(),
+	}
+}
+
+// NewServerStatsHandler enables OpenCensus stats and trace
+// for gRPC clients. If these features need to be indiviually turned
```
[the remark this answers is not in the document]
rakyll: Done.

plugins/grpc/grpc_test.go
rakyll: Ack.

rakyll:
PTAL!

PR #131, opened 2017-11-19
Add net/http integration to trace the outgoing requests

rakyll:


plugins/http/httptrace/httptrace.go
```diff
@@ -0,0 +1,63 @@
...
+	req = req.WithContext(trace.WithSpan(req.Context(), span))
+	resp, err := t.base().RoundTrip(req)
+	// TODO(jbd): Add status and attributes.
+	span.End()
+	return resp, err
+}
+
+// CancelRequest cancels an in-flight request by closing its connection.
+func (t *Transport) CancelRequest(req *http.Request) {
+	type canceler interface {
```
odeke-em: Could we hoist this out of this method?

rakyll: Not going to optimize anything. It is self-contained here.

odeke-em: Aye aye, cool :)

rakyll:
/cc @jba

We will also use this plugin in the tracing of the google-api-go-client clients.

jba:
LGTM.

plugins/http/httptrace/httptrace.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Transport is an http.RoundTripper that traces the outgoing requests.
+type Transport struct {
+	// Base is the base http.RoundTripper to be used to do the actual request.
+	//
+	// Optional. If nil, http.DefaultTransport is used.
+	Base http.RoundTripper
+}
+
+// RoundTrip creates a trace.Span and inserts it into the outgoing request's headers.
```
nvartolomei: isn't it missing `inserts it into the outgoing request's headers` part?

rakyll: True, we probably shouldn't merge this until we sort out how to determine which vendor specific header we should inject depending on the registered injectors. I am going to label it as a "do not review".

rakyll:
Closed via https://github.com/census-instrumentation/opencensus-go/commit/6c0bc0d00b45f8e5e7a4cfc70071722739f1f1b1.

PR #142, opened 2017-11-21
Add view and measure name validation

### LINKED ISSUE #139: Validate names, descriptions, units, etc.
User input need to be validated whether they are fitting the restrictions we set. Use the Java client as a reference to validate the input.

### LINKED ISSUE #141: exporter/stackdriver: MetricDescriptor name should be under domain "custom.googleapis.com/opencensus/"
Thanks @bogdandrutu for catching this.

For the Stackdriver Stats exporter, the name of MetricDescriptor needs to be under "custom.googleapis.com/opencensus/". The format of the whole name should be `String.format("projects/%s/metricDescriptors/custom.googleapis.com%%2Fopencensus%%2F%s", projectId, viewName.replace("/", "%2F"))`. Stackdriver might return a permission denied error if we don't follow the name convention.

See https://github.com/census-instrumentation/opencensus-java/pull/827/files#diff-d2886f5b11a5e3be7a00a8f5fff5a1f0R84 for changes to Java exporter. Also refer to https://cloud.google.com/monitoring/custom-metrics/creating-metrics#custom_metric_names.

rakyll:
Do not allow names to be non-ASCII and above the permitted length limit.

Fixes #139.


stats/validation.go
```diff
@@ -0,0 +1,52 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package stats
+
+import "fmt"
+
+const (
+	maxViewNameLength    = 256
```
songy23: This one should be 255 as well - thanks for catching that.

rakyll: Done.

rakyll:
PTAL

PR #153, opened 2017-11-27
Rename directory "plugins" to "plugin"

### LINKED ISSUE #152: Rename plugins to plugin
We should either use singular of plural. exporter subdirectory is singular, plugins is plural.

/cc @odeke-em

odeke-em:
Fixes #152


odeke-em:
Thank you for the review @rakyll!

PR #156, opened 2017-11-27
Access atomic variables always atomically

### LINKED ISSUE #151: plugin/grpc/grpcstats: handling of atomics
rpcData.reqCount and rpcData.respCount are written to atomically but not read atomically, which doesn't seem right.

rakyll:
- Convert reqCount and respCount to int64 to match the measure type.
- Don't just write them atomically, also read them atomically.

Fixes #151.


rakyll:
But we are casting them back to int64, that's why I don't know why uint64 was in use.

odeke-em:
Yeah, I noticed that too, great point.

PR #160, opened 2017-11-27
Cache Views instead of Measures for Exporter

songy23:
Each `View` will be uploaded as one `MetricDescriptor`, instead of `Measure`. One `Measure` could have multiple `View`s, and they should be uploaded as different `MetricDescriptor`s.

### DIFF
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -48,8 +48,8 @@ type Exporter struct {
 	bundler *bundler.Bundler
 	o       Options
 
-	measuresMu sync.Mutex
-	measures   map[stats.Measure]struct{} // measures already created remotely
+	viewMu          sync.Mutex
+	registeredViews map[string]stats.View // Views already created remotely
 
 	c *monitoring.MetricClient
 }
@@ -91,9 +91,9 @@ func NewExporter(o Options) (*Exporter, error) {
 		return nil, err
 	}
 	e := &Exporter{
-		c:        client,
-		o:        o,
-		measures: make(map[stats.Measure]struct{}),
+		c:               client,
+		o:               o,
+		registeredViews: make(map[string]stats.View),
 	}
 	e.bundler = bundler.NewBundler((*stats.ViewData)(nil), func(bundle interface{}) {
 		vds := bundle.([]*stats.ViewData)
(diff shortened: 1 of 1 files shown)


exporter/stats/stackdriver/stackdriver.go
```diff
@@ -48,8 +48,8 @@ type Exporter struct {
 	bundler *bundler.Bundler
 	o       Options
 
-	measuresMu sync.Mutex
-	measures   map[stats.Measure]struct{} // measures already created remotely
+	viewMu          sync.Mutex
+	registeredViews map[string]stats.View // Views already created remotely
```
rakyll: createdViews maybe?

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -91,9 +91,9 @@ func NewExporter(o Options) (*Exporter, error) {
 		return nil, err
 	}
 	e := &Exporter{
-		c:        client,
-		o:        o,
-		measures: make(map[stats.Measure]struct{}),
+		c:               client,
+		o:               o,
+		registeredViews: make(map[string]stats.View),
```
rakyll: map[string]struct{} is enough, we don't need to access the view.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -187,14 +187,14 @@ func (e *Exporter) makeReq(vds []*stats.ViewData) *monitoringpb.CreateTimeSeries
...
 
 func (e *Exporter) createMeasure(ctx context.Context, vd *stats.ViewData) error {
-	e.measuresMu.Lock()
-	defer e.measuresMu.Unlock()
+	e.viewMu.Lock()
+	defer e.viewMu.Unlock()
 
 	m := vd.View.Measure()
 	agg := vd.View.Aggregation()
 	window := vd.View.Window()
```
rakyll: name := d.View.Name()

Then use it elsewhere.

exporter/stats/stackdriver/stackdriver.go
songy23: Done.

exporter/stats/stackdriver/stackdriver.go
songy23: Done.

exporter/stats/stackdriver/stackdriver.go
songy23: Done. I named it `viewName`, to avoid confusion with `metricName`.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -48,8 +48,8 @@ type Exporter struct {
 	bundler *bundler.Bundler
 	o       Options
 
-	measuresMu sync.Mutex
-	measures   map[stats.Measure]struct{} // measures already created remotely
+	viewMu       sync.Mutex
```
rakyll: createdViewsMu

Always name the mutex after the field it is protecting.

songy23: Done.

songy23:
Thanks!

Issue #145, opened 2017-11-21
exporter/stackdriver: add labels


rakyll:
Metrics are uploaded without the recorded tags being attached as Stackdriver Monitoring labels. Fix the exporter by adding the labels.

PR #161, opened 2017-11-27
Export Label and LabelDescriptor to Stackdriver

songy23:
Fix https://github.com/census-instrumentation/opencensus-go/issues/145

### DIFF
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -327,9 +327,8 @@ func newLabelDescriptors(keys []tag.Key) []*labelpb.LabelDescriptor {
 		labelDescriptors = append(
 			labelDescriptors,
 			&labelpb.LabelDescriptor{
-				Key:         key.Name(),
-				ValueType:   labelpb.LabelDescriptor_STRING, // We only use string tags
-				Description: "OpenCensus TagKey",
+				Key:       key.Name(),
+				ValueType: labelpb.LabelDescriptor_STRING, // We only use string tags
 			})
 	}
 	return labelDescriptors
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -38,7 +38,7 @@ import (
 	"google.golang.org/api/option"
 	"google.golang.org/api/support/bundler"
 	distributionpb "google.golang.org/genproto/googleapis/api/distribution"
-	labeldescriptorpb "google.golang.org/genproto/googleapis/api/label"
+	labelpb "google.golang.org/genproto/googleapis/api/label"
 	metricpb "google.golang.org/genproto/googleapis/api/metric"
 	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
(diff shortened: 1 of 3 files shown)


exporter/stats/stackdriver/stackdriver.go
```diff
@@ -309,3 +312,25 @@ func namespacedViewName(v string, escaped bool) string {
...
+
+func newLabelDescriptors(keys []tag.Key) []*labelpb.LabelDescriptor {
+	labelDescriptors := make([]*labelpb.LabelDescriptor, len(keys))
+	for _, key := range keys {
+		labelDescriptors = append(
+			labelDescriptors,
+			&labelpb.LabelDescriptor{
+				Key:         key.Name(),
+				ValueType:   labelpb.LabelDescriptor_STRING, // We only use string tags
+				Description: "OpenCensus TagKey",
```
rakyll: Do we need a description? If yes, I'd prefer we use "OpenCensus tag key". TagKey sounds programmatic.

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -35,7 +36,13 @@ func TestExporter_makeReq(t *testing.T) {
 	}
 	defer stats.DeleteMeasure(m)
 
-	cumView, err := stats.NewView("cumview", "desc", nil, m, stats.CountAggregation{}, stats.Cumulative{})
+	key, err := tag.NewKey("test-key")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tag.Delete(key)
```
rakyll: This is not deleting the key but returning a mutator to delete a value keyed with key.

(FWIW, I think the naming for Insert, Upsert, Delete, etc is not great and creates confusion.)

exporter/stats/stackdriver/stackdriver.go
songy23: This is not a required filed, I think it's safe to omit it for now.

exporter/stats/stackdriver/stackdriver_test.go
songy23: Agree, this method is a little confusing to me at first. Removed.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -309,3 +312,24 @@ func namespacedViewName(v string, escaped bool) string {
...
+func newLabels(tags []tag.Tag) map[string]string {
+	labels := make(map[string]string)
+	for _, tag := range tags {
+		labels[tag.Key.Name()] = tag.Value
+	}
+	return labels
+}
+
+func newLabelDescriptors(keys []tag.Key) []*labelpb.LabelDescriptor {
+	labelDescriptors := make([]*labelpb.LabelDescriptor, len(keys))
```
jcd2: This is setting the slice size to len(keys), not its capacity.  So:
labelDescriptors := make([]*labelpb.LabelDescriptor, 0, len(keys))

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -309,3 +312,24 @@ func namespacedViewName(v string, escaped bool) string {
...
+	for _, key := range keys {
+		labelDescriptors = append(
+			labelDescriptors,
+			&labelpb.LabelDescriptor{
+				Key:       key.Name(),
```
jcd2: The tag keys will need to be escaped for Stackdriver.  It limits length to 100 and only accepts some characters.  Letters, digits and underscores work.

songy23: Sure, I added string sanitization in https://github.com/census-instrumentation/opencensus-go/pull/167

exporter/stats/stackdriver/stackdriver.go
songy23: Updated in https://github.com/census-instrumentation/opencensus-go/pull/167

PR #163, opened 2017-11-28
Export ClientStatsHandler and ServerStatsHandler from grpcstats

rakyll:
Exporting these handlers makes the package more compatible with
the grpctrace package and allow us to document the behavior via godoc.

/cc @dfawley

### DIFF
--- a/plugin/grpc/example_test.go
+++ b/plugin/grpc/example_test.go
@@ -1,0 +1,50 @@
+// Copyright 2017, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


jcd2:
Exporting the types looks fine but I don't think we should delete the New... constructors.  Being able to use zero-values directly is nice, but while the constructor is currently trivial, we don't want to rule out adding more to it later.

rakyll:
I added the constructors back and added a TODO to remove them if they are not necessary. If the zero value will be always usable (seems like it will be), it is not encouraged to have a constructor for the zero value.

PR #167, opened 2017-11-28
Add sanitization to tag key names for Exporter

songy23:
Quote from @jcd2 :
> The tag keys will need to be escaped for Stackdriver. It limits length to 100 and only accepts some characters. Letters, digits and underscores work.

### DIFF
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -407,9 +407,11 @@ func sanitize(s string) string {
 	return strings.Map(sanitizeRune, s)
 }
 
+// converts anything that is not a letter or digit to an underscore
 func sanitizeRune(r rune) rune {
-	if unicode.IsLetter(r) || unicode.IsDigit(r) || r == 95 {
+	if unicode.IsLetter(r) || unicode.IsDigit(r) {
 		return r
 	}
-	return 95 // Sanitize everything that is not letter, digit or underscore to underscore
+	// Everything else turns into an underscore
+	return '_'
 }
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -26,8 +26,10 @@ import (
 	"net/url"
 	"path"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
+	"unicode"
 
 	"go.opencensus.io/tag"
 
@@ -87,6 +89,8 @@ type Options struct {
 	// Optional.
 	BundleCountThreshold int
 }
+
+const labelKeySizeLimit = 100
 
 // NewExporter returns an exporter that uploads stats data to Stackdriver Monitoring.
 func NewExporter(o Options) (*Exporter, error) {
@@ -328,18 +332,18 @@ func namespacedViewName(v string, escaped bool) string {
 func newLabels(tags []tag.Tag) map[string]string {
 	labels := make(map[string]string)
 	for _, tag := range tags {
-		labels[tag.Key.Name()] = tag.Value
+		labels[sanitize(tag.Key.Name())] = tag.Value
 	}
 	return labels
 }
 
 func newLabelDescriptors(keys []tag.Key) []*labelpb.LabelDescriptor {
-	labelDescriptors := make([]*labelpb.LabelDescriptor, len(keys))
+	labelDescriptors := make([]*labelpb.LabelDescriptor, 0, len(keys))
 	for _, key := range keys {
 		labelDescriptors = append(
 			labelDescriptors,
 			&labelpb.LabelDescriptor{
-				Key:       key.Name(),
+				Key:       sanitize(key.Name()),
 				ValueType: labelpb.LabelDescriptor_STRING, // We only use string tags
 			})
 	}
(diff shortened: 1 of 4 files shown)


exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -148,8 +149,8 @@ func TestExporter_makeReq(t *testing.T) {
 }
 
 func TestEqualAggWindowTagKeys(t *testing.T) {
-	key1, _ := tag.NewKey("test-key-one")
-	key2, _ := tag.NewKey("test-key-two")
+	key1, _ := tag.NewKey("test_key_one")
```
rakyll: We also need some validation checks to the NewKey, are any rules here overlapping with the validation check we will implement for NewKey?

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -249,6 +250,39 @@ func TestEqualAggWindowTagKeys(t *testing.T) {
...
+			expected: strings.Repeat("a", 100),
+		},
+		{
+			name:     "replace character",
+			input:    "test/key-1",
+			expected: "test_key_1",
+		},
+		{
+			name:     "don't modify alphanumeric string",
+			input:    "testkey1",
```
rakyll: I'd add an input with all the acceptable characters for this case.

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -249,6 +250,39 @@ func TestEqualAggWindowTagKeys(t *testing.T) {
 	}
 }
 
+func TestSanitize(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
```
rakyll: s/expected/want

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -249,6 +250,39 @@ func TestEqualAggWindowTagKeys(t *testing.T) {
...
+		{
+			name:     "don't modify alphanumeric string",
+			input:    "testkey1",
+			expected: "testkey1",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			actual := sanitize(tt.input)
```
rakyll: if got, want := sanitize(tt.input), tt.want; got != want {
      t.Errorf("sanitize() = %q; want %q", got, want)
}

got/want is the jargon in Go and use %q to escape.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -397,3 +401,17 @@ func equalAggWindowTagKeys(md *metricpb.MetricDescriptor, agg stats.Aggregation,
...
+
+func sanitize(s string) string {
+	if len(s) > labelKeySizeLimit {
+		s = s[:labelKeySizeLimit]
+	}
+	return strings.Map(sanitizeRune, s)
+}
+
+func sanitizeRune(r rune) rune {
+	if !(unicode.IsLetter(r) || unicode.IsDigit(r) || r == 95) {
```
rakyll:
```
if unicode.IsLetter(r) || unicode.IsDigit(r) || r == 95) {
 return r
}
return 95
```

reads easier to me.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -328,18 +332,18 @@ func namespacedViewName(v string, escaped bool) string {
...
 	for _, tag := range tags {
-		labels[tag.Key.Name()] = tag.Value
+		labels[sanitize(tag.Key.Name())] = tag.Value
 	}
 	return labels
 }
 
 func newLabelDescriptors(keys []tag.Key) []*labelpb.LabelDescriptor {
-	labelDescriptors := make([]*labelpb.LabelDescriptor, len(keys))
+	labelDescriptors := make([]*labelpb.LabelDescriptor, 0, len(keys))
```
rakyll: Or you can just do

for i, key := range keys {
labelDescriptors[i] = ...

to avoid append all together and not caring about size vs capacity when making the slice.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -397,3 +399,17 @@ func equalAggWindowTagKeys(md *metricpb.MetricDescriptor, agg stats.Aggregation,
...
+
+func sanitize(s string) string {
+	if len(s) > labelKeySizeLimit {
+		s = s[:labelKeySizeLimit]
+	}
+	return strings.Map(sanitizeRune, s)
+}
+
+func sanitizeRune(r rune) rune {
+	if unicode.IsLetter(r) || unicode.IsDigit(r) || r == 95 {
```
odeke-em: Could we please use '_' instead of the code for an underscore?

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -397,3 +399,17 @@ func equalAggWindowTagKeys(md *metricpb.MetricDescriptor, agg stats.Aggregation,
...
+		return r
+	}
+	return 95 // Sanitize everything that is not letter, digit or underscore to underscore
```
odeke-em: Ditto about '_'

odeke-em: I think we can move this comment to the top of the function to make it
```go
// converts anything that is not a letter or digit to an underscore
func sanitizeRune(r rune) rune {
     if unicode.IsLetter(r) || unicode.IsDigit(r) {
        return r
    }
    // Everything else turns into an underscore
    return '_'
}
```

exporter/stats/stackdriver/stackdriver.go
songy23: Sure, done.

exporter/stats/stackdriver/stackdriver.go
songy23: Done.

songy23: Done.

odeke-em:
@songy23 another suggestion: since we'll be adding other exporters and reusing the sanitize code, perhaps we could move and expose Sanitize to `internal`?

songy23:
Sure, I've moved it to package internal. We might need other sanitize method for other exporters though, depending on the restrictions of services.

songy23:
Thanks!

PR #169, opened 2017-11-28
Always depend on the stats handler constructors

rakyll:
A follow up for https://github.com/census-instrumentation/opencensus-go/pull/163. A bad merge happened and reverted this change.


rakyll:
We keep exporting the structs because their godoc helps us document the behavior. We may end up not exporting the NewX constructors because they don't add much value right now.

Issue #109, opened 2017-11-13
stats: implement mean and sum aggregations


rakyll:
As well as aggregation data types:

```
type MeanData float64
type SumData float64
```

PR #172, opened 2017-11-29
Add Sum and Mean Aggregation and AggregationData

songy23:
Fix https://github.com/census-instrumentation/opencensus-go/issues/109

### DIFF
--- a/stats/aggregation_data.go
+++ b/stats/aggregation_data.go
@@ -31,6 +31,8 @@ type AggregationData interface {
 	clear()
 }
 
+const epsilon = 1e-9
+
 // CountData is the aggregated data for a CountAggregation.
 // A count aggregation processes data and counts the recordings.
 //
@@ -88,7 +90,7 @@ func newSumData(v float64) *SumData {
 func (a *SumData) isAggregate() bool { return true }
 
 func (a *SumData) addSample(v interface{}) {
-	// Both float64 and int64 values will be wrapped to float64
+	// Both float64 and int64 values will be cast to float64
 	var f float64
 	switch x := v.(type) {
 	case int64:
@@ -98,7 +100,7 @@ func (a *SumData) addSample(v interface{}) {
 	default:
 		return
 	}
-	*a = SumData(float64(*a) + f)
+	*a += SumData(f)
 }
 
 func (a *SumData) multiplyByFraction(fraction float64) AggregationData {
@@ -122,8 +124,6 @@ func (a *SumData) equal(other AggregationData) bool {
 	if !ok {
 		return false
 	}
-
-	epsilon := math.Pow10(-9)
 	return math.Pow(float64(*a)-float64(*a2), 2) < epsilon
 }
 
@@ -132,11 +132,11 @@ func (a *SumData) equal(other AggregationData) bool {
 //
 // Most users won't directly access mean data.
 type MeanData struct {
-	Count int64   // number of data points aggregated
+	Count float64 // number of data points aggregated
 	Mean  float64 // mean of all data points
 }
 
-func newMeanData(mean float64, count int64) *MeanData {
+func newMeanData(mean float64, count float64) *MeanData {
 	return &MeanData{
 		Mean:  mean,
 		Count: count,
@@ -167,9 +167,9 @@ func (a *MeanData) addSample(v interface{}) {
 	a.Mean = a.Mean + (f-a.Mean)/float64(a.Count)
 }
 
-// Only Mean will be mutiplied by the fraction, Count will remain the same.
+// Only Count will be mutiplied by the fraction, Mean will remain the same.
 func (a *MeanData) multiplyByFraction(fraction float64) AggregationData {
-	return newMeanData(a.Mean*fraction, a.Count)
+	return newMeanData(a.Mean, a.Count*fraction)
 }
 
 func (a *MeanData) addToIt(av AggregationData) {
@@ -182,7 +182,7 @@ func (a *MeanData) addToIt(av AggregationData) {
 		return
 	}
 
-	a.Mean = (a.Sum() + other.Sum()) / float64(a.Count+other.Count)
+	a.Mean = (a.Sum() + other.Sum()) / (a.Count + other.Count)
 	a.Count = a.Count + other.Count
 }
 
@@ -196,8 +196,6 @@ func (a *MeanData) equal(other AggregationData) bool {
 	if !ok {
 		return false
 	}
-
-	epsilon := math.Pow10(-9)
 	return a.Count == a2.Count && math.Pow(a.Mean-a2.Mean, 2) < epsilon
 }
 
@@ -352,6 +350,5 @@ func (a *DistributionData) equal(other AggregationData) bool {
 			return false
 		}
 	}
-	epsilon := math.Pow10(-9)
 	return a.Count == a2.Count && a.Min == a2.Min && a.Max == a2.Max && math.Pow(a.Mean-a2.Mean, 2) < epsilon && math.Pow(a.variance()-a2.variance(), 2) < epsilon
 }
--- a/stats/aggregation_data.go
+++ b/stats/aggregation_data.go
@@ -93,10 +93,8 @@ func (a *SumData) addSample(v interface{}) {
 	switch x := v.(type) {
 	case int64:
 		f = float64(x)
-		break
 	case float64:
 		f = x
-		break
 	default:
 		return
 	}
@@ -155,10 +153,8 @@ func (a *MeanData) addSample(v interface{}) {
 	switch x := v.(type) {
 	case int64:
 		f = float64(x)
-		break
 	case float64:
 		f = x
-		break
 	default:
 		return
 	}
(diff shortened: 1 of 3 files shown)


stats/aggregation_data.go
```diff
@@ -74,6 +74,137 @@ func (a *CountData) equal(other AggregationData) bool {
...
+	if a.Count == 1 {
+		a.Mean = f
+		return
+	}
+	a.Mean = a.Mean + (f-a.Mean)/float64(a.Count)
+}
+
+// Only Mean will be mutiplied by the fraction, Count will remain the same.
+func (a *MeanData) multiplyByFraction(fraction float64) AggregationData {
+	return newMeanData(a.Mean*fraction, a.Count)
```
jcd2: I think we want it the other way around: multiply the count by the fraction and keep the mean the same.

songy23: That could cause some loss of precision, because ultimately we want a fractional sum (mean * count) so that we can combine and calculate the interval mean stats ((m1.sum + m2.sum) / (m1.count + m2.count)). Multiplying the count by the fraction will floor it.

songy23: Or if rounding is expected, I'm fine with changing this.

stats/view_test.go
```diff
@@ -594,6 +584,298 @@ func Test_View_MeasureFloat64_AggregationCount_WindowSlidingTime(t *testing.T) {
 	}
 }
 
+func Test_View_MeasureFloat64_AggregationSum_WindowCumulative(t *testing.T) {
+	k1, _ := tag.NewKey("k1")
+	k2, _ := tag.NewKey("k2")
+	k3, _ := tag.NewKey("k3")
+	agg1 := SumAggregation{}
```
rakyll: s/agg1/agg

rakyll: Do we need this as a variable, what about:

view, err := NewView("VF1", "desc VF1", []tag.Key{k1, k2}, nil, SumAggregation{}, Cumulative{})

stats/view_test.go
```diff
@@ -594,6 +584,298 @@ func Test_View_MeasureFloat64_AggregationCount_WindowSlidingTime(t *testing.T) {
...
+					[]tag.Tag{{Key: k1, Value: "v1 is a very long value key"}, {Key: k2, Value: "v2 is a very long value key"}},
+					newSumData(12),
+				},
+			},
+		},
+	}
+
+	for _, tc := range tcs {
+		view.clearRows()
+		view.subscribe()
```
rakyll: Use view.Subscribe instead. unexported subscribe is view internal, external users shouldn't depend on it.

stats/view_test.go
```diff
@@ -594,6 +584,298 @@ func Test_View_MeasureFloat64_AggregationCount_WindowSlidingTime(t *testing.T) {
...
+				{
+					[]tag.Tag{{Key: k1, Value: "v1 is a very long value key"}, {Key: k2, Value: "v2 is a very long value key"}},
+					newSumData(12),
+				},
+			},
+		},
+	}
+
+	for _, tc := range tcs {
+		view.clearRows()
```
rakyll: I'd unregister at the end of this test instead and clear the collected rows.

All unexported methods on view are for convince, only view tests should care about them and call them.

stats/view_test.go
```diff
@@ -594,6 +584,298 @@ func Test_View_MeasureFloat64_AggregationCount_WindowSlidingTime(t *testing.T) {
...
+			}
+		}
+	}
+}
+
+func Test_View_MeasureFloat64_AggregationMean_WindowCumulative(t *testing.T) {
+	k1, _ := tag.NewKey("k1")
+	k2, _ := tag.NewKey("k2")
+	k3, _ := tag.NewKey("k3")
+	agg1 := MeanAggregation{}
```
rakyll: ditto here

stats/view_test.go
```diff
@@ -594,6 +584,298 @@ func Test_View_MeasureFloat64_AggregationCount_WindowSlidingTime(t *testing.T) {
...
+				{
+					[]tag.Tag{{Key: k1, Value: "v1"}, {Key: k2, Value: "v2"}},
+					newSumData(5),
+				},
+			},
+		},
+		{
+			"4",
+			[]record{
+				{1, []tagString{{k1, "v1 is a very long value key"}}},
```
rakyll: Do we care about the validation of the long values in the scope of this test? This case seems to be a bit out of the place.

stats/aggregation_data.go
```diff
@@ -74,6 +74,137 @@ func (a *CountData) equal(other AggregationData) bool {
...
+
+func (a *SumData) isAggregate() bool { return true }
+
+func (a *SumData) addSample(v interface{}) {
+	// Both float64 and int64 values will be wrapped to float64
+	var f float64
+	switch x := v.(type) {
+	case int64:
+		f = float64(x)
+		break
```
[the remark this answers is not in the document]
songy23: Done. I copied this from https://github.com/census-instrumentation/opencensus-go/blob/master/stats/aggregation_data.go#L117, I've removed `break` there too.

stats/aggregation_data.go
```diff
@@ -74,6 +74,137 @@ func (a *CountData) equal(other AggregationData) bool {
...
+	case float64:
+		f = x
+		break
```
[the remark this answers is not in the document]
songy23: Done.

stats/aggregation_data.go
```diff
@@ -74,6 +74,137 @@ func (a *CountData) equal(other AggregationData) bool {
...
+func (a *MeanData) Sum() float64 { return a.Mean * float64(a.Count) }
+
+func (a *MeanData) isAggregate() bool { return true }
+
+func (a *MeanData) addSample(v interface{}) {
+	var f float64
+	switch x := v.(type) {
+	case int64:
+		f = float64(x)
+		break
```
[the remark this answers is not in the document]
songy23: Done.

stats/aggregation_data.go
```diff
@@ -74,6 +74,137 @@ func (a *CountData) equal(other AggregationData) bool {
...
+	case float64:
+		f = x
+		break
```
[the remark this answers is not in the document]
songy23: Done.

stats/aggregation_data.go
jcd2: If you multiply the mean by the fraction, the numerator of ((m1.sum + m2.sum) / (m1.count + m2.count)) will be correct but the denominator will be too high.  Multiplying the count by the fraction will fix both.

Let's also change Count to be a float64 to avoid the precision problem.

stats/aggregation_data.go
```diff
@@ -74,6 +74,133 @@ func (a *CountData) equal(other AggregationData) bool {
...
+
+func newSumData(v float64) *SumData {
+	tmp := SumData(v)
+	return &tmp
+}
+
+func (a *SumData) isAggregate() bool { return true }
+
+func (a *SumData) addSample(v interface{}) {
+	// Both float64 and int64 values will be wrapped to float64
```
odeke-em: Minor nit:
s/wrapped/cast/g

stats/aggregation_data.go
```diff
@@ -74,6 +74,133 @@ func (a *CountData) equal(other AggregationData) bool {
...
+	var f float64
+	switch x := v.(type) {
+	case int64:
+		f = float64(x)
+	case float64:
+		f = x
+	default:
+		return
+	}
+	*a = SumData(float64(*a) + f)
```
odeke-em: How about we modify this to:
```go
*a += SumData(f)
```

stats/aggregation_data.go
```diff
@@ -74,6 +74,133 @@ func (a *CountData) equal(other AggregationData) bool {
...
+	*a = 0
+}
+
+func (a *SumData) equal(other AggregationData) bool {
+	a2, ok := other.(*SumData)
+	if !ok {
+		return false
+	}
+
+	epsilon := math.Pow10(-9)
```
odeke-em: Minor nit, in the future, we can make this a const i.e
```go
const epsilon = 1e-9
```

stats/aggregation_data.go
```diff
@@ -74,6 +74,133 @@ func (a *CountData) equal(other AggregationData) bool {
...
+	a.Mean = 0
+}
+
+func (a *MeanData) equal(other AggregationData) bool {
+	a2, ok := other.(*MeanData)
+	if !ok {
+		return false
+	}
+
+	epsilon := math.Pow10(-9)
```
odeke-em: Ditto for this epsilon value, we can gut out this function call with the const epsilon.

stats/aggregation_data.go
songy23: Done.

stats/aggregation_data.go
songy23: Looks better, updated.

stats/aggregation_data.go
songy23: Done.

stats/aggregation_data.go
songy23: SGTM, updated.

stats/aggregation_data.go
songy23: Done.

songy23:
Thanks!

Issue #173, opened 2017-11-29
exporter/stats/stackdriver: key sanitization issues


jcd2:
Keys have to start with a letter, they can't start with a digit or underscore.  We could sanitize in those cases by adding "key_" or some other string at the front.

The key should be sanitized in the equalAggWindowTagKeys check -- otherwise when we read back the MetricDescriptor it won't appear to be compatible with the view.

A more subtle issue is that sanitizing two different tag keys in a view could result in the same label, which could cause the CreateMetricDescriptor and CreateTimeSeries calls to fail.  To fix it we'd have to use a sanitization scheme that applied to the whole set of tag keys and avoided duplicates.

rakyll:
Also, make sure the key size is still equal or below the allowed limit after sanitization.

odeke-em:
So for the checks

Check | Completed | Source Code
---|---|---
Keys have to start with a letter, they can't start with a digit or underscore. We could sanitize in those cases by adding "key_" or some other string at the front. | Yes | https://github.com/census-instrumentation/opencensus-go/blob/41d5f3842541ca8c85e96e15688ef0472d2daaf2/internal/sanitize.go#L33-L39
Keys should be sanitized in the equalAggWindow | Yes | https://github.com/census-instrumentation/opencensus-go/blob/41d5f3842541ca8c85e96e15688ef0472d2daaf2/exporter/stats/stackdriver/stackdriver.go#L494-L497
Also, make sure the key size is still equal or below the allowed limit after sanitization. | Yes | https://github.com/census-instrumentation/opencensus-go/blob/41d5f3842541ca8c85e96e15688ef0472d2daaf2/internal/sanitize.go#L30-L32
A more subtle issue is that sanitizing two different tag keys in a view could result in the same label, which could cause the CreateMetricDescriptor and CreateTimeSeries calls to fail. To fix it we'd have to use a sanitization scheme that applied to the whole set of tag keys and avoided duplicates.| No |

We are only left with one check on that wishlist of sanitization i.e.
`
A more subtle issue is that sanitizing two different tag keys in a view could result in the same label, which could cause the CreateMetricDescriptor and CreateTimeSeries calls to fail. To fix it we'd have to use a sanitization scheme that applied to the whole set of tag keys and avoided duplicates.
`
which we can do by perhaps a sanitization method that memoizes its previously sanitized keys and would be invoked as a first time key sanitization method, returning an error if something failed.

odeke-em:
[...]

PR #175, opened 2017-11-30
Sanitize tag keys when looking up for the matching metric descriptor

rakyll:
Updates #173.

### DIFF
--- a/exporter/stats/stackdriver/stackdriver_test.go
+++ b/exporter/stats/stackdriver/stackdriver_test.go
@@ -148,8 +148,8 @@ func TestExporter_makeReq(t *testing.T) {
 }
 
 func TestEqualAggWindowTagKeys(t *testing.T) {
-	key1, _ := tag.NewKey("test_key_one")
-	key2, _ := tag.NewKey("test_key_two")
+	key1, _ := tag.NewKey("test-key-one")
+	key2, _ := tag.NewKey("test-key-two")
 	tests := []struct {
 		name    string
 		md      *metricpb.MetricDescriptor
--- a/internal/sanitize.go
+++ b/internal/sanitize.go
@@ -24,6 +24,15 @@ const labelKeySizeLimit = 100
 // Sanitize returns a string that is trunacated to 100 characters if it's too
 // long, and replaces non-alphanumeric characters to underscores.
 func Sanitize(s string) string {
+	if len(s) == 0 {
+		return s
+	}
+	if unicode.IsDigit(rune(s[0])) {
+		s = "key_" + s
+	}
+	if s[0] == '_' {
+		s = "key" + s
+	}
 	if len(s) > labelKeySizeLimit {
 		s = s[:labelKeySizeLimit]
 	}
--- a/internal/sanitize_test.go
+++ b/internal/sanitize_test.go
@@ -37,8 +37,18 @@ func TestSanitize(t *testing.T) {
 		},
 		{
 			name:  "don't modify alphanumeric string",
-			input: "0123456789_ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz",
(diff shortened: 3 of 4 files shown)


exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -148,8 +148,8 @@ func TestExporter_makeReq(t *testing.T) {
 }
 
 func TestEqualAggWindowTagKeys(t *testing.T) {
-	key1, _ := tag.NewKey("test_key_one")
-	key2, _ := tag.NewKey("test_key_two")
+	key1, _ := tag.NewKey("test-key-one")
```
odeke-em: Nice update, to test that we actually sanitize.

rakyll:
Can you take another look? I also fixed the sanitization rule for the strings starting with digits or underscores.

internal/sanitize.go
```diff
@@ -24,6 +24,15 @@ const labelKeySizeLimit = 100
 // Sanitize returns a string that is trunacated to 100 characters if it's too
 // long, and replaces non-alphanumeric characters to underscores.
 func Sanitize(s string) string {
+	if len(s) == 0 {
+		return s
```
songy23: Should we also check empty string in stackdriver.go? I don't think Stackdriver can take empty as label key.

rakyll: We should check that at tag.NewKey. Validation in the tag package is not very complete, I am going to work on that later today.

internal/sanitize_test.go
```diff
@@ -37,8 +37,18 @@ func TestSanitize(t *testing.T) {
 		},
 		{
 			name:  "don't modify alphanumeric string",
-			input: "0123456789_ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz",
-			want:  "0123456789_ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz",
+			input: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_0123456789",
```
odeke-em: Perhaps this name also needs to change, as the precondition is now false.
We shall modify an alphanumeric string iff
the first key is either "_" or a digit

Maybe we can further update this to
```go
  {
    name: "don't modify alphanumeric string with letters as the prefix",
    input: " "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_0123456789",
    want: ""ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_0123456789",
},
```

Also is there a doc for StackDrive metric key sanitization?  I ask because for the past 3 days there have been many issues opened, each one coming with new rules so perhaps it's time for us to consolidate them in one place?

internal/sanitize.go
odeke-em: @songy23 you might be interested in seeing this issue https://github.com/census-instrumentation/opencensus-go/issues/171 in relation to empty label keys.

internal/sanitize_test.go
rakyll: I just named it "valid input". There is no doc that lists the limitations unfortunately :(

internal/sanitize.go
songy23: I see, thanks.

jcd2:
lgtm

internal/sanitize.go
```diff
@@ -24,6 +24,15 @@ const labelKeySizeLimit = 100
...
+	}
+	if unicode.IsDigit(rune(s[0])) {
+		s = "key_" + s
+	}
+	if s[0] == '_' {
```
jcd2: This misses the case where the string begins with a character that is not an underscore, but will be an underscore after escaping.

rakyll: https://github.com/census-instrumentation/opencensus-go/pull/186 is sent.

PR #178, opened 2017-11-30
Validate key tags and values

### LINKED ISSUE #171: tag.validateKeyName should reject 0-length keys

rakyll:
Validation rules are explained at
https://github.com/census-instrumentation/opencensus-specs/issues/12.

Fixes #171.

### DIFF
--- a/tag/map_codec_test.go
+++ b/tag/map_codec_test.go
@@ -22,10 +22,12 @@ import (
 	"testing"
 )
 
+var keys []Key
+
 func Test_EncodeDecode_Set(t *testing.T) {
 	k1, _ := NewKey("k1")
 	k2, _ := NewKey("k2")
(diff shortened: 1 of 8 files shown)


tag/map_codec_test.go
```diff
@@ -70,7 +72,7 @@ func Test_EncodeDecode_Set(t *testing.T) {
 				{k1, "v1"},
 				{k2, "v2"},
 				{k3, "v3"},
-				{k4, "v4 is very weird <>.,?/'\";:`~!@#$%^&*()_-+={[}]|\\"},
+				{k4, "v4_"},
```
jcd2: The values don't get escaped, so it's probably a good idea to keep the weird string we had here before.

rakyll: Done.

Issue #168, opened 2017-11-28
exporter/prometheus: implement one


odeke-em:
Assigned to me by @rakyll, this issue is a placeholder for that exporter.

rakyll:
Thanks for this tracking bug!

I was explaining the other day what the API may look like:

```
package prometheus

type Exporter struct {} ...

func (e *Exporter) Handler() http.Handler {}
```

odeke-em:
@rakyll I am going to close this issue since the  first version of the exporter was implemented. Any bugs that we find can be independently reported.

PR #181, opened 2017-11-30
Initial commit of the Prometheus exporter

rakyll:
odeke-em@ and I found a way to reuse the Prometheus Go package
to render the endpoint. This change is an initial commit with the APIs.

The exporter is heavily under construction.

Updates #168.

### DIFF
--- a/examples/stats/prometheus/main.go
+++ b/examples/stats/prometheus/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"math/rand"
 	"net/http"
 	"time"
 
@@ -65,8 +66,14 @@ func main() {
 		log.Fatalf("Cannot subscribe to the view: %v", err)
 	}
 
-	// Record data points.
-	stats.Record(ctx, videoSize.M(25648))
+	go func() {
+		ctx := context.Background()
+		for {
+			// Record some data points.
+			stats.Record(ctx, videoSize.M(rand.Int63()))
+			<-time.After(time.Millisecond * time.Duration(1+rand.Intn(400)))
+		}
+	}()
 
 	// Wait for a duration longer than reporting duration to ensure the stats
 	// library reports the collected data.
--- a/examples/stats/prometheus/main.go
+++ b/examples/stats/prometheus/main.go
@@ -39,17 +39,17 @@ func main() {
 
 	// Create measures. The program will record measures for the size of
 	// processed videos and the nubmer of videos marked as spam.
-	videoSize, err := stats.NewMeasureInt64("my.org/measures/video_size", "size of processed videos", "MBy")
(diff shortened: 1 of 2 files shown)


rakyll:
An example of the export endpoint:

export

exporter/stats/prometheus/prometheus.go
```diff
@@ -0,0 +1,114 @@
...
+	if e.opts.OnError != nil {
+		e.opts.OnError(err)
+		return
+	}
+	log.Printf("Failed to export to Prometheus: %v", err)
+}
+
+// Export exports to the Prometheus if view data has one or more rows.
+func (e *Exporter) Export(vd *stats.ViewData) {
+	if len(vd.Rows) > 0 {
```
odeke-em: I think we meant the opposite :)

exporter/stats/prometheus/prometheus.go
```diff
@@ -0,0 +1,114 @@
...
+		for _, t := range r.Tags {
+			labels = append(labels, internal.Sanitize(t.Key.Name()))
+		}
+		// Sanitize the view name.
+		desc := prometheus.NewDesc(view.Name(), view.Description(), labels, nil)
+		e.c.descCh <- desc
+
+		// TODO(jbd): Support other metric types.
+		switch v := r.Data.(type) {
+		case *stats.CountData:
```
odeke-em: This is awesome. PTAL at the one for DistributionAggregation

https://github.com/orijtech/opencensus-go/blob/d6201c4289f86c30fa85ba0e103f9ad1e201988d/exporter/stats/prometheus/prometheus.go#L188-L207

exporter/stats/prometheus/prometheus.go
rakyll: Oops, sorry. Fixed.

exporter/stats/prometheus/prometheus.go
rakyll: Neat 👍

examples/stats/prometheus/main.go
```diff
@@ -0,0 +1,77 @@
...
+	// Set reporting period to report data at every second.
+	stats.SetReportingPeriod(1 * time.Second)
+
+	// Subscribe will allow view data to be exported.
+	// Once no longer need, you can unsubscribe from the view.
+	if err := view.Subscribe(); err != nil {
+		log.Fatalf("Cannot subscribe to the view: %v", err)
+	}
+
+	// Record data points.
```
odeke-em: One other thing, could we get this to continually send/simulate data
```go
// Simulate transcoding going on in the background
	go func() {
		ctx := context.Background()
		for {
			// Record data points.
			stats.Record(ctx, videoSize.M(rand.Int63()))
			<-time.After(time.Millisecond * time.Duration(1+rand.Intn(450)))
		}
	}()
```
which will demo the whole time series concept and on refresh will produce new data.

PR #186, opened 2017-11-30
Never allow sanitized strings to start with underscore

rakyll:
The sanitization may replace the first character with underscore.
Append key prefix after sanitization is completed to make
sure the end result is not starting with an underscore.

### DIFF
--- a/internal/sanitize.go
+++ b/internal/sanitize.go
@@ -27,16 +27,17 @@ func Sanitize(s string) string {
 	if len(s) == 0 {
 		return s
 	}
+	if len(s) > labelKeySizeLimit {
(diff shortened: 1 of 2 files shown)


odeke-em:
Great find @jcd2 :)

internal/sanitize.go
```diff
@@ -27,16 +27,17 @@ func Sanitize(s string) string {
 	if len(s) == 0 {
 		return s
 	}
+	if len(s) > labelKeySizeLimit {
```
jcd2: This truncation still needs to happen last, because s might grow by 3 or 4 characters below.

PR #192, opened 2017-11-30
Set the default namespace if none is given

rakyll:
Previously, we removed a bunch of namespace related accessors but forgot to set the default if given is empty.

### DIFF
--- a/exporter/stats/prometheus/prometheus.go
+++ b/exporter/stats/prometheus/prometheus.go
@@ -55,6 +55,9 @@ type Options struct {
 
 // NewExporter returns an exporter that exports stats to Prometheus.
 func NewExporter(o Options) (*Exporter, error) {
+	if o.Namespace == "" {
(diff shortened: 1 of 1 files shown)


odeke-em:
We actually already have this code in https://github.com/census-instrumentation/opencensus-go/blob/a7bfe253256f6623febf0eda458222389ba5a1ce/exporter/stats/prometheus/prometheus.go#L270-L273

odeke-em:
Please use collector.namespace

odeke-em:
Could we instead make this line
https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/stats/prometheus/prometheus.go#L82
`c.namespace`?

odeke-em:
Or alternatively, please remove that code from newCollector so that we use your new change.

rakyll:
Chatted offline and I removed it from the newCollector.

PR #194, opened 2017-11-30
Report only the latest state

rakyll:
Drop all accumulated view data, given OpenCensus always reports
the latest state of the aggregation.


rakyll:
Thanks to @odeke-em for reporting this :)

Issue #183, opened 2017-11-30
stats: clearRows called incorrectly


jcd2:
At the end of reportUsage, clearRows is being called for views with Interval windows, but this means intervals will only contain data since the last export.  (And it's being called for Cumulative windows too, because the type assertion is *Cumulative instead of Cumulative.)

I think we should delete the call to clearRows for now, and add a TODO to implement a garbage-collection scheme instead.

odeke-em:
Am assuming we should tie these issues together https://github.com/census-instrumentation/opencensus-go/issues/184, that is #184 is the precursor to this one?

rakyll:
I agree that we should tie them, we can update https://github.com/census-instrumentation/opencensus-go/issues/184 with the details explained here and close this as a duplicate.

ramonza:
Going to leave this for the stats changes that @rakyll is working on

rakyll:
clearRows is only called to cleanup the collection for cumulative windows at unsubscription, we can close it now.

PR #195, opened 2017-11-30
Do not drop data from cumulative views

rakyll:
Updates #183.

### DIFF
--- a/stats/worker.go
+++ b/stats/worker.go
@@ -294,7 +294,12 @@ func (w *worker) reportUsage(now time.Time) {
 			e.Export(viewData)
 		}
 		exportersMu.Unlock()
-		if _, ok := v.Window().(*Cumulative); !ok {
+		switch v.Window().(type) {
+		case *Cumulative:
+			// noop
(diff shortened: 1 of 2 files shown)


rakyll:
Let me actually add a test for this.

stats/worker.go
```diff
@@ -294,7 +294,12 @@ func (w *worker) reportUsage(now time.Time) {
...
 		}
 		exportersMu.Unlock()
-		if _, ok := v.Window().(*Cumulative); !ok {
+		switch v.Window().(type) {
+		case *Cumulative:
+			// noop
+		case Cumulative:
+			// noop
+		default:
 			v.clearRows()
```
jcd2: I don't think we want to call clearRows at all -- otherwise if we have a one-minute window, it will only contain the last 10 seconds because we keep clearing it.

rakyll:
This PR is not fixing #183, just dealing with all cumulative windows correctly so we can have a demo. I am going to work on a fix later.

Issue #196, opened 2017-12-01
exporter/stats/stackdriver: error checking for GetMetricDescriptor


jcd2:
equalAggWindowTagKeys is called immediately after GetMetricDescriptor without checking if err is non-nil.  The call to equalAggWindowTagKeys can move into the "if err == nil" block below it instead.

Issue #203, opened 2017-12-01
exporter/stats/stackdriver: nil pointer reference


rakyll:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x142dcea]

goroutine 116 [running]:
go.opencensus.io/exporter/stats/stackdriver.equalAggWindowTagKeys(0x0, 0x15e4140, 0x18ab390, 0x15e4180, 0x18ab390, 0xc420180050, 0x3, 0x3, 0x15e0140, 0xc4202c47b0)
	/Users/jbd/src/go.opencensus.io/exporter/stats/stackdriver/stackdriver.go:359 +0x3a
go.opencensus.io/exporter/stats/stackdriver.(*Exporter).createMeasure(0xc4201987e0, 0x15e7740, 0xc4200260a8, 0xc420430140, 0x0, 0x0)
	/Users/jbd/src/go.opencensus.io/exporter/stats/stackdriver/stackdriver.go:228 +0x3cd
go.opencensus.io/exporter/stats/stackdriver.(*Exporter).upload(0xc4201987e0, 0xc420472010, 0x2, 0x2, 0x1, 0x148b640)
	/Users/jbd/src/go.opencensus.io/exporter/stats/stackdriver/stackdriver.go:150 +0x98
go.opencensus.io/exporter/stats/stackdriver.NewExporter.func1(0x148b640, 0xc4204620a0)
	/Users/jbd/src/go.opencensus.io/exporter/stats/stackdriver/stackdriver.go:107 +0x63
google.golang.org/api/support/bundler.(*Bundler).startFlushLocked.func1(0xc4201928c0, 0x148b640, 0xc4204620a0, 0x97, 0x2, 0xc420278000, 0x0)
	/Users/jbd/src/google.golang.org/api/support/bundler/bundler.go:256 +0xbf
created by google.golang.org/api/support/bundler.(*Bundler).startFlushLocked
	/Users/jbd/src/google.golang.org/api/support/bundler/bundler.go:245 +0x17f
```

jcd2:
This is probably #196

rakyll:
Yes, it is :) Dup of #196.

PR #205, opened 2017-12-01
Only call equalAggWindowTagKeys if MetricDescriptor exists

songy23:
Fix https://github.com/census-instrumentation/opencensus-go/issues/196


songy23:
Should also fix https://github.com/census-instrumentation/opencensus-go/issues/203

PR #211, opened 2017-12-02
Use stdlib context package for all non gRPC packages

### LINKED ISSUE #199: remove golang.org/x/net/context from all non gRPC packages

odeke-em:
For all non gRPC packages, import
  "context"
instead of
  "golang.org/x/net/context"

Fixes #199


odeke-em:
Thanks y'all.

PR #216, opened 2017-12-03
Remove Measure's dependency on Views

### LINKED ISSUE #198: stats: remove Measure's dependency on Views
Measure should not depend on view by implementing addView, etc. Views using a measure can be tracked via the worker.

odeke-em:
Removes the need for each Measure having to implement:
* addView
* removeView
* viewsCount

by tracking which views use a Measure with a map
keyed by Measure, values *Views and tracked by the worker.

Fixes #198

### DIFF
--- a/stats/measure_int64.go
+++ b/stats/measure_int64.go
@@ -20,7 +20,6 @@ type MeasureInt64 struct {
 	name        string
 	unit        string
 	description string
-	views       map[*View]bool
 }
 
 // Name returns the name of the measure.
@@ -38,15 +37,7 @@ func (m *MeasureInt64) Unit() string {
 	return m.unit
 }
 
-func (m *MeasureInt64) addView(v *View) {
-	m.views[v] = true
-}
-
-func (m *MeasureInt64) removeView(v *View) {
-	delete(m.views, v)
-}
-
-func (m *MeasureInt64) viewsCount() int { return len(m.views) }
+var _ Measure = (*MeasureInt64)(nil)
 
 // M creates a new int64 measurement.
 // Use Record to record measurements.
--- a/stats/measure.go
+++ b/stats/measure.go
@@ -29,10 +29,6 @@ type Measure interface {
 	Name() string
 	Description() string
 	Unit() string
-
-	addView(v *View)
-	removeView(v *View)
-	viewsCount() int
 }
 
 // Measurement is the numeric value measured when recording stats. Each measure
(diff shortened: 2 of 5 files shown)


stats/measure_int64.go
```diff
@@ -38,15 +37,7 @@ func (m *MeasureInt64) Unit() string {
...
-func (m *MeasureInt64) addView(v *View) {
-	m.views[v] = true
-}
-
-func (m *MeasureInt64) removeView(v *View) {
-	delete(m.views, v)
-}
-
-func (m *MeasureInt64) viewsCount() int { return len(m.views) }
+var _ Measure = (*MeasureInt64)(nil)
```
rakyll: Can we also do it for MeasureFloat64?

rakyll:
And please fix the broken tests.

stats/measure_int64.go
odeke-em: Roger that, I'll do that too and fix the tests, thanks!

jcd2:
This seems fine (although slightly slower) but I'm not sure why it's better than what we had.  Is it just to make the code neater under the hood?

odeke-em:
@jcd2 yes in deed, for simpler code under the hood. Also could you share some insight as to the speed(slower) of the change or is this a hypothesis?

jcd2:
> @jcd2 yes in deed, for simpler code under the hood. Also could you share some insight as to the speed(slower) of the change or is this a hypothesis?

It's just that there's an extra map lookup when recording: recordReq.addSamplesForMeasure has to look up the measure in the map w.measureToViews, whereas before the map of views was in the measure itself.

odeke-em:
@jcd2 gotcha, the new benefit is less code duplication too.

rakyll:
Can we fix the tests and merge this? I will try to split views and aggregation into a different package and need this to be in.

odeke-em:
This change introduced a weird issue where it seems like Measures with the supposedly same key are not going into the same map entry, I haven't yet wrapped my head around it but I went through each call site to ensure parity with the old code.

odeke-em:
Alright, am taking another stab at this.

odeke-em:
@rakyll did you want to switch scenes for a little and take a look at this, I've spent too much on this PR and figured I could expend that time instead on other issues to get them out the way. I'll mark this as WIP and digest on the new findings as we knock out the other issues.

rakyll:
No worries, I can take a look.

rakyll:
Dup of https://github.com/census-instrumentation/opencensus-go/pull/260.

PR #218, opened 2017-12-03
travis: update to 1.9.x go version

dvrkps:


odeke-em:
Thank you for the PR @dvrkps!
I think we still have the latest version in the travis.yml file at 1.8 because
are still on Go1.8 on Google App Engine,  hence switching to 1.9 would mean that if
things broke for 1.8 we couldn't easily catch them from test failures.

However, I'll let those with context advise /cc @songy23 @rakyll @zombiezen.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.8.x
+  - 1.9.x
```
rakyll: Can we have both? 1.8 and 1.9?

rakyll:
Also, please do not prefix the commit message. Start with capital, "Add Go 1.9.x for Travis".

I will add a CI check for the commit message formatting soon.

.travis.yml
dvrkps: @rakyll done

rakyll:
Thanks much, @dvrkps!

PR #219, opened 2017-12-03
Sum, Mean Aggregations support for Prometheus exporter

### LINKED ISSUE #215: exporter/stats/prometheus: implement the sum and mean aggregation support for cumulative window
Implement the Prometheus export support for the sum and mean aggregations. We will initially only implement the cumulative window support for these aggregations.

odeke-em:
Fixes #215


examples/stats/prometheus/main.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"go.opencensus.io/exporter/stats/prometheus"
 	"go.opencensus.io/stats"
```
rakyll: Please revert the changes to the example. We want to keep the small and not want to demonstrate every possible aggregation.

exporter/stats/prometheus/prometheus.go
```diff
@@ -213,11 +213,17 @@ func (c *collector) toMetric(desc *prometheus.Desc, view *stats.View, row *stats
 		}
 		return hist, nil
 
-	case stats.SumAggregation:
-		panic("stats.SumData not supported yet")
+	case stats.MeanAggregation:
+		data := row.Data.(*stats.MeanData)
+		hist, err := prometheus.NewConstHistogram(desc, uint64(data.Count), data.Sum(), nil, tagValues(row.Tags)...)
```
rakyll: Just return prometheus.NewConstHistogram(desc, uint64(data.Count), data.Sum(), nil, tagValues(row.Tags)...)

rakyll: This should be a gauge value, isn't it? We do not have the distribution buckets to create a useful histogram metric type.

odeke-em: My bad, I wonder why I even thought so, fixing this in a second.

examples/stats/prometheus/main.go
odeke-em: Roger that, I can do that. I guess I've perhaps send a follow up PR that uses the aggregations in tests.

odeke-em:
Awesome, thank you for the review @rakyll!

PR #208, opened 2017-12-01
Support sum and mean for Stackdriver exporter

songy23:
Fix https://github.com/census-instrumentation/opencensus-go/issues/180

PR #224, opened 2017-12-04
Use DistributionValue for MeanData, check Sum and Mean in equalAggWindowTagKeys

songy23:
As a follow up of https://github.com/census-instrumentation/opencensus-go/pull/208

### DIFF
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -313,6 +313,14 @@ func newTypedValue(view *stats.View, r *stats.Row) *monitoringpb.TypedValue {
 				Count: int64(v.Count),
 				Mean:  v.Mean,
 				SumOfSquaredDeviation: 0,
+				BucketOptions: &distributionpb.Distribution_BucketOptions{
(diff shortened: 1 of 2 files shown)


exporter/stats/stackdriver/stackdriver.go
```diff
@@ -308,8 +308,12 @@ func newTypedValue(view *stats.View, r *stats.Row) *monitoringpb.TypedValue {
 			DoubleValue: float64(*v),
 		}}
 	case *stats.MeanData:
-		return &monitoringpb.TypedValue{Value: &monitoringpb.TypedValue_DoubleValue{
-			DoubleValue: v.Mean,
+		return &monitoringpb.TypedValue{Value: &monitoringpb.TypedValue_DistributionValue{
+			DistributionValue: &distributionpb.Distribution{
```
jcd2: You will also need to set BucketOptions and BucketCounts, similar to the DistributionData case below.

Bounds can just be {0} and BucketCounts {0, int64(v.Count)}.

songy23: I see, updated.

PR #217, opened 2017-12-03
Fix Aggregation types for Views in plugin/grpc/grpcstats

odeke-em:
Fix up the aggregation types for the various stats.Views
in plugin/grpc/grpcstats to ensure they conform to those
from the reference Java implementation as per
https://github.com/census-instrumentation/opencensus-java/blob/2b464864e3dd3f80e8e4c9dc72fccc225444a939/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViewConstants.java#L113-L658

Also added a test that at runtime ensures that the types match
up what we expect lest we get errors such as

```shell
=== RUN   TestViewsAggregationsConform
--- FAIL: TestViewsAggregationsConform (0.00s)
  client_metrics_test.go:33: Item on line: 40 got stats.DistributionAggregation want stats.MeanAggregation
  client_metrics_test.go:33: Item on line: 41 got stats.DistributionAggregation want stats.MeanAggregation
  client_metrics_test.go:33: Item on line: 42 got stats.DistributionAggregation want stats.MeanAggregation
FAIL
exit status 1
FAIL  go.opencensus.io/plugin/grpc/grpcstats  0.011s
```

Fixes #210

PR #225, opened 2017-12-04
Fix accidental View name typo and add tests for plugin/grpc/grpcstats

odeke-em:
Reported by @jcd2, I accidentally typo'd
  " rpc.io/client/error_count/cumulative"
instead of
  "grpc.io/client/error_count/cumulative"
in PR #217.

Added a test to ensure that we never regress/repeat this,
because the code has many moving parts and such a typo
or copy and paste errors would have lurked if not for
a sharp eye or an explicit error.


odeke-em:
Cool, thank you @songy23 for the review!

PR #226, opened 2017-12-05
Only support stats.Cumulative window for Prometheus

### LINKED ISSUE #214: exporter/stats/prometheus: disable Interval window
We currently do not support windows other than the cumulative window. Return an error saying "Interval windows are not supported for now".

odeke-em:
If a View doesn't have a stats.Cumulative
window at export time, skip collecting it.

Fixes #214

### DIFF
--- a/exporter/stats/prometheus/prometheus.go
+++ b/exporter/stats/prometheus/prometheus.go
@@ -72,9 +72,20 @@ func NewExporter(o Options) (*Exporter, error) {
 var _ http.Handler = (*Exporter)(nil)
 var _ stats.Exporter = (*Exporter)(nil)
 
+func allowedWindowType(v *stats.View) bool {
+	// TODO: (@rakyll, @odeke-em): Only the cumulative window will
+	// be exported in this version. Support others in the future.
+	// See Issue https://github.com/census-instrumentation/opencensus-go/issues/214
+	_, ok := v.Window().(stats.Cumulative)
+	return ok
+}
+
 func (c *collector) registerViews(views ...*stats.View) {
 	count := 0
 	for _, view := range views {
+		if !allowedWindowType(view) {
+			continue
+		}
 		sig := viewSignature(c.opts.Namespace, view)
 		c.registeredViewsMu.Lock()
 		_, ok := c.registeredViews[sig]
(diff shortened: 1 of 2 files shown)


exporter/stats/prometheus/prometheus.go
```diff
@@ -170,14 +170,14 @@ func (c *collector) Describe(ch chan<- *prometheus.Desc) {
...
 // for example when the HTTP endpoint is invoked by Prometheus.
 func (c *collector) Collect(ch chan<- prometheus.Metric) {
-	c.mu.Lock()
-	views := make(map[string]*stats.ViewData, len(c.viewData))
-	for i, vd := range c.viewData {
-		views[i] = vd
-	}
-	c.mu.Unlock()
-
 	for _, vd := range c.viewData {
```
rakyll: Can we reject them at Export instead?

odeke-em: I was following suit with https://github.com/census-instrumentation/opencensus-go/blob/9b9df23b5a299ba044805cdd2c28764702cfddab/exporter/stats/stackdriver/stackdriver.go#L168-L172 and the suggestion from the issue originally, but for sure, it even makes it easier to reject them in the Export method.

odeke-em:
@rakyll I've also added non Cumulative View rejection in Export by proxy of method addView. I extracted the code into a helper and we perform that check at two different call sites:
* Collect
* Export
PTAL

exporter/stats/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,82 @@
...
+				View: newView(stats.CountAggregation{}, stats.Cumulative{}),
+				Rows: []*stats.Row{
+					{nil, &count1},
+				},
+			},
+			want: 1,
+		},
+		2: {
+			vds: &stats.ViewData{
+				View: newView(stats.CountAggregation{}, fauxWindow{}),
```
rakyll: Just use Interval window instead.

odeke-em: Nice, thanks. Updating that in a second.

odeke-em:
Thank you for the review @rakyll, I addressed your feedback, tests passing and I got your blessing with a LGTM, so I'll merge.

PR #229, opened 2017-12-07
trace: improve doc

jba:
Add a package-level comment that has a couple of standard use cases.

Add an example.

### DIFF
--- a/trace/examples_test.go
+++ b/trace/examples_test.go
@@ -15,16 +15,27 @@
 package trace_test
 
 import (
-	"context"
+	"fmt"
 
 	"go.opencensus.io/trace"
+	"golang.org/x/net/context"
 )
 
 // This example shows how to use StartSpan and EndSpan to capture
 // a function execution in a Span. It assumes that the function
 // has a context.Context argument.
-func ExampleStartSpan(ctx context.Context) {
-	ctx = trace.StartSpan(ctx, "my/package.Function")
-	defer trace.EndSpan(ctx)
-	// TODO: body of function
+func ExampleStartSpan() {
+	printEvens := func(ctx context.Context) {
+		ctx = trace.StartSpan(ctx, "my/package.Function")
+		defer trace.EndSpan(ctx)
+
+		for i := 0; i < 10; i++ {
+			if i%2 == 0 {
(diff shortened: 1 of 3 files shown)


trace/examples_test.go
```diff
@@ -0,0 +1,30 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trace_test
+
+import (
+	"context"
```
odeke-em: Could we use "golang.org/x/net/context" for as much compatibility with Go1.6?

trace/examples_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"go.opencensus.io/trace"
+)
+
+// This example shows how to use StartSpan and EndSpan to capture
+// a function execution in a Span. It assumes that the function
+// has a context.Context argument.
+func ExampleStartSpan(ctx context.Context) {
```
odeke-em: Just a suggestion to follow the standard Go examples, could we perhaps make this
```go
func ExampleStartSpan() {
    printEvens := func(ctx context.Context) {
         ctx = trace.StartSpan(ctx, "my/package.Function")
         defer trace.EndSpan(ctx)
         
         for i := 0; i < 10; i++ {
             if i%2 == 0 {
                fmt.Printf("Even!\n")
             }
         }
    }
    ctx := context.Background()
    printEvens(ctx)
}
```
Please feel free to push back if my example looks complex but I just wanted to minimize passing a context into an exported example.

odeke-em: Actually Travis CI complained that the example be niladic thus shouldn't take any arguments https://travis-ci.org/census-instrumentation/opencensus-go/jobs/313030043#L690

trace/examples_test.go
jba: Done.

trace/examples_test.go
jba: Done.

Issue #197, opened 2017-12-01
trace: SetDefaultSampler should set default sampler for nil


rakyll:
There is no way to revert back to the default sampler once user sets a custom sampler. SetDefaultSampler can use the default sample if nil sampler is given.

rakyll:
Closed via https://github.com/census-instrumentation/opencensus-go/pull/231.

PR #231, opened 2017-12-11
Update SetDefaultSampler to revert to default sampler

goller:
Hey there!  I was looking through the issues for one that would get me into the code and found #197 

I'm not 100% sure this is what is needed, but, it's fun to get in and see how it is built!


rakyll:
Thanks for the PR! We haven't agreed this should be the expected behavior yet but @jcd2 can provide feedback.

PR #236, opened 2017-12-13
Add test for Exporter.createMeasure

rakyll:
This test will be followed up with a test asserts the protos.

### DIFF
--- a/exporter/stats/stackdriver/stackdriver_test.go
+++ b/exporter/stats/stackdriver/stackdriver_test.go
@@ -581,13 +581,13 @@ func TestExporter_createMeasure(t *testing.T) {
 		t.Errorf("Exporter.createMeasure() error = %v", err)
 	}
 	if count := getCalls; count != 1 {
-		t.Errorf("getMetricDescriptor needs to be called for twice; called %v times", count)
-	}
-	if count := createCalls; count != 1 {
 		t.Errorf("getMetricDescriptor needs to be called for once; called %v times", count)
 	}
+	if count := createCalls; count != 1 {
+		t.Errorf("createMetricDescriptor needs to be called for once; called %v times", count)
+	}
 	if count := len(e.createdViews); count != 1 {
-		t.Errorf("len(e.createdViews) = %v; want &v", count)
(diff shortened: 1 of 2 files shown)


exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -525,6 +527,70 @@ func TestEqualAggWindowTagKeys(t *testing.T) {
...
+	}
+
+	ctx := context.Background()
+	if err := e.createMeasure(ctx, vd); err != nil {
+		t.Errorf("Exporter.createMeasure() error = %v", err)
+	}
+	if err := e.createMeasure(ctx, vd); err != nil {
+		t.Errorf("Exporter.createMeasure() error = %v", err)
+	}
+	if count := getCalls; count != 1 {
```
songy23: s/count != 1/count != 2?

rakyll: No, 1. Because we will cache the MetricDescriptor in the createdViews map. We don't do an RPC each time we call createMeasure.

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -525,6 +527,70 @@ func TestEqualAggWindowTagKeys(t *testing.T) {
...
+		t.Errorf("getMetricDescriptor needs to be called for twice; called %v times", count)
```
songy23: s/twice/once

exporter/stats/stackdriver/stackdriver_test.go
songy23: I see

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -525,6 +527,70 @@ func TestEqualAggWindowTagKeys(t *testing.T) {
...
+	}
+	if count := createCalls; count != 1 {
+		t.Errorf("getMetricDescriptor needs to be called for once; called %v times", count)
```
songy23: s/getMetricDescriptor/createMetricDescriptor

PR #240, opened 2017-12-13
Remove use of unsafe

jcd2:
Also change length in the internal format to a single byte, since value lengths
are now restricted to 255.

### DIFF
--- a/internal/tagencoding/tagencoding.go
+++ b/internal/tagencoding/tagencoding.go
@@ -17,12 +17,6 @@
 // used interally by the stats collector.
 package tagencoding
 
-import (
-	"unsafe"
-)
-
-var sizeOfUint16 = (int)(unsafe.Sizeof(uint16(0)))
-
 type Values struct {
 	Buffer     []byte
 	WriteIndex int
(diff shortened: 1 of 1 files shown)


internal/tagencoding/tagencoding.go
```diff
@@ -38,14 +32,11 @@ func (vb *Values) growIfRequired(expected int) {
 }
 
 func (vb *Values) WriteValue(v []byte) {
-	length := len(v)
-	vb.growIfRequired(sizeOfUint16 + length)
+	length := len(v) & 0xff
```
rakyll: length := byte(len(v))

jcd2: That's not clearly any better to me -- length is used as an int on lines 36 and 49, so your way is 3 type conversions instead of 1.  :)

odeke-em: Did we instead mean to use `0xffff` uint16 and not uint8?

internal/tagencoding/tagencoding.go
```diff
@@ -38,14 +32,11 @@ func (vb *Values) growIfRequired(expected int) {
...
+	vb.growIfRequired(1 + length)
 
 	// writing length of v
-	bytes := *(*[2]byte)(unsafe.Pointer(&length))
-	vb.Buffer[vb.WriteIndex] = bytes[0]
-	vb.WriteIndex++
-	vb.Buffer[vb.WriteIndex] = bytes[1]
+	vb.Buffer[vb.WriteIndex] = byte(length)
```
odeke-em: This doesn't seem right though as we are writing into the buffer a max of 0xffff, unless am mistaken.

odeke-em: To keep parity with the old code, we need two different vb.WriteIndex additions or just one but not truncating length to 0xff only, please advise on this.

internal/tagencoding/tagencoding.go
jcd2: I changed the encoding from uint16 to uint8, since we restrict tag values to length 255 now.

(Which I remembered when I tried to test this code with values longer than 255.  :))

odeke-em: Oops I missed the description and jumped straight into code, @rakyll corrected me offline. My apologies.

odeke-em: @jcd2 gotcha gotcha, thanks for the PR.

PR #242, opened 2017-12-13
Add an exporter which sends trace spans to a Zipkin server

jcd2:


exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+		case trace.MessageEventTypeSent:
+			return model.Client
+		case trace.MessageEventTypeRecv:
+			return model.Server
+		}
+	}
+	return model.Undetermined
+}
+
+func zipkinSpan(s *trace.SpanData, localEndpoint *model.Endpoint) model.SpanModel {
```
rakyll: I'd add a few test cases specifically for this function.

exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+// zipkinURL is the address of the Zipkin server.
+//
+// localEndpoint sets the local endpoint of exported spans.  It can be
+// constructed with github.com/openzipkin/zipkin-go.NewEndpoint, e.g.:
+// 	localEndpoint, err := NewEndpoint("my server", listener.Addr().String())
+// localEndpoint can be nil.
+//
+// opts can be used to pass options to the underlying Zipkin http reporter
+// library.
+func NewExporter(zipkinURL string, localEndpoint *model.Endpoint, opts ...http.ReporterOption) (*Exporter, error) {
```
rakyll: zipkinURL can default to the localhost:[defaultport]

exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+// 	import (
+// 		zk "github.com/openzipkin/zipkin-go"
+// 		"go.opencensus.io/trace/adaptor/zipkin"
+// 	)
+//	...
+//		localEndpoint, err := zk.NewEndpoint("my server", myHostPort)
+// 		if err != nil {
+// 			log.Print(err)
+// 		}
+// 		exporter, err := zipkin.NewExporter("http://localhost:9411/api/v2/spans", localEndpoint)
```
rakyll: Should we should ask for the full path or the host:port?

Do we plan to support multiple API versions? @adriancole what are your thoughts?

exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package zipkin contains an exporter for Zipkin.
+//
+// Example:
+//
+// 	import (
+// 		zk "github.com/openzipkin/zipkin-go"
```
adriancole: nit I might avoid zk as usually that means zookeeper. if zipkin is not too long, I'd use that.

exporter/trace/zipkin/zipkin.go
adriancole: the endpoint url is something some mess with, for example, proxy-mounting it, etc. I think default encoding format should be v2, and when needed a separate option (v3 isn't likely for a year or two.. not even planned)

exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+// 		"go.opencensus.io/trace/adaptor/zipkin"
+// 	)
+//	...
+//		localEndpoint, err := zk.NewEndpoint("my server", myHostPort)
```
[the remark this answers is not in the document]
adriancole: nit: spaces are ok, but probably best to use single-words. The canonical zipkin example has two services named "frontend" and "backend". The endpoint should default to link-local address if possible to look that up.

exporter/trace/zipkin/zipkin.go
adriancole: we usually make the zipkinUrl fully qualified due to people doing it over https or remounting it, or clones

exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+
+func convertSpanID(s trace.SpanID) model.ID {
+	return model.ID(binary.BigEndian.Uint64(s[:]))
+}
+
+func spanKind(s *trace.SpanData) model.Kind {
+	if s.HasRemoteParent {
+		return model.Server
+	}
+	if strings.HasPrefix(s.Name, "Sent.") {
```
adriancole: presumably there's no way to tell if messaging is in use (SPMC type stuff)

exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+	z := model.SpanModel{
+		SpanContext: model.SpanContext{
+			TraceID: convertTraceID(sc.TraceID),
+			ID:      convertSpanID(sc.SpanID),
+			Sampled: &sampledTrue,
+		},
+		Kind:          spanKind(s),
+		Name:          s.Name,
+		Timestamp:     s.StartTime,
+		Shared:        false,
```
adriancole: unless required, you can leave this out as shared defaults to false and json will be smaller.

exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+			Timestamp: a.Time,
+			Value:     a.Message,
+		})
+	}
+	for _, m := range s.MessageEvents {
+		a := model.Annotation{
+			Timestamp: m.Time,
+		}
+		switch m.EventType {
+		case trace.MessageEventTypeSent:
```
adriancole: if sent/receive are about span.kind, I'd leave it out and just substitute span.timestamp/duration. If they are enclosed and not the same, then leave in.

exporter/trace/zipkin/zipkin_test.go
```diff
@@ -0,0 +1,132 @@
...
+					Timestamp: time.Time{},
+					Value:     "SENT",
+				},
+			},
+			Tags: map[string]string{
+				"stringkey":                 "value",
+				"intkey":                    "42",
+				"boolkey":                   "true",
+				"census.status_code":        "INVALID_ARGUMENT",
+				"census.status_description": "error",
```
adriancole: error should be a top-level tag as it affects the UI. ex "error" -> "INVALID_ARGUMENT"

exporter/trace/zipkin/codes.go
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package zipkin
+
+type code int
+
+// Canonical status code enum.
+const (
+	OK code = iota
```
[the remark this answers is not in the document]
jcchavezs: +1

exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+// 	localEndpoint, err := NewEndpoint("my server", listener.Addr().String())
+// localEndpoint can be nil.
+//
+// opts can be used to pass options to the underlying Zipkin http reporter
+// library.
+func NewExporter(zipkinURL string, localEndpoint *model.Endpoint, opts ...http.ReporterOption) (*Exporter, error) {
+	return &Exporter{
+		reporter:      http.NewReporter(zipkinURL, opts...),
+		localEndpoint: localEndpoint,
+	}, nil
```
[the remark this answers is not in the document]
jcchavezs: Apparently none but the interface returns an error.

exporter/trace/zipkin/codes.go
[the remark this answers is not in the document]
basvanbeek: Looks like C/C++ constant formatting this way, exporting being a side effect of the formatting choice.

I'd prefer them more Go idiomatic `failedPrecondition` instead of `FAILED_PRECONDITION`

exporter/trace/zipkin/zipkin.go
basvanbeek: Currently `zipkin-go` supports HTTP V2 only. However I expect to shortly have a Kafka reporter as well, which will be the preferred exporter for high throughput. Other potential stream transports for the (near) future include rabbit. The `zipkin.NewExporter` function only handles the HTTP case.

Are we going to have additional methods to handle these (and maybe rename this function to `NewHTTPExporter` or would it make sense to adjust the parameter list for the `NewExporter` function to handle alternative transports?

exporter/trace/zipkin/zipkin.go
basvanbeek: Actually it will not have effect on the JSON payload because the `omitempty` Go tag is set and false is the boolean default/zero value.

I do like it being mentioned here so it is explicit that OpenCensus does not support Zipkin's shared span feature.

exporter/trace/zipkin/zipkin_test.go
```diff
@@ -0,0 +1,132 @@
...
+			TraceID:      trace.TraceID{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16},
+			SpanID:       trace.SpanID{17, 18, 19, 20, 21, 22, 23, 24},
+			TraceOptions: 1,
+		},
+		Name:      "name",
+		StartTime: now,
+		EndTime:   now.Add(24 * time.Hour),
+		Attributes: map[string]interface{}{
+			"stringkey": "value",
+			"intkey":    int64(42),
```
basvanbeek: Douglas Adams will be proud of you :)

exporter/trace/zipkin/zipkin_test.go
```diff
@@ -0,0 +1,132 @@
...
+	})
+	var data []byte
+	select {
+	case data = <-ch:
+	case <-time.After(2 * time.Second):
+		t.Fatalf("span was not exported")
+	}
+	// alter some fields that are custom-encoded by model.SpanModel.MarshalJSON so
+	// that we can unmarshal them.
+	data = regexp.MustCompile(`"timestamp":[0-9]*`).ReplaceAll(data, []byte(`"timestamp":"2006-01-02T15:04:05+07:00"`))
```
basvanbeek: You can remove the regex lines as model.SpanModel also has custom Unmarshal logic to rebuild from JSON.

exporter/trace/zipkin/zipkin_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	got[0].Annotations[0].Timestamp = time.Time{}
+	want := []model.SpanModel{
+		model.SpanModel{
+			SpanContext: model.SpanContext{
+				TraceID: model.TraceID{
+					High: 0x102030405060708,
+					Low:  0x90a0b0c0d0e0f10,
+				},
+				ID: 0x1,
```
basvanbeek: This is not the ID sent but the result of the regex manipulation. Use `0x1112131415161718` here just like TraceID is properly constructed'

exporter/trace/zipkin/zipkin_test.go
```diff
@@ -0,0 +1,132 @@
...
+			},
+			Name:      "name",
+			Kind:      "CLIENT",
+			Timestamp: time.Time{},
+			Duration:  86400000000,
```
basvanbeek: Go duration is in nanoseconds... Zipkin uses microseconds. with the model unmarshal logic it will correcty unmarshal the microsecond representation into proper Go duration. You can use `24 * time.Hour` here

exporter/trace/zipkin/zipkin.go
jcd2: No.  Maybe in the future we'll add kinds to opencensus spans.

exporter/trace/zipkin/zipkin.go
jcd2: Done.

exporter/trace/zipkin/zipkin.go
jcd2: Yes this line is a no-op which I left in to be explicit.

exporter/trace/zipkin/codes.go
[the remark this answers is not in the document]
jcd2: This was so the string function could be autogenerated by the stringer tool.  Changed to use an array.

exporter/trace/zipkin/zipkin.go
jcd2: Changed to openzipkin.

exporter/trace/zipkin/zipkin.go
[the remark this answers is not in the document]
jcd2: Done.

exporter/trace/zipkin/zipkin.go
jcd2: In that case I'll change the exporter to take a reporter parameter, then it should be compatible with new reporter types automatically.

exporter/trace/zipkin/zipkin.go
jcd2: The http reporter requires it to be fully qualified.

exporter/trace/zipkin/zipkin.go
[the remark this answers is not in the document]
jcd2: This was to leave open the possibility that reporter construction could return an error in the future.  Since reporter construction is now done by the user, I'll delete the error return value.

exporter/trace/zipkin/zipkin.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// Export exports a span to a Zipkin server.
+func (e *Exporter) Export(s *trace.SpanData) {
+	e.reporter.Send(zipkinSpan(s, e.localEndpoint))
+}
+
+const (
+	statusCodeTagKey        = "census.status_code"
```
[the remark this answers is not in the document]
jcd2: Done (but status_code -> "error").  I'll file an issue for the Java version to change.

exporter/trace/zipkin/zipkin.go
jcd2: This is to create annotations for individual message events, if they're in the census span.

exporter/trace/zipkin/zipkin_test.go
jcd2: I hope he'd think I know where my towel is.  :)

exporter/trace/zipkin/zipkin_test.go
jcd2: Now it does.  :)  Changed.

exporter/trace/zipkin/zipkin_test.go
jcd2: Done.

exporter/trace/zipkin/zipkin_test.go
jcd2: Done.

exporter/trace/zipkin/zipkin_test.go
jcd2: Done.  I now have peach-coloured error spans.  :)

odeke-em:
Nice! I've updated the wiki entry here https://github.com/census-instrumentation/opencensus-go/wiki/Implemented-exporters.

PR #244, opened 2017-12-13
Record error status code not the error message

### LINKED ISSUE #222: plugin/grpc/grpcstats: error tag values
The tag value for rpc status is the full text of the error, but it should just be a string of the error code.

rakyll:
Also make the status key name is consistent with other language implementations.

Fixes #222.

### DIFF
--- a/plugin/grpc/grpcstats/client_handler.go
+++ b/plugin/grpc/grpcstats/client_handler.go
@@ -25,6 +25,7 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/stats"
+	"google.golang.org/grpc/status"
 )
 
 // ClientStatsHandler is a stats.Handler implementation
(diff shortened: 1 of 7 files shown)


plugin/grpc/grpcstats/client_handler.go
```diff
@@ -160,12 +161,14 @@ func (h *ClientStatsHandler) handleRPCEnd(ctx context.Context, s *stats.End) {
...
 
 	if s.Error != nil {
-		errorCode := s.Error.Error()
-		newTagMap, err := tag.NewMap(ctx,
-			tag.Upsert(keyOpStatus, errorCode),
-		)
-		if err == nil {
-			ctx = tag.NewContext(ctx, newTagMap)
+		s, ok := status.FromError(s.Error)
+		if ok {
```
odeke-em: Could we keep the old path for cases where FromError returns ok: false?
```go
var errorCode string
if s, ok := status.FromError(s.Error); ok {
   errorCode = s.Code().String()
} else {
   errorCode = s.Error().Error()
}
newTagMap, err := tag.NewMap(ctx, tag.Upsert(keyStatus, errorCode))
if err != nil {
   ctx = tag.NewContext(ctx, newTagMap)
}
```

plugin/grpc/grpcstats/grpcstats.go
```diff
@@ -65,21 +65,21 @@ var (
...
+		log.Fatalf("Cannot create service key: %v", err)
 	}
-	if keyMethod, err = tag.NewKey("grpc.method"); err != nil {
-		log.Fatalf("Cannot create grpc.method key: %v", err)
+	if keyMethod, err = tag.NewKey("method"); err != nil {
+		log.Fatalf("Cannot create method key: %v", err)
 	}
-	if keyOpStatus, err = tag.NewKey("grpc.opstatus"); err != nil {
-		log.Fatalf("Cannot create grpc.opstatus key: %v", err)
+	if keyStatus, err = tag.NewKey("canonical_status"); err != nil {
```
odeke-em: How come the change here? I guess this is from what other libs do eh?

rakyll: It is the key Java implementation uses.

plugin/grpc/grpcstats/client_handler.go
rakyll: We don't do it because a free formatted message is not a dimension.

plugin/grpc/grpcstats/grpcstats.go
odeke-em: Gotcha, thanks.

plugin/grpc/grpcstats/client_handler.go
odeke-em: But we did use it before though, that's a fallback.

PR #247, opened 2017-12-13
Handle overflow and oversized items in exporter/stats/stackdriver bundler

### LINKED ISSUE #202: exporter/stats/stackdriver: handle overflow and oversized items when adding to bundler

odeke-em:
Handle overflow and oversized items in exporter/stats/stackdriver bundler

After a bundler addition, on encountering:

* bundler.ErrOversizedItem

individually upload each item

In the case of:
* bundler.ErrOverflow
do not try to reupload it, but instead log the failure

The above two behaviors mimick what the trace exporter does in
https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/trace/stackdriver/stackdriver.go#L131

Fixes #202

### DIFF
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -104,9 +104,7 @@ func NewExporter(o Options) (*Exporter, error) {
 	}
 	e.bundler = bundler.NewBundler((*stats.ViewData)(nil), func(bundle interface{}) {
 		vds := bundle.([]*stats.ViewData)
-		if err := e.upload(vds); err != nil {
-			e.onError(err)
-		}
+		e.handleUpload(vds...)
 	})
 	e.bundler.DelayThreshold = e.o.BundleDelayThreshold
 	e.bundler.BundleCountThreshold = e.o.BundleCountThreshold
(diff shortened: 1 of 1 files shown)


exporter/stats/stackdriver/stackdriver.go
```diff
@@ -119,7 +117,23 @@ func (e *Exporter) Export(vd *stats.ViewData) {
...
+		return
+	case bundler.ErrOversizedItem, bundler.ErrOverflow:
+		go e.handleUpload(vd)
+	default:
+		e.onError(err)
+	}
+}
+
+// handleUpload handles uploading
+// a slice of ViewData as well a  error handling.
```
songy23: nit: as well a -> as well as

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -119,7 +117,23 @@ func (e *Exporter) Export(vd *stats.ViewData) {
 	if len(vd.Rows) == 0 {
 		return
 	}
-	e.bundler.Add(vd, 1)
+	err := e.bundler.Add(vd, 1)
+	switch err {
+	case nil:
+		return
+	case bundler.ErrOversizedItem, bundler.ErrOverflow:
+		go e.handleUpload(vd)
```
rakyll: It might be a bit aggressive to do this on ErrOverflow. In the case of lack of resources, we might not be consuming the bundler fast enough. By creating more coroutines, we will contribute to the problem.

odeke-em: Hmm, interesting. Would you recommend instead that we pause for a couple of seconds in the case of ErrOverflow, or should we just drop it?

rakyll: Dropping and logging it sounds reasonable.

odeke-em:
@songy23 @rakyll I've updated it, PTAL.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -119,7 +117,25 @@ func (e *Exporter) Export(vd *stats.ViewData) {
...
 	}
-	e.bundler.Add(vd, 1)
+	err := e.bundler.Add(vd, 1)
+	switch err {
+	case nil:
+		return
+	case bundler.ErrOversizedItem:
+		go e.handleUpload(vd)
+	case bundler.ErrOverflow:
+		e.onError(errors.New("OpenCensus stats/Stackdriver exporter: failed to upload View: buffer full"))
```
rakyll: "failed to upload: buffer full"

onError will automatically add a Stackdriver exporter to the log.

odeke-em: Aye aye, thanks. Let me update that.

odeke-em:
Changes update. Thank you @songy23 and @rakyll for the reviews!

PR #248, opened 2017-12-13
Start time should be the time collection first happened for cum views

### LINKED ISSUE #200: stats: start time in ViewData
For views with a cumulative window, this should be the time from which the data has been accumulated.

rakyll:
Fixes #200.

### DIFF
--- a/stats/worker.go
+++ b/stats/worker.go
@@ -30,10 +30,9 @@ func init() {
 }
 
 type worker struct {
-	measuresByName map[string]Measure
-	measures       map[Measure]bool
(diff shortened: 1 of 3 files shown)


songy23:
Please fix the tests.

odeke-em:
Yeah the previous test was a little racy in combination with the other tests but on its own it passed
```shell
$ go test -race -run=ReportUsage
PASS
ok  	go.opencensus.io/stats	1.329s
```

rakyll:
I fixed the race.

Issue #166, opened 2017-11-28
exporter/stackdriver: Support custom MonitoredResource labels


songy23:
For now the Stackdriver Stats Exporter will always use a global MonitoredResource with no associated labels: https://cloud.google.com/monitoring/api/resources#tag_global.

In the next version we may consider allowing users to set custom labels (e.g GCE instance-id) for their MonitoredResource in the exporter.

odeke-em:
Milestone moved since @songy23 is rocking this already with a PR.

PR #249, opened 2017-12-13
Support custom MonitoredResource in Stackdriver Exporter Options

songy23:
Fixes https://github.com/census-instrumentation/opencensus-go/issues/166.

This is required if we want to allow multiple processes exporting to the same metric on Stackdriver Monitoring. More background here: https://github.com/census-instrumentation/opencensus-java/pull/870.

### DIFF
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -88,8 +88,8 @@ type Options struct {
 	// Optional.
 	BundleCountThreshold int
 
-	// Resource is the Stackdriver MonitoredResource representing a
-	// resource that can be used for monitoring.
+	// Resource is an optional field that represents the Stackdriver
+	// MonitoredResource, a resource that can be used for monitoring.
 	// If no custom ResourceDescriptor is set, a default MonitoredResource
 	// with type global and no resource labels will be used.
 	// Optional.
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -87,6 +87,13 @@ type Options struct {
 	// can be buffered before batch uploading them to the backend.
 	// Optional.
 	BundleCountThreshold int
+
+	// Resource is the Stackdriver MonitoredResource representing a
+	// resource that can be used for monitoring.
+	// If no custom ResourceDescriptor is set, a default MonitoredResource
+	// with type global and no resource labels will be used.
+	// Optional.
+	Resource *monitoredrespb.MonitoredResource
 }
 
 // NewExporter returns an exporter that uploads stats data to Stackdriver Monitoring.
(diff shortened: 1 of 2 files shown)


exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -591,6 +591,119 @@ func TestExporter_createMeasure(t *testing.T) {
...
+	end := start.Add(time.Minute)
+	count1 := stats.CountData(10)
+	count2 := stats.CountData(16)
+
+	resource := &monitoredrespb.MonitoredResource{
+		Type:   "gce_instance",
+		Labels: map[string]string{"instance_id": "instance", "zone": "us-west-1a"},
+	}
+
+	tests := []struct {
```
rakyll: Can we also test the case if no custom resource is provided?

songy23: This should already be covered in the previous tests (TestExporter_makeReq), since they are using the default resource.

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -591,6 +591,119 @@ func TestExporter_createMeasure(t *testing.T) {
...
+	key, err := tag.NewKey("test_key")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	cumView, err := stats.NewView("cumview", "desc", []tag.Key{key}, m, stats.CountAggregation{}, stats.Cumulative{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := stats.RegisterView(cumView); err != nil {
```
odeke-em: Could we make this consistent with cumView.Unregister below by using
```go
cumView.Subscribe()
```
and then below:
```go
defer cumView.Unsubscribe()
```

I'll actually file a bug to either add stats.UnregisterView or to remove stats.RegisterView entirely.

songy23: Sure, updated. I didn't update other tests though, because if I do so I'll get an error saying "cannot register measure "test-measure"; another measure with the same name is already registered".

odeke-em:
Thanks @songy23, I've filed https://github.com/census-instrumentation/opencensus-go/issues/250 for consistency of the API.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -87,6 +87,13 @@ type Options struct {
 	// can be buffered before batch uploading them to the backend.
 	// Optional.
 	BundleCountThreshold int
+
+	// Resource is the Stackdriver MonitoredResource representing a
```
odeke-em: Just a doc suggestion, could we spell out upfront that Resource is optionally, say:
```go
// Resource is an optional field that represents the Stackdriver MonitoredResource
```

songy23: Sure, done.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -87,6 +87,13 @@ type Options struct {
 	// can be buffered before batch uploading them to the backend.
 	// Optional.
 	BundleCountThreshold int
+
+	// Resource is an optional field that represents the Stackdriver
+	// MonitoredResource, a resource that can be used for monitoring.
+	// If no custom ResourceDescriptor is set, a default MonitoredResource
+	// with type global and no resource labels will be used.
+	// Optional.
```
odeke-em: Oops, we forgot to remove this, apologies, I should have spelled that out in my previous request :)

odeke-em: ^since we now explicitly spell out that Resource is an optional field, or I guess that's the style of the repo?

songy23: Oh I saw other fields also have this "Optional" word, do you mean we need to remove these?

odeke-em: Nah, it's alright then, LGTM.

songy23:
Thanks!

PR #251, opened 2017-12-14
Remove *View.Unregister, add stats.RegisterView for API consistency

### LINKED ISSUE #250: stats: add UnregisterView and then remove *View.Unregister for parity of APIs
I was just reviewing https://github.com/census-instrumentation/opencensus-go/pull/249/files and noticed that the stats API has
stats.RegisterView
but no matching call to stats.UnregisterView

and that instead the caller has to do *View.Unregister(), of which there also isn't a match method *View.Register()

Notice that RegisterView and *View.UnregisterView look like the perfect match
https://github.com/census-instrumentation/opencensus-go/blob/86bea64327bd7d353334ff858a2e6675a3122b7b/stats/worker.go#L115-L122
and
https://github.com/census-instrumentation/opencensus-go/blob/86bea64327bd7d353334ff858a2e6675a3122b7b/stats/worker.go#L127-L134

I'd say let's deprecate/remove *View.Unregister and instead add the UnregisterView helper since
we already have *View.Subscribe and *View.Unsubscribe of which Subscribe and Unsubscribe logically use RegisterView and UnregisterView respectively.

odeke-em:
Fixes #250

### DIFF
--- a/internal/readme/stats.go
+++ b/internal/readme/stats.go
@@ -99,7 +99,7 @@ func statsExamples() {
 	_ = v
 
 	// START unregisterView
-	if v.Unregister(); err != nil {
+	if stats.UnregisterView(v); err != nil {
 		log.Fatal(err)
 	}
 	// END unregisterView
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ Unregister view:
 
 [embedmd]:# (stats.go unregisterView)
 ```go
-if v.Unregister(); err != nil {
+if stats.UnregisterView(v); err != nil {
 	log.Fatal(err)
 }
 ```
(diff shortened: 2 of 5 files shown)


internal/readme/stats.go
```diff
@@ -99,7 +99,7 @@ func statsExamples() {
 	_ = v
 
 	// START unregisterView
-	if v.Unregister(); err != nil {
+	if stats.UnregisterView(v); err != nil {
```
songy23: Can you also run embedmd to re-generate the root README file?

odeke-em: Ah yes, thanks for that catch. Let me do so.

odeke-em: @songy23 done, PTAL.

odeke-em:
Thanks y'all for the review!

PR #252, opened 2017-12-14
Add a opencensus_task label to Stackdriver exporter

songy23:
A supplementary change to https://github.com/census-instrumentation/opencensus-go/pull/249.

**Background**: sometimes using a MonitoredResource is not enough to uniquely identify stats. For example, if there are multiple exporters running on the same GCE instance, they will all have the same MonitoredResource. If they export stats to the same Stackdriver metric, there will still be a race condition. To fix that, @jcd2 suggest to have a reserved `opencensus_task` label that can further separate the stats. The value of this label will be in the format "go-pid@hostname".

Equivalent changes for Java: https://github.com/census-instrumentation/opencensus-java/pull/885.

### DIFF
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -102,6 +102,7 @@ type Options struct {
 }
 
 // NewExporter returns an exporter that uploads stats data to Stackdriver Monitoring.
+// Only one Stackdriver exporter should be created per process.
 func NewExporter(o Options) (*Exporter, error) {
 	opts := append(o.ClientOptions, option.WithUserAgent(internal.UserAgent))
 	client, err := monitoring.NewMetricClient(context.Background(), opts...)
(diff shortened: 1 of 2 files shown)


exporter/stats/stackdriver/stackdriver.go
```diff
@@ -136,6 +142,16 @@ func (e *Exporter) Export(vd *stats.ViewData) {
 	}
 }
 
+// getTaskValue returns a task label value in the format of
+// "go-<pid>@<hostname>".
+func getTaskValue() string {
```
rakyll: This assumes you can only run one Stackdriver exporter per process.

Add a note about that?

songy23: That's right, actually in Java we enforce exporter to be singleton: https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java#L261

I added a note about this for NewExporter().

rakyll:
LGTM but we should enforce a singleton instead of allowing users to create new exporters. Want to file an issue?

songy23:
Sure, I filed https://github.com/census-instrumentation/opencensus-go/issues/258 to keep track of this.

PR #254, opened 2017-12-14
Represent mean and sum as untyped values

rakyll:
Gauges are a special type of value and their value is
determined by running a function that calculates the
some value and reports on the gauge usually periodically.

Sum and mean are not gauges but untyped simple metrics.

### DIFF
--- a/exporter/stats/prometheus/prometheus.go
+++ b/exporter/stats/prometheus/prometheus.go
@@ -219,11 +219,11 @@ func (c *collector) toMetric(desc *prometheus.Desc, view *stats.View, row *stats
 
 	case stats.MeanAggregation:
 		data := row.Data.(*stats.MeanData)
-		return prometheus.NewConstMetric(desc, prometheus.GaugeValue, data.Mean, tagValues(row.Tags)...)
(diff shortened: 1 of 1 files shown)


odeke-em:
So my thinking was that Gauge values are for measurements that go arbitrarily go up or down, and they can be used for measured values such as mean, sum, number of goroutines, please see https://prometheus.io/docs/concepts/metric_types/
screen shot 2017-12-14 at 12 00 19 am

Am just trying to understand Prometheus more :)

rakyll:
Their docs are not doing a good job explaining what a gauge is and how it differs. A gauge is a special type of measurement that can go up and down and its value is determined by running a function periodically when scrape/export is happening. Their docs suggest a gauge to have:

- A way to track in-progress requests in some piece of code/function.
- A way to time a piece of code and set the gauge to its duration in seconds. 

https://prometheus.io/docs/instrumenting/writing_clientlibs/#gauge

OpenCensus doesn't have support for gauges right now. We should not export a value to Prometheus as a gauge if they are not true gauges.

odeke-em:
Alright, thanks for that update. LGTM!

PR #260, opened 2017-12-14
Reduce cyclic dependencies between Measures and Views

rakyll:
Fixes #216.


odeke-em:
I believe this PR fixed https://github.com/census-instrumentation/opencensus-go/issues/198

PR #262, opened 2017-12-15
Fix stats doc typo

odeke-em:
Noticed while reading through the godoc https://godoc.org/go.opencensus.io/stats#RegisterExporter
screen shot 2017-12-15 at 2 21 43 pm


odeke-em:
Thank you @rakyll!

PR #264, opened 2017-12-15
Enforce singleton on Stackdriver's NewExporter per process

### LINKED ISSUE #258: exporter/stats/stackdriver: enforce singleton on exporter per process
Each process should have at most one running Stackdriver exporter.

odeke-em:
For stats
stackdriver packages, NewExporter now implements the singleton
per unique ProjectID so that any subsequent invocations will return an error.
The purpose of fixing this bug is to avoid race conditions
on StackDriver monitoring when multiple exporters are sending
over the same data to the same ProjectID initialized instance of StackDriver.

Fixes #258.

### DIFF
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -104,14 +104,10 @@ type Options struct {
 
 // Enforces the singleton on NewExporter per projectID per process
 // lest there will be races with Stackdriver.
-var seenProjectIDs struct {
-	m map[string]bool
-	sync.Mutex
-}
-
-func init() {
-	seenProjectIDs.m = make(map[string]bool)
-}
+var (
+	seenProjectsMu sync.Mutex
+	seenProjects   = make(map[string]bool)
+)
 
 var (
 	errBlankProjectID    = errors.New("expecting a non-blank ProjectID")
@@ -126,16 +122,14 @@ func NewExporter(o Options) (*Exporter, error) {
 		return nil, errBlankProjectID
 	}
 
-	seenProjectIDs.Lock()
-	_, seen := seenProjectIDs.m[o.ProjectID]
-	seenProjectIDs.Unlock()
+	seenProjectsMu.Lock()
+	defer seenProjectsMu.Unlock()
+	_, seen := seenProjects[o.ProjectID]
 	if seen {
 		return nil, errSingletonExporter
 	}
 
-	seenProjectIDs.Lock()
-	seenProjectIDs.m[o.ProjectID] = true
-	seenProjectIDs.Unlock()
+	seenProjects[o.ProjectID] = true
 
 	opts := append(o.ClientOptions, option.WithUserAgent(internal.UserAgent))
 	client, err := monitoring.NewMetricClient(context.Background(), opts...)
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -28,6 +28,7 @@ import (
 	"path"
 	"reflect"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -112,12 +113,19 @@ func init() {
 	seenProjectIDs.m = make(map[string]bool)
 }
 
-var errSingletonExporter = errors.New("only one exporter can be created per GCP ID per process")
+var (
+	errBlankProjectID    = errors.New("expecting a non-blank ProjectID")
+	errSingletonExporter = errors.New("only one exporter can be created per unique ProjectID per process")
+)
 
 // NewExporter returns an exporter that uploads stats data to Stackdriver Monitoring.
-// Only one Stackdriver exporter should be created per GCP ID per process, any subsequent
-// invocations of NewExporter will return an error.
+// Only one Stackdriver exporter should be created per ProjectID per process, any subsequent
+// invocations of NewExporter with the same ProjectID will return an error.
 func NewExporter(o Options) (*Exporter, error) {
+	if strings.TrimSpace(o.ProjectID) == "" {
+		return nil, errBlankProjectID
+	}
+
 	seenProjectIDs.Lock()
 	_, seen := seenProjectIDs.m[o.ProjectID]
 	seenProjectIDs.Unlock()
--- a/exporter/stats/stackdriver/stackdriver.go
+++ b/exporter/stats/stackdriver/stackdriver.go
@@ -101,9 +101,34 @@ type Options struct {
 	Resource *monitoredrespb.MonitoredResource
 }
 
+// Enforces the singleton on NewExporter per projectID per process
+// lest there will be races with Stackdriver.
+var seenProjectIDs struct {
+	m map[string]bool
+	sync.Mutex
+}
+
+func init() {
+	seenProjectIDs.m = make(map[string]bool)
+}
+
+var errSingletonExporter = errors.New("only one exporter can be created per GCP ID per process")
+
 // NewExporter returns an exporter that uploads stats data to Stackdriver Monitoring.
-// Only one Stackdriver exporter should be created per process.
+// Only one Stackdriver exporter should be created per GCP ID per process, any subsequent
+// invocations of NewExporter will return an error.
 func NewExporter(o Options) (*Exporter, error) {
+	seenProjectIDs.Lock()
+	_, seen := seenProjectIDs.m[o.ProjectID]
+	seenProjectIDs.Unlock()
+	if seen {
+		return nil, errSingletonExporter
+	}
+
+	seenProjectIDs.Lock()
+	seenProjectIDs.m[o.ProjectID] = true
+	seenProjectIDs.Unlock()
+
 	opts := append(o.ClientOptions, option.WithUserAgent(internal.UserAgent))
 	client, err := monitoring.NewMetricClient(context.Background(), opts...)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


exporter/stats/stackdriver/stackdriver.go
```diff
@@ -101,9 +101,25 @@ type Options struct {
 	Resource *monitoredrespb.MonitoredResource
 }
 
+// Enforces the singleton on NewExporter per process
+// lest there will be races with Stackdriver.
+// See Issue https://github.com/census-instrumentation/opencensus-go/issues/258
```
rakyll: You don't have to give a link to the issue here. The commit message for this change is going to mention the issue, we don't have to put links in the source code.

exporter/trace/stackdriver/stackdriver.go
```diff
@@ -31,6 +31,7 @@ package stackdriver
 
 import (
 	"context"
+	"errors"
```
rakyll: Trace is not affected by this problem. It is only the stats exporter.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -101,9 +101,25 @@ type Options struct {
...
+var newExporterOnce sync.Once
+var errSingletonExporter = errors.New("only one exporter can be created per process")
+
 // NewExporter returns an exporter that uploads stats data to Stackdriver Monitoring.
-// Only one Stackdriver exporter should be created per process.
+// Only one Stackdriver exporter should be created per process, any subsequent
```
rakyll: We should allow multiple exporters if the project ID is different. We should only return an error if the user is trying to initiate a new exporter for the same project ID.

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -34,6 +34,25 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// Ensure only one exporter per process, any
+// subsequent invocations of NewExporter should fail.
+// See Issue https://github.com/census-instrumentation/opencensus-go/issues/258
```
rakyll: Here as well.

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -34,6 +34,25 @@ import (
...
+func TestNewExporterSingletonPerProcess(t *testing.T) {
+	exp, err := NewExporter(Options{})
+	if err != nil {
+		t.Fatalf("the first initialization should pass successfully")
```
rakyll: t.Fatalf("NewExporter() = %v", err)

exporter/stats/stackdriver/stackdriver.go
odeke-em: Cool, I'll remove that.

exporter/stats/stackdriver/stackdriver.go
odeke-em: Oh okay, thanks for the clarification.

exporter/stats/stackdriver/stackdriver_test.go
odeke-em: Roger that, thanks!

exporter/trace/stackdriver/stackdriver.go
odeke-em: How come trace isn't affected?

exporter/stats/stackdriver/stackdriver_test.go
odeke-em: Roger that, thanks!

exporter/trace/stackdriver/stackdriver.go
rakyll: Stackdriver Monitoring has a requirement that the timelines need to be uploaded in a particular order. That makes us always upload data in chronological order. If one exporter uploads data earlier than the other one, it will be rejected by the API endpoint.

Stackdriver Trace doesn't have the same restriction.

songy23:
You might want to use a fake Google Credentials with a fake AccessToken for tests: https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporterTest.java#L42. I'm not sure how to do that in Go though.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -101,9 +102,41 @@ type Options struct {
...
+	errBlankProjectID    = errors.New("expecting a non-blank ProjectID")
+	errSingletonExporter = errors.New("only one exporter can be created per unique ProjectID per process")
+)
+
 // NewExporter returns an exporter that uploads stats data to Stackdriver Monitoring.
-// Only one Stackdriver exporter should be created per process.
+// Only one Stackdriver exporter should be created per ProjectID per process, any subsequent
+// invocations of NewExporter with the same ProjectID will return an error.
 func NewExporter(o Options) (*Exporter, error) {
+	if strings.TrimSpace(o.ProjectID) == "" {
```
rakyll: seenProjectIDs.Lock()
defer seenProjectIDs.Unlock()

We dont care holding this mutex in the lifetime of NewExporter. This function will be called very few times.

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -101,9 +102,41 @@ type Options struct {
 	Resource *monitoredrespb.MonitoredResource
 }
 
+// Enforces the singleton on NewExporter per projectID per process
+// lest there will be races with Stackdriver.
+var seenProjectIDs struct {
+	m map[string]bool
+	sync.Mutex
```
rakyll: var seenProjectIDs struct {
        sync.Mutex
 	m map[string]struct{}
}

exporter/stats/stackdriver/stackdriver.go
```diff
@@ -101,9 +102,41 @@ type Options struct {
...
+}
+
+func init() {
+	seenProjectIDs.m = make(map[string]bool)
```
rakyll: I'd prefer us to remove seenProjectIDs type. Then initialization will be at the same line:

var (
seenProjectMu sync.Mutex
seenProjects = make(map[string]struct{})
)

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -34,6 +34,42 @@ import (
...
 )
 
+func TestRejectBlankProjectID(t *testing.T) {
+	ids := []string{"", "     ", " "}
+	for _, projectID := range ids {
+		opts := Options{ProjectID: projectID}
+		exp, err := NewExporter(opts)
+		if err == nil || exp != nil {
+			t.Errorf("%q ProjectID must be rejected: NewExporter() = %v err = %q",
+				projectID, exp, err)
```
rakyll: Keep projectID, exp, and err on the previous line.

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -34,6 +34,42 @@ import (
...
+		exp, err := NewExporter(opts)
+		if err != nil {
+			t.Errorf("NewExporter() projectID = %q err = %q", projectID, err)
+			continue
+		}
+		if exp == nil {
+			t.Errorf("NewExporter returned a nil Exporter")
+			continue
+		}
+		for i := 0; i < 10; i++ {
```
rakyll: You don't need to iterate for 10 times.

Just try to call NewExporter once more.

exporter/stats/stackdriver/stackdriver.go
odeke-em: Great point, thank you.

exporter/stats/stackdriver/stackdriver.go
odeke-em: True true.

exporter/stats/stackdriver/stackdriver_test.go
odeke-em: It was a run off line, but sure, I can def update that. Thanks.

odeke-em:
Oh cool, thanks for pointing that out @songy23! Yes I'll pass in some custom credentials.

exporter/stats/stackdriver/stackdriver_test.go
```diff
@@ -24,16 +24,53 @@ import (
...
 	"google.golang.org/genproto/googleapis/api/metric"
 	metricpb "google.golang.org/genproto/googleapis/api/metric"
 	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
+var authOptions = []option.ClientOption{option.WithGRPCConn(&grpc.ClientConn{})}
```
odeke-em: @songy23 since you had mentioned/asked, here is how you can do it in Go, just create a fake gRPC client connection and that'll avoid looking for default credentials.

odeke-em:
Thanks y'all for the ideas and reviews!

exporter/stats/stackdriver/stackdriver_test.go
songy23: I see, it's somewhat different from Stackdriver client in Java. Thanks!

odeke-em: Here is what would mirror the Java code you posted
```go
const ts string
func (ts) Token() (*oauth2.Token, error) {
    return &Token{AccessToken: "fakeToken"}, nil
}
option.WithTokenTokenSource(ts{})
```
but notice how involved that is than `option.WithGRPCConn`, plus it requires an extra import of "golang.org/x/oauth2"

PR #268, opened 2017-12-18
Enforce singleton for Exporter per instance

odeke-em:
As per
https://prometheus.io/docs/instrumenting/writing_exporters/#deployment
we shouldn't have more than one exporter running per instance


odeke-em:
Thank you @songy23!

PR #271, opened 2017-12-18
Validate Key,Value pairs in seed Map for tag.NewContext

### LINKED ISSUE #269: tag: NewMap should validate Key and Value elements added
There is currently a TODO in tag/NewMap
https://github.com/census-instrumentation/opencensus-go/blob/84ea724b0b3b3f6d2efb36b27347511d8821e079/tag/map.go#L148-L149

Let's implement this validation.

odeke-em:
Fixes #269

Validate the Key, Value pairs from the seed *Map that was
propagated in the input Context.

### DIFF
--- a/tag/map.go
+++ b/tag/map.go
@@ -93,7 +93,7 @@ func Insert(k Key, v string) Mutator {
 	return &mutator{
 		fn: func(m *Map) (*Map, error) {
 			if !checkValue(v) {
-				return nil, errInvalid
+				return nil, errInvalidValue
 			}
 			m.insert(k, v)
 			return m, nil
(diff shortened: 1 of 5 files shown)


tag/map.go
```diff
@@ -146,11 +146,16 @@ func Delete(k Key) Mutator {
 // NewMap returns a new tag map originated from the incoming context
 // and modified with the provided mutators.
 func NewMap(ctx context.Context, mutator ...Mutator) (*Map, error) {
-	// TODO(jbd): Implement validation of keys and values.
 	m := newMap(0)
 	orig := FromContext(ctx)
 	if orig != nil {
 		for k, v := range orig.m {
+			if !checkKeyName(k.Name()) {
```
songy23: nit: shall we check for nil before this?

odeke-em: Would be a nice check although it can't be nil though as that's a `Key{}` not `*Key{}`

songy23: Ahh that's right, LGTM then.

odeke-em:
Thanks y'all!

PR #274, opened 2017-12-21
/exporter/trace/zipkin fix example

tw1nk:
The example used an incorrect import path: `go.opencensus.io/trace/adaptor/zipkin`
It should be: `go.opencensus.io/exporter/trace/zipkin`


tw1nk:
I signed it!

rakyll:
Thanks much!

PR #280, opened 2017-12-24
Fix syntax for view.Subscribe example.

wcn3:


wcn3:
r:rakyll

odeke-em:
Thank you for the catch @wcn3!

So to update the README, it's an undocumented process that involves the following:
1. Instead of editing the README itself, please edit internal/readme/README.md
2. Please run these instructions https://github.com/census-instrumentation/opencensus-go/tree/master/internal/readme

wcn3:
Done. It looks like I've also picked up some other changes to the README.

odeke-em:
Just one more request: so for the changes that got added to the README, could you please propagate them back to the internal like you did for the snippets, then regenerate them? Ideally only changes from internal/readme/README.md should appear in README.md.

wcn3:
Done. PTAL.

PR #282, opened 2017-12-29
Tweak wording of StartSpan for an already existent span

odeke-em:
The previous doc seemed to imply that a child span would
be created if "the child span" already existed but
we meant to say that the child span would only be created
if there was an already existent span with the same name.

### DIFF
--- a/trace/doc.go
+++ b/trace/doc.go
@@ -46,8 +46,8 @@ lines to the top of the function:
     ctx = trace.StartSpan(ctx, "your choice of name")
     defer trace.EndSpan(ctx)
 
-Invoking StartSpan with the name of an existing span will create a child span.
-Otherwise it will create a new top-level span.
+StartSpan will create a new top-level span if the context
+doesn't contain another span, otherwise it will create a child span.
 
 As a suggestion, use the fully-qualified function name as the span name, e.g.
 "github.com/me/mypackage.Run".
--- a/trace/doc.go
+++ b/trace/doc.go
@@ -46,8 +46,8 @@ lines to the top of the function:
     ctx = trace.StartSpan(ctx, "your choice of name")
     defer trace.EndSpan(ctx)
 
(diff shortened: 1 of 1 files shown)


trace/doc.go
```diff
@@ -46,8 +46,8 @@ lines to the top of the function:
     ctx = trace.StartSpan(ctx, "your choice of name")
     defer trace.EndSpan(ctx)
 
-StartSpan will create a child span if one already exists, and will create a
-new top-level span otherwise.
+Invoking StartSpan with the name of an already existent span will create a child span,
```
jba: s/already existent/existing/

Period after "child span".

odeke-em: Thanks @jba, I'll update this in a second.

trace/doc.go
```diff
@@ -46,8 +46,8 @@ lines to the top of the function:
     ctx = trace.StartSpan(ctx, "your choice of name")
     defer trace.EndSpan(ctx)
 
-StartSpan will create a child span if one already exists, and will create a
-new top-level span otherwise.
+Invoking StartSpan with the name of an existing span will create a child span.
```
rakyll: Is this true? It shouldn't have anything to do with the span name. It should create a child span if there is a span in the current context.

rakyll: I'd say, StartSpan will create a new top-level span if the context doesn't contain another span.

Otherwise, it creates a child.

odeke-em: Wow, what was I thinking, for some reason I thought I had read through code that reused the parent if there existed a parent in the context but also if the name was the same. Sorry about that, and thank you @rakyll for this correction.

odeke-em:
Awesome, thank you!

rakyll:
Can you rebase this? The Travis build is stuck.

odeke-em:
Sorry, hadn't seen this. For sure, let me do this.

PR #283, opened 2017-12-29
Implement a SanitizeNoClash that errors on same result but different inputs

odeke-em:
Updates #173

A sanitizer that returns an error if the input produces a sanitization
that clashes with a previous input that's lexicographically different
from the current one.

For example:

Sanitize("_0123?foo")    = "key_0123_foo"
Sanitize("key_0123?foo") = "key_0123_foo"
Sanitize("0123_foo")	 = "key_0123_foo"

and all the above clash so it returns an error telling the caller
what and where the clash is.

### DIFF
--- a/internal/sanitize.go
+++ b/internal/sanitize.go
@@ -15,7 +15,9 @@
 package internal
 
 import (
+	"fmt"
 	"strings"
+	"sync"
 	"unicode"
 )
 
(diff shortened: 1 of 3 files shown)


internal/sanitize.go
```diff
@@ -40,6 +42,32 @@ func Sanitize(s string) string {
 	return s
 }
 
+var (
+	smu       sync.Mutex
+	sanitized = make(map[string]string)
+)
+
+// SanitizeNoClash returns an error if in's sanitization clashes
```
rakyll: We should modify the sanitized string rather than returning an error. At least, we should try to do a few iterations and trying to find a properly sanitized combination.

I'd not spend much time on this PR right now, the sanitization strategy should be implemented similarly at all other languages for consistency and requires a spec.

ramonza:
ping @odeke-em

odeke-em:
Thanks for the ping @ramonza, I believe that we are waiting on a spec for the sanitization strategy hence this PR is on hold.

PR #284, opened 2017-12-31
tag: remove unnecessary check due to type

ferhatelmas:
also fixes a small typo in error message


tag/map_codec.go
```diff
@@ -127,8 +127,8 @@ func (eg *encoderGRPC) readBytesWithVarintLen() ([]byte, error) {
 
 	valueStart += eg.readIdx
 	valueEnd := valueStart + int(length)
-	if valueEnd > len(eg.buf) || length < 0 {
```
odeke-em: True, length is a uint64 and the signature for binary.Uvarint says it'll be zero on any error so that's a redundant check.

PR #287, opened 2018-01-04
Simplify the trace package

rakyll:
- NewSpan now can create children; (*Span).StartSpan is removed.
- s/StartSpanOptions/StartOptions to help the verbosity.
- Only connivance functions works against the context object, Span type provides the full API.
- Introspection functions are prefixed with Report* that make them distinguishable.

### DIFF
--- a/trace/spanstore.go
+++ b/trace/spanstore.go
@@ -29,47 +29,8 @@ var (
 	spanStores = make(map[string]*spanStore)
 )
 
-// LatencySampledSpans returns a sample of successful spans.
-//
-// minLatency is the minimum latency of spans to be returned.
-// maxLatency, if nonzero, is the maximum latency of spans to be returned.
-func LatencySampledSpans(name string, minLatency, maxLatency time.Duration) []*SpanData {
-	s := spanStoreForName(name)
-	if s == nil {
-		return nil
-	}
-	var out []*SpanData
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	for i, b := range s.latency {
-		min, max := latencyBucketBounds(i)
-		if i+1 != len(s.latency) && max <= minLatency {
-			continue
-		}
-		if maxLatency != 0 && maxLatency < min {
-			continue
-		}
-		for _, sd := range b.buffer {
-			if sd == nil {
-				break
-			}
-			if minLatency != 0 || maxLatency != 0 {
-				d := sd.EndTime.Sub(sd.StartTime)
-				if d < minLatency {
-					continue
-				}
-				if maxLatency != 0 && d > maxLatency {
-					continue
-				}
-			}
-			out = append(out, sd)
-		}
-	}
-	return out
-}
-
-// ActiveSpans returns the active spans for the given name.
-func ActiveSpans(name string) []*SpanData {
+// ReportActiveSpans returns the active spans for the given name.
+func ReportActiveSpans(name string) []*SpanData {
 	s := spanStoreForName(name)
 	if s == nil {
 		return nil
@@ -83,10 +44,10 @@ func ActiveSpans(name string) []*SpanData {
 	return out
 }
 
-// ErrorSampledSpans returns a sample of error spans.
+// ReportSpansByError returns a sample of error spans.
 //
 // If code is nonzero, only spans with that status code are returned.
-func ErrorSampledSpans(name string, code int32) []*SpanData {
+func ReportSpansByError(name string, code int32) []*SpanData {
 	s := spanStoreForName(name)
 	if s == nil {
 		return nil
@@ -146,27 +107,8 @@ func ConfigureBucketSizes(bcs []BucketConfiguration) {
 	}
 }
 
-// PerMethodSummary is a summary of the spans stored for a single span name.
-type PerMethodSummary struct {
-	Active         int
-	LatencyBuckets []LatencyBucketSummary
-	ErrorBuckets   []ErrorBucketSummary
-}
-
-// LatencyBucketSummary is a summary of a latency bucket.
-type LatencyBucketSummary struct {
-	MinLatency, MaxLatency time.Duration
-	Size                   int
-}
-
-// ErrorBucketSummary is a summary of an error bucket.
-type ErrorBucketSummary struct {
-	ErrorCode int32
-	Size      int
-}
-
-// SampledSpansSummary returns a summary of what spans are being stored for each span name.
-func SampledSpansSummary() map[string]PerMethodSummary {
+// ReportSpansPerMethod returns a summary of what spans are being stored for each span name.
+func ReportSpansPerMethod() map[string]PerMethodSummary {
 	out := make(map[string]PerMethodSummary)
 	ssmu.RLock()
 	defer ssmu.RUnlock()
@@ -193,6 +135,64 @@ func SampledSpansSummary() map[string]PerMethodSummary {
 		out[name] = p
 	}
 	return out
+}
+
+// ReportSpansByLatency returns a sample of successful spans.
+//
+// minLatency is the minimum latency of spans to be returned.
+// maxLatency, if nonzero, is the maximum latency of spans to be returned.
+func ReportSpansByLatency(name string, minLatency, maxLatency time.Duration) []*SpanData {
+	s := spanStoreForName(name)
+	if s == nil {
+		return nil
+	}
+	var out []*SpanData
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for i, b := range s.latency {
+		min, max := latencyBucketBounds(i)
+		if i+1 != len(s.latency) && max <= minLatency {
+			continue
+		}
+		if maxLatency != 0 && maxLatency < min {
+			continue
+		}
+		for _, sd := range b.buffer {
+			if sd == nil {
+				break
+			}
+			if minLatency != 0 || maxLatency != 0 {
+				d := sd.EndTime.Sub(sd.StartTime)
+				if d < minLatency {
+					continue
+				}
+				if maxLatency != 0 && d > maxLatency {
+					continue
+				}
+			}
+			out = append(out, sd)
+		}
+	}
+	return out
+}
+
+// PerMethodSummary is a summary of the spans stored for a single span name.
+type PerMethodSummary struct {
+	Active         int
+	LatencyBuckets []LatencyBucketSummary
+	ErrorBuckets   []ErrorBucketSummary
+}
+
+// LatencyBucketSummary is a summary of a latency bucket.
+type LatencyBucketSummary struct {
+	MinLatency, MaxLatency time.Duration
+	Size                   int
+}
+
+// ErrorBucketSummary is a summary of an error bucket.
+type ErrorBucketSummary struct {
+	ErrorCode int32
+	Size      int
 }
 
 // spanStore keeps track of spans stored for a particular span name.
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -43,15 +43,6 @@ type Span struct {
 	exportOnce sync.Once
 }
 
-// IsRecordingEvents returns true if events are being recorded for the current span.
-func IsRecordingEvents(ctx context.Context) bool {
-	s, ok := ctx.Value(contextKey{}).(*Span)
-	if !ok {
-		return false
-	}
-	return s.IsRecordingEvents()
-}
-
 // IsRecordingEvents returns true if events are being recorded for this span.
 func (s *Span) IsRecordingEvents() bool {
 	if s == nil {
@@ -62,15 +53,6 @@ func (s *Span) IsRecordingEvents() bool {
 
 // TraceOptions contains options associated with a trace span.
 type TraceOptions uint32
-
-// IsSampled returns true if the current span will be exported.
-func IsSampled(ctx context.Context) bool {
-	s, ok := ctx.Value(contextKey{}).(*Span)
-	if !ok {
-		return false
-	}
-	return s.IsSampled()
-}
 
 // IsSampled returns true if this span will be exported.
 func (s *Span) IsSampled() bool {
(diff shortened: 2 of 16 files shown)


trace/spanstore.go
```diff
@@ -29,47 +29,8 @@ var (
...
-			out = append(out, sd)
-		}
-	}
-	return out
-}
-
-// ActiveSpans returns the active spans for the given name.
-func ActiveSpans(name string) []*SpanData {
+// ReportActiveSpans returns the active spans for the given name.
+func ReportActiveSpans(name string) []*SpanData {
```
jba: Could the Report functions go in another package? I see they are useful for a server status page, but that is really a different use case from adding traces, even though many servers will do both. For example, a client may want traces but will not have a status page.

trace/trace.go
```diff
@@ -469,44 +313,16 @@ func (s *Span) lazyPrintfInternal(attributes []Attribute, format string, a ...in
...
-	}
-	s.LazyPrintfWithAttributes(attributes, format, a...)
-}
-
-// LazyPrintfWithAttributes adds an annotation with attributes.
+// Logf adds an annotation with attributes.
 //
 // The format string is evaluated with its arguments only when the annotation text is needed.
-func (s *Span) LazyPrintfWithAttributes(attributes []Attribute, format string, a ...interface{}) {
+func (s *Span) Logf(attributes []Attribute, format string, a ...interface{}) {
```
jba: I think you made the right choice in keeping the lazy one, but you should take care to mention how that can cause problems. I think it could be as simple as ".... only when the annotation text is needed, so any arguments should not be modified subsequently" or something like that.

trace/spanstore.go
rakyll: There is a natural cyclic dependency between reporting functions and the trace.SpanData type. If we are not introducing a new type to return from these functions, it is not possible to reorg them to a different package. I really wished that it was possible to split them.

rakyll:
One more option that is not represented here is that we can have a hybrid approach similar to the tags package. Starting spans can be done via context.Context and everything else can be delegated to the Span objects.

In the tags package, this is how we enforce users not to drop the tags in the current context.

```
func StartSpan(ctx context.Context, name string) (context.Context, *Span)
func StartSpanWithOptions(ctx context.Context, name string, o StartSpanOptions) context.Context (context.Context, *Span)
func StartSpanWithRemoteParent(ctx context.Context, name string, parent SpanContext, o StartSpanOptions) (context.Context, *Span)
```

trace/trace.go
ramonza: `lazyPrintfInternal` formats the string so there isn't an opportunity for arguments to be modified subsequently in the current implementation. I assume you mean we should leave open the door to allow future implementation changes that would retain the arguments longer? If that's the case, I don't know if we should that - it seems like asking too much of the developer to think that much about a logging function.

ramonza: (Unless of course the contemplated implementation change is well-understood, imminent and unavoidable)

adriancole:
I suppose lack of clarity around if the current context is a pre-allocated
span vs the intended parent isnt an issue as a call site would explain it..

Ex a context based start would be used in framework based code that
intentionally pre-allocates a span for purpose of adding tags or setting
ids in scope, but without affecting the duration of the span

(Just speaking out loud not saying any change needed)

adriancole:
Meant to say "pre-allocates a context"

trace/trace.go
jba: I thought that format call was just a placeholder, and the intent was always to do defer the formatting for longer. Currently there isn't a real difference between the lazy and non-lazy cases, I believe.

jba:
> Starting spans can be done via context.Context and everything else can be delegated to the Span objects.

I would have a complete set of methods on *Span. Then I would add functions on context.Context as needed for convenience. Based on my (limited) experience, only Start/EndSpan would actually be used enough to warrant context functions.

rakyll:
> I would have a complete set of methods on *Span. Then I would add functions on context.Context as needed for convenience.

I think this is a reasonable solution. Span methods can always represent the low-level API whereas we can have utilities as we need them.

I pushed a new change where I added Start* and EndSpan functions back. I made the starting functions return the newly initiated span object. So the user can easily depend on the low-level API even if they use the convenience functions.

```
ctx, span = trace.StartSpan(ctx, "your choice of name")
defer trace.EndSpan(ctx)

span.SetAttributes(...)
```

adriancole:
As a point of reference, Brave (java tracer) uses this api internally
(keyed on context) as it allows us to migrate span apis with the smallest
dependency between the before and after (context). Also propagation can be
simpler with only passing context. Not sure if applies here

ramonza:
What do you think about making a subpackage `trace/span` so that we can say things like:
```
import "go.opencensus.io/trace/span"
...
ctx := span.Start()
span.Logf(ctx, ...)
```

Rename types: trace.SpanContext -> span.Context, trace.Span -> span.Span (?)

rakyll:
Span seems to be the core type in the trace package. I think we shouldn't separate Span from the trace package.

```
import "go.opencensus.io/trace/span"
...
ctx := span.Start()
span.Logf(ctx, ...)
```

IMHO, this is even a more confusing API given it is not clear whether the library is relying on passing the span object or the context around.

> As a point of reference, Brave (java tracer) uses this api internally
(keyed on context) as it allows us to migrate span apis with the smallest
dependency between the before and after (context). 

This is why I don't think we should hide the Span type. For use cases where there is no incoming context, users have flexibility to propagate traces in whichever way they prefer. 

Context is also specific to request-scoped data. We are not limiting spans to be associated with RPCs or requests, it feels like a better call to have Span type as the first class abstraction.

Also Go context's lifetime is different than the spans it is carrying around. Users have to make a conscious effort and understand which span they are working against at any time.

We all agree that providing utilities to make it easier to propagate spans around in the current context is good but it is not covering all the use cases. I suggest the current API, keeping the Span type and only providing utilities for those who are propagating via context.

rakyll:
Also quick note that context objects are not supposed to be kept as state, from https://golang.org/pkg/context/:

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx.

There are places in the stdlib such as Request.Context where it is a state but this is due to the fact that context is added to Go later and we couldn't break the APIs due to Go 1.x API promise.

I think it is the best to allow user to have access to the span type rather than expecting them to pin the current context which is an anti-pattern in Go.

ramonza:
it wasn't clear from the code but I wasn't suggesting making Context a state variable (I was proposing span.Context as a type (span is the package)).

Lots of functions in package trace start with Span. It seems to me to "want" to be called package span. Tracing is the concept but "span" is what you deal with when instrumenting code. This is a naming issue and is somewhat independent of the consideration of whether we want to be trace.Span-based or context.Context-based in APIs.

> IMHO, this is even a more confusing API given it is not clear whether the library is relying on passing the span object or the context around.

I agree that this is confusing, I would like to suggest resolving the confusion by using context.Context instead of trace.Span.

> Context is also specific to request-scoped data. We are not limiting spans to be associated with RPCs or requests, it feels like a better call to have Span type as the first class abstraction.

I'm not sure about this, they feel pretty similar to me. You could for example use context.Context to  scope an expensive local computation, or create context.Context around each disk access. In these cases, many of the context.Context facilities are just as useful outside of an RPC situation (deadlines, cancellation). It would be odd to invent a new context-like abstraction for these use cases IMO just because they are not RPC.

>Also Go context's lifetime is different than the spans it is carrying around. Users have to make a conscious effort and understand which span they are working against at any time.

I thought this as well but is it really the case? You can't ever really disassociate a span from a context.Context once it is started (as show in TestStartSpanAfterEnd). So effectively once you have started a Span on a context.Context they are really inextricable. I think it is more confusing to allow them as separate objects once this link has been made because there is no way to change the Span associated with a context.Context, or even to remove it.

Looking ahead, I think avoiding context.Context also precludes future interesting things we can do to join the tags in the context with the trace (for example #341). The fact that OpenCensus supports both tracing and stats is a unique strength and if we don't pass the context.Context to all trace methods, we break the chain that links the trace back to the stats/tags. 

On the other hand if we do pass the context to all tracing methods, then we may as well pass *only* the context and avoid begging the question of which Span you mean (the arg or the one in the context.Context).

rakyll:
> Lots of functions in package trace start with Span. It seems to me to "want" to be called package span.

Package names are the namespace in Go. For example, span.Span is a lint error due to the stutter. trace.Span represents a trace span which is also obvious from the name. 

> Looking ahead, I think avoiding context.Context also precludes future interesting things we can do to join the tags in the context with the trace.

The purpose of having the Span object is not mutually exclusive to be able to propagate it via context. If you are propagating the span, you should always use context if possible. There is no question about it. But if you need to extract and store the span data as state, you should be able to do it too. User should not depend on the span objects when propagating.

jba:
> Package names are the namespace in Go. For example, span.Span is a lint error due to the stutter. trace.Span represents a trace span which is also obvious from the name.

+1

I'm also on @rakyll's side w.r.t Span vs. Context. Maybe I'm missing your point, @ramonza, but a span is clearly a distinct thing with its own fields and operations, and I think it's just clearer to expose it as its own type.

rakyll:
@ramonza and @bogdandrutu chatted offline and I applied a few more changes:

- We will have StartSpanX and EndSpan functions for convince.
- NewSpan now can create children so I removed the (*Span).StartSpan.
- Starting and ending functions/methods can now utilize the the given tags.
- Introduced EndOptions when ending spans to be able to apply tags when ending spans.
- s/StartSpanOptions/StartOptions to help the verbosity.

PTAL

trace/spanstore.go
ramonza: I think in a subsequent PR we should explore moving this functionality to an in-memory Exporter.

trace/trace.go
```diff
@@ -123,73 +107,83 @@ func WithSpan(parent context.Context, s *Span) context.Context {
...
-// StartSpan starts a new child span.
+// EndSpan ends the current span.
 //
-// If s is nil, creates a new trace and span, like the function NewSpan.
-func (s *Span) StartSpan(name string) *Span {
-	return s.StartSpanWithOptions(name, StartSpanOptions{})
+// The context passed to EndSpan will still refer to the now-ended span, so any
+// code that adds more information to it, like SetSpanStatus, should be called
+// before EndSpan.
+func EndSpan(ctx context.Context) {
```
ramonza: I don't think this convenience method is carrying it's weight, I think we should remove it and require the user to call End on the Span object.

trace/trace.go
```diff
@@ -246,21 +240,14 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
...
-func EndSpan(ctx context.Context) {
-	s, ok := ctx.Value(contextKey{}).(*Span)
-	if !ok {
-		return
-	}
-	s.End()
+// EndOptions represents the options
+// when ending a span.
+type EndOptions struct {
+	Tags *tag.Map
```
ramonza: I don't think we should do this for now. Recall @bogdandrutu's concern about repeated tags because they were propagated from the caller. What we really want here is to only add tags that were created in the scope of *this* span, not propagated tags (since they should be recorded in the parent Span).

trace/spanstore.go
```diff
@@ -195,6 +137,64 @@ func SampledSpansSummary() map[string]PerMethodSummary {
 	return out
 }
 
+// ReportSpansByLatency returns a sample of successful spans.
+//
+// minLatency is the minimum latency of spans to be returned.
+// maxLatency, if nonzero, is the maximum latency of spans to be returned.
+func ReportSpansByLatency(name string, minLatency, maxLatency time.Duration) []*SpanData {
```
ramonza: I would be OK removing this functionality until we find the right home for it.

trace/trace.go
```diff
@@ -64,15 +57,6 @@ func (s *Span) IsRecordingEvents() bool {
 // TraceOptions contains options associated with a trace span.
 type TraceOptions uint32
 
-// IsSampled returns true if the current span will be exported.
-func IsSampled(ctx context.Context) bool {
```
ramonza: We should remove this from Span as well and have it only on SpanContext

trace/trace.go
```diff
@@ -123,73 +107,83 @@ func WithSpan(parent context.Context, s *Span) context.Context {
...
 //
-// If s is nil, creates a new trace and span, like the function NewSpan.
-func (s *Span) StartSpanWithOptions(name string, o StartSpanOptions) *Span {
-	if s != nil {
-		return startSpanInternal(name, true, s.spanContext, false, o)
-	}
-	return startSpanInternal(name, false, SpanContext{}, false, o)
+// If there is an existing span in ctx, it is ignored -- the returned Span is a
+// child of the span specified by parent.
+func StartSpanWithRemoteParent(ctx context.Context, name string, parent SpanContext, o StartOptions) (context.Context, *Span) {
```
ramonza: I don't think this convenience method is carrying it's weight.

trace/trace.go
```diff
@@ -123,73 +107,83 @@ func WithSpan(parent context.Context, s *Span) context.Context {
...
+	if parent != nil {
+		hasParent = true
+		parentSpanContext = parent.SpanContext()
+	}
+	return startSpanInternal(name, hasParent, parentSpanContext, false, o)
 }
 
 // NewSpanWithRemoteParent returns a new span with the given parent SpanContext.
-func NewSpanWithRemoteParent(name string, parent SpanContext, o StartSpanOptions) *Span {
+func NewSpanWithRemoteParent(name string, parent SpanContext, o StartOptions) *Span {
```
ramonza: Can we just call this NewSpan and remove the above NewSpan? Caller can always explicitly retrieve the SpanContext from the parent span and then call this method.

trace/trace.go
ramonza: I think we should tackle this separately.

trace/trace.go
```diff
@@ -123,73 +107,83 @@ func WithSpan(parent context.Context, s *Span) context.Context {
...
-type StartSpanOptions struct {
+// StartOptions contains options concerning how a span is started.
+type StartOptions struct {
 	// RecordEvents indicates whether to record data for this span, and include
 	// the span in a local span store.
 	// Events will also be recorded if the span will be exported.
 	RecordEvents bool
-	Sampler      // if non-nil, the Sampler to consult for this span.
+
+	Sampler // if non-nil, the Sampler to consult for this span.
```
jba: nit: by embedding, you're giving StartOptions all the methods of Sampler, which feels wrong. Even though it stutters, it is safer to write `Sampler Sampler`.

trace/trace.go
```diff
@@ -123,73 +107,83 @@ func WithSpan(parent context.Context, s *Span) context.Context {
...
-	return s.StartSpanWithOptions(name, StartSpanOptions{})
+// The context passed to EndSpan will still refer to the now-ended span, so any
+// code that adds more information to it, like SetSpanStatus, should be called
+// before EndSpan.
+func EndSpan(ctx context.Context) {
+	s, ok := ctx.Value(contextKey{}).(*Span)
+	if !ok {
+		return
+	}
+	s.End(EndOptions{Tags: tag.FromContext(ctx)})
```
jba: This extra bit does seem like it would be nice to have EndSpan for convenience.

trace/trace.go
jba: Agreed.

trace/trace.go
jba: But I see below that you may defer tags until later, in which case I agree this can be removed.

trace/spanstore.go
rakyll: +1.

trace/spanstore.go
rakyll: +1.

trace/trace.go
rakyll: Done

trace/trace.go
rakyll: We cannot remove this because NewSpanWithRemoteParent inherits the start options from the parent child.

PR #294, opened 2018-01-05
Introduce tag.Do for profiling

### LINKED ISSUE #45: tag: add pprof.Do equivalent to record profiler samples with a tag map
Add a [pprof.Do](https://golang.org/pkg/runtime/pprof/#Do) like utility to record profiler samples with a given tag map. Do looks for the tags inside the given context and sets them as pprof labels. Any goroutine started in f needs to propagate the tag map via c in order to be able to keep setting the labels in the new goroutines. 

```
func Do(ctx context.Context, m *Map, f func(c context.Context))
```

rakyll:
tag.Do labels the profiler samples with the key values from the tag map. It is only enabled for users
who are running Go 1.9 and above.

Fixes #45.

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,30 @@ ctx = trace.StartSpan(ctx, "your choice of name")
 defer trace.EndSpan(ctx)
 ```
 
+More tracing examples are coming soon...
+
+## Profiling
+
+OpenCensus tags can be applied as profiler labels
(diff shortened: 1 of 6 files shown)


rakyll:
Attaching a sample output. "opencensus-key:opencensus-value" is coming from a tag map.

profiles

PR #298, opened 2018-01-07
tag.Do: call f on pre-1.9 to preserve program behavior.

ramonza:
pprof.Do internally calls f. On pre-1.9, we should also call f so the caller
doesn't themselves need to write code conditional on the Go version.


rakyll:
Ooops :)

PR #299, opened 2018-01-07
DO NOT MERGE: make tag.Do install the tags in the context

ramonza:


rakyll:
IMO, https://github.com/census-instrumentation/opencensus-go/pull/300 is a better option. pprof.Do is focused on applying the label set momentarily because users are not expected to propagate the labels around. For OpenCensus, it is that the tags are propagated in the context.

ramonza:
Agreed. Closing this one.

PR #304, opened 2018-01-08
tag: simplify key type check in decode

ferhatelmas:
Would remove `break` first since it's unnecessary but then thought `if` would be simpler.

### DIFF
--- a/tag/map_codec.go
+++ b/tag/map_codec.go
@@ -195,10 +195,7 @@ func Decode(bytes []byte) (*Map, error) {
 	for !eg.readEnded() {
 		typ := keyType(eg.readByte())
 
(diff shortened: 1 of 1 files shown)


tag/map_codec.go
```diff
@@ -195,10 +195,7 @@ func Decode(bytes []byte) (*Map, error) {
 	for !eg.readEnded() {
 		typ := keyType(eg.readByte())
```
rakyll: This is dead code now. We won't have any other types than string. I will do a round of internal cleanup to rewrite this section.

ferhatelmas: Great, ping me if you think I might be useful somewhere.

Issue #285, opened 2018-01-03
trace: support Trace context header


mhindery:
We are currently using https://godoc.org/cloud.google.com/go/trace for tracing a microservice to Stackdriver. Seeing as it is declared obsolete, we are moving to the replacement package https://godoc.org/go.opencensus.io/trace. These seems to be an important feature for us missing, namely taking over a trace context from an incoming request header. The introduction code after the first example code block on the old package's godoc page describes it:

`Calling SpanFromRequest will create a new trace span for an incoming HTTP request. If the request contains a trace context header, it is used to determine the trace ID. Otherwise, a new trace ID is created.`

This allows to create and link traces crossing boundaries over multiple services. I can't find any similar functionality in the new package, so traces are confined within a single 'program'. Is this intentional? Are we supposed to implement this ourselves by checking and parsing the header and creating new child Spans with custom SpanOptions having set the TraceID and Span ID? I do think this is convenient functionality which is useful enough to be provided out-of-the-box for everybody and like to have it added.

rakyll:
We are working on a vendor agnostic approach to propagate traces via HTTP. Until the multi-vendor support is in, we might provide a Stackdriver-specific utility.

@bogdandrutu, thoughts?

rakyll:
Also please consider taking a look at [httptrace](https://godoc.org/github.com/rakyll/goutil/opencensus/plugin/http/httptrace). It provides a reference how to propagate the header for Stackdriver Trace.

PR #306, opened 2018-01-09
 Add HTTP propagation support for traces

rakyll:
The change introduces propagation.HTTPFormat. Implementations of HTTPFormat allows trace context to be propagated in HTTP requests.

The change also includes tracing integrations with the net/http package.

Updates #285.

### DIFF
--- a/plugin/http/httptrace/httptrace.go
+++ b/plugin/http/httptrace/httptrace.go
@@ -23,11 +23,11 @@ import (
 	"go.opencensus.io/trace/propagation"
 )
 
-const httpHeader = `X-Cloud-Trace-Context`
+// TODO(jbd): Add godoc examples.
 
 type transport struct {
 	base        http.RoundTripper
-	propogators []propagation.HTTPPropagator
+	propogators []propagation.HTTPFormat
 }
 
 // RoundTrip creates a trace.Span and inserts it into the outgoing request's headers.
@@ -35,12 +35,14 @@ type transport struct {
 // the request's context.
 func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	name := "Sent" + strings.Replace(req.URL.String(), req.URL.Scheme, ".", -1)
+	// TODO(jbd): Discuss whether we want to prefix
+	// outgoing requests with Sent.
 	ctx := trace.StartSpan(req.Context(), name)
 	req = req.WithContext(ctx)
 
 	span := trace.FromContext(ctx)
 	for _, p := range t.propogators {
-		req = p.ToRequest(span.SpanContext(), req)
+		p.ToRequest(span.SpanContext(), req)
 	}
 
 	resp, err := t.base.RoundTrip(req)
@@ -63,10 +65,11 @@ func (t *transport) CancelRequest(req *http.Request) {
 // NewTransport returns an http.RoundTripper that traces the outgoing requests.
 //
 // All the requests are done by the given base roundtripper. If nil is given,
-// http.DefaultTransport is used.
+// http.DefaultTransport is used. If its base HTTP RoundTripper implements CancelRequest,
+// the returned round tripper will be cancelable.
 //
 // Traces are propagated via the provided HTTP propagation mechanisms.
-func NewTransport(base http.RoundTripper, p ...propagation.HTTPPropagator) http.RoundTripper {
+func NewTransport(base http.RoundTripper, p ...propagation.HTTPFormat) http.RoundTripper {
 	if base == nil {
 		base = http.DefaultTransport
 	}
@@ -80,16 +83,16 @@ func NewTransport(base http.RoundTripper, p ...propagation.HTTPPropagator) http.
 //
 //    span := trace.FromContext(r.Context())
 //
-// The span will be auto finished by the handler.
+// The span will be automatically ended by the handler.
 //
 // Incoming propagation mechanism is determined by the given HTTP propagators.
-func NewHandler(base http.Handler, p ...propagation.HTTPPropagator) http.Handler {
+func NewHandler(base http.Handler, p ...propagation.HTTPFormat) http.Handler {
 	return &handler{handler: base, propogators: p}
 }
 
 type handler struct {
 	handler     http.Handler
-	propogators []propagation.HTTPPropagator
+	propogators []propagation.HTTPFormat
 }
 
 func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


plugin/http/httptrace/httptrace_test.go
```diff
@@ -0,0 +1,149 @@
...
+			rt.RoundTrip(req)
+
+			req = <-transport.ch
+			span := trace.FromContext(req.Context())
+
+			if header := req.Header.Get("trace"); header == "" {
+				t.Fatalf("Trace header = empty; want valid trace header")
+			}
+			if span == nil {
+				t.Fatalf("Got no spans in req context; want one")
```
ramonza: does this assertion actually test anything in the current package? seems like it just tests that trace.WithSpan works

plugin/http/httptrace/httptrace_test.go
```diff
@@ -0,0 +1,149 @@
...
+			if tt.parent != nil {
+				if got, want := span.SpanContext().TraceID, tt.parent.SpanContext().TraceID; got != want {
+					t.Errorf("span.SpanContext().TraceID=%v; want %v", got, want)
+				}
+			}
+		})
+	}
+}
+
+func TestHandler(t *testing.T) {
```
ramonza: it would be nice to have an example that uses both client and server together, but maybe this is better left until we have a concrete HTTPPropagator

plugin/http/httptrace/httptrace.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	resp, err := t.base.RoundTrip(req)
+
+	// TODO(jbd): Add status and attributes.
+	trace.EndSpan(ctx)
+	return resp, err
+}
+
+// CancelRequest cancels an in-flight request by closing its connection.
```
odeke-em: Could we add:
```go
// If its base HTTP RoundTripper implements CancelRequest
```

plugin/http/httptrace/httptrace.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// NewHandler returns a http.Handler from the given handler
+// that is aware of the incoming request's span.
+// The span can be extracted from the incoming request in handler
+// functions from incoming request's context:
+//
+//    span := trace.FromContext(r.Context())
+//
+// The span will be auto finished by the handler.
```
odeke-em: Let's make this more clear
```go
The span will be automatically ended by the handler
```
if I understood the intention of the code and the above statement, properly.

plugin/http/httptrace/httptrace.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"go.opencensus.io/trace"
+	"go.opencensus.io/trace/propagation"
+)
+
+const httpHeader = `X-Cloud-Trace-Context`
+
+type transport struct {
+	base        http.RoundTripper
+	propogators []propagation.HTTPPropagator
```
ramonza: Do we want to encourage people to add multiple propagators? Seems like it would be more common to standardize on a particular format or perhaps to accept multiple formats but produce only a single format.

plugin/http/httptrace/httptrace.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// RoundTrip creates a trace.Span and inserts it into the outgoing request's headers.
+// The created span can follow a parent span, if a parent is presented in
+// the request's context.
+func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	name := "Sent" + strings.Replace(req.URL.String(), req.URL.Scheme, ".", -1)
```
ramonza: The name here strikes me as odd because "sent" seems like the name of an event rather than a description of the span itself. Is this standardized somewhere? I would be less surprised by a name starting with "Client" or "HTTP"

plugin/http/httptrace/httptrace.go
```diff
@@ -0,0 +1,120 @@
...
+	return &handler{handler: base, propogators: p}
+}
+
+type handler struct {
+	handler     http.Handler
+	propogators []propagation.HTTPPropagator
+}
+
+func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	name := "Recv" + strings.Replace(r.URL.String(), r.URL.Scheme, ".", -1)
```
ramonza: Similar comment to "Sent" on the client. Also, can we move this variable declaration closer to its use?

plugin/http/httptrace/httptrace_test.go
rakyll: I am adding a TODO to add examples.

trace/propagation/propagation.go
```diff
@@ -89,3 +91,16 @@ func FromBinary(b []byte) (sc trace.SpanContext, ok bool) {
...
+
+// HTTPPropagator propoagates span contexts in HTTP requests.
+//
+// FromRequest extracts a span context from incoming
+// requests. If the request doesn't contain a trace header,
+// nil is returned.
+//
+// ToRequest modifies the given request to include the given
+// span context in the request. Modified request is returned.
+type HTTPPropagator interface {
```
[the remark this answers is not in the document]
rakyll: I agree that HTTPPropagator is a bad name. I wished that we had a short alternative more representative than HTTP. I am leaving a TODO to rename it if we can find a better name.

trace/propagation/propagation.go
```diff
@@ -89,3 +91,16 @@ func FromBinary(b []byte) (sc trace.SpanContext, ok bool) {
...
+	FromRequest(req *http.Request) (sc trace.SpanContext, ok bool)
+	ToRequest(sc trace.SpanContext, req *http.Request) *http.Request
```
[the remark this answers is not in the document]
rakyll: Done

plugin/http/httptrace/httptrace_test.go
rakyll: It tests whether RoundTrip created a new span (that has the given parent as its parent span) and used the propagator to modify the HTTP header called "trace".

rakyll:
@ramonza @odeke-em please take another look!

plugin/http/httptrace/httptrace.go
ramonza: I think this is the kind of thing that should be documented across languages in census-instrumentation/opencensus-specs#13

trace/propagation/propagation.go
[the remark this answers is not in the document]
ramonza: HTTPFormat? this is similar to the Java name

odeke-em:
Sorry, I was away from my computer for the whole of yesterday.

PR #307, opened 2018-01-09
Regenerate README

rakyll:


rakyll:
README is not reflecting the changes from the source markdown. Also, it is not embedding the snippets properly.

rakyll:
I failed to make embedmd use README as the source file. I applied the changes to the source markdown and regenerated the README.

@ramonza, I am guessing you are using the upstream embedmd rather than my own [fork](https://github.com/rakyll/embedmd). My fork allows users to tag the beginning and the ending of the snippet. If you use the upstream embedmd, it will fail to extract based on the START,END pragmas. The README explains how to go get it: https://github.com/census-instrumentation/opencensus-go/tree/master/internal/readme.

ramonza:
Oops!

PR #310, opened 2018-01-10
Add HTTP propagation support for Stackdriver Trace

### LINKED ISSUE #285: trace: support Trace context header

rakyll:
Fixes #285.

### DIFF
--- a/exporter/trace/stackdriver/stackdriver.go
+++ b/exporter/trace/stackdriver/stackdriver.go
@@ -28,7 +28,7 @@
 // 	trace.RegisterExporter(exporter)
 //
 // The package uses Application Default Credentials to authenticate.  See
-// https://developers.google.com/identity/protocols/application-default-credentials.
+// https://developers.google.com/identity/protocols/application-default-credentials
 //
 // Exporter can be used to propagate traces over HTTP requests for Stackdriver Trace.
 //
@@ -58,6 +58,7 @@ import (
 	"time"
 
 	"go.opencensus.io/internal"
+	"go.opencensus.io/trace/propagation"
 
 	tracingclient "cloud.google.com/go/trace/apiv2"
 	"go.opencensus.io/trace"
@@ -66,7 +67,10 @@ import (
 	tracepb "google.golang.org/genproto/googleapis/devtools/cloudtrace/v2"
 )
 
-const httpHeader = `X-Cloud-Trace-Context`
+const (
+	httpHeader        = `X-Cloud-Trace-Context`
+	httpHeaderMaxSize = 200
+)
 
 // Exporter is an implementation of trace.Exporter that uploads spans to
 // Stackdriver.
@@ -83,6 +87,7 @@ type Exporter struct {
 }
 
 var _ trace.Exporter = (*Exporter)(nil)
+var _ propagation.HTTPFormat = (*Exporter)(nil)
 
 // Options contains options for configuring an exporter.
 //
@@ -192,7 +197,7 @@ func (e *Exporter) FromRequest(req *http.Request) (sc trace.SpanContext, ok bool
 	// See https://cloud.google.com/trace/docs/faq for the header format.
 	// Return if the header is empty or missing, or if the header is unreasonably
 	// large, to avoid making unnecessary copies of a large string.
-	if h == "" || len(h) > 200 {
+	if h == "" || len(h) > httpHeaderMaxSize {
 		return trace.SpanContext{}, false
 	}
 
--- a/exporter/trace/stackdriver/stackdriver.go
+++ b/exporter/trace/stackdriver/stackdriver.go
@@ -45,7 +45,6 @@
 package stackdriver
 
 import (
-	"bytes"
 	"context"
 	"encoding/binary"
 	"encoding/hex"
@@ -205,7 +204,6 @@ func (e *Exporter) FromRequest(req *http.Request) (sc trace.SpanContext, ok bool
 	slash := strings.Index(h, `/`)
 	if slash == -1 {
 		return trace.SpanContext{}, false
-
 	}
 	tid, h := h[:slash], h[slash+1:]
 
@@ -233,12 +231,10 @@ func (e *Exporter) FromRequest(req *http.Request) (sc trace.SpanContext, ok bool
 	// Parse the options field, options field is optional.
 	if !strings.HasPrefix(h, "o=") {
 		return sc, true
-
 	}
 	o, err := strconv.ParseUint(h[2:], 10, 64)
 	if err != nil {
 		return trace.SpanContext{}, false
-
 	}
 	sc.TraceOptions = trace.TraceOptions(o)
 	return sc, true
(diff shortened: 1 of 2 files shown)


exporter/trace/stackdriver/stackdriver.go
```diff
@@ -47,8 +66,13 @@ import (
 	tracepb "google.golang.org/genproto/googleapis/devtools/cloudtrace/v2"
 )
 
+const httpHeader = `X-Cloud-Trace-Context`
+
 // Exporter is an implementation of trace.Exporter that uploads spans to
 // Stackdriver.
+//
+// Exporter also implements trace/propagation.HTTPFormat and can
```
odeke-em: Below it, let's add a compile type assertion
```go
var _ propagation.HTTPFormat = (*Exporter)(nil)
```

exporter/trace/stackdriver/stackdriver.go
```diff
@@ -162,6 +186,71 @@ func (e *Exporter) uploadToStackdriver(spans []*trace.SpanData) {
 	}
 }
 
+// FromRequest extracts a Stakdriver Trace span context from incoming requests.
+func (e *Exporter) FromRequest(req *http.Request) (sc trace.SpanContext, ok bool) {
+	h := req.Header.Get(httpHeader)
+	// See https://cloud.google.com/trace/docs/faq for the header format.
+	// Return if the header is empty or missing, or if the header is unreasonably
+	// large, to avoid making unnecessary copies of a large string.
+	if h == "" || len(h) > 200 {
```
odeke-em: Could we please extract this magic number limit and make it a const perhaps?

exporter/trace/stackdriver/stackdriver.go
```diff
@@ -162,6 +186,71 @@ func (e *Exporter) uploadToStackdriver(spans []*trace.SpanData) {
...
+		return trace.SpanContext{}, false
+	}
+
+	// Parse the trace id field.
+	slash := strings.Index(h, `/`)
+	if slash == -1 {
+		return trace.SpanContext{}, false
+
```
odeke-em: Spurious line, let's remove it.

exporter/trace/stackdriver/stackdriver.go
```diff
@@ -162,6 +186,71 @@ func (e *Exporter) uploadToStackdriver(spans []*trace.SpanData) {
...
+	}
+
+	buf = make([]byte, 8)
+	binary.PutUvarint(buf, sid)
+	copy(sc.SpanID[:], buf)
+
+	// Parse the options field, options field is optional.
+	if !strings.HasPrefix(h, "o=") {
+		return sc, true
+
```
odeke-em: Another spurious line, please remove it.

exporter/trace/stackdriver/stackdriver.go
```diff
@@ -162,6 +186,71 @@ func (e *Exporter) uploadToStackdriver(spans []*trace.SpanData) {
...
+	}
+	o, err := strconv.ParseUint(h[2:], 10, 64)
+	if err != nil {
+		return trace.SpanContext{}, false
+
```
odeke-em: ditto, spurious line.

exporter/trace/stackdriver/stackdriver.go
```diff
@@ -162,6 +186,71 @@ func (e *Exporter) uploadToStackdriver(spans []*trace.SpanData) {
...
+// ToRequest modifies the given request to include a Stackdriver Trace header.
+func (e *Exporter) ToRequest(sc trace.SpanContext, req *http.Request) {
+	var b bytes.Buffer
+	b.WriteString(hex.EncodeToString(sc.TraceID[:]))
+	sid, _ := binary.Uvarint(sc.SpanID[:])
+	b.WriteString("/")
+	b.WriteString(strconv.FormatUint(sid, 10))
+	b.WriteString(";o=")
+	b.WriteString(strconv.FormatInt(int64(sc.TraceOptions), 10))
```
odeke-em: Would it perhaps be more clear to avoid all this calls but do them once as
```go
sid, _ := binary.Uvarint(sc.SpanID[:])
fmt.Fprintf(buf, "%x/%d;o=%d", sc.TraceID[:], sid, int64(sc.TraceOptions))
```

exporter/trace/stackdriver/stackdriver.go
```diff
@@ -28,13 +28,32 @@
 // 	trace.RegisterExporter(exporter)
 //
 // The package uses Application Default Credentials to authenticate.  See
-// https://developers.google.com/identity/protocols/application-default-credentials
+// https://developers.google.com/identity/protocols/application-default-credentials.
```
odeke-em: Also could we not put the "." right after the URL? Some browsers such as Safari at times mistakenly pick up the "." and create invalid URLs.

exporter/trace/stackdriver/stackdriver.go
rakyll: Done

exporter/trace/stackdriver/stackdriver.go
rakyll: What do you mean by spurious? Are you suggesting to use naked returns?

If so, we end up sometimes using the naked returns and sometimes not. For example, the following case is naked return friendly.

```
 if slash == -1 {
 		return trace.SpanContext{}, false
}
```

And the following is not because ok becomes false. I don't want to assign the right values right before returning.

```
o, err := strconv.ParseUint(h[2:], 10, 64)
 	if err != nil {
 		return trace.SpanContext{}, false
```

In order to be consistent and be more readable, I prefer the current style.

rakyll:
Please take another look.

exporter/trace/stackdriver/stackdriver.go
odeke-em: By `spurious line` I meant, you have a newline after the return statement and before the closing brace e.g
```go
if foo {
    return bar

}
```

exporter/trace/stackdriver/stackdriver.go
odeke-em: `spurious newline` :)

rakyll:
All done, thanks.

odeke-em:
Awesome, thanks!

PR #317, opened 2018-01-11
Export Transport

### LINKED ISSUE #309: Support httptrace.ClientTrace for annotating traces
We should provide an implementation that harvests events and annotates the current trace (added in #306)

rakyll:
Then, we can use the Transport struct to set options such as
the one required to fix #309.

### DIFF
--- a/plugin/http/httptrace/httptrace.go
+++ b/plugin/http/httptrace/httptrace.go
@@ -25,15 +25,26 @@ import (
 
 // TODO(jbd): Add godoc examples.
 
-type transport struct {
-	base    http.RoundTripper
-	formats []propagation.HTTPFormat
+// Transport is an http.RoundTripper that traces the outgoing requests.
+//
+// Use NewTransport to create new transports.
+type Transport struct {
+	// Base represents the underlying roundtripper that does the actual requests.
+	// If none is given, http.DefaultTransport is used.
+	//
+	// If base HTTP roundtripper implements CancelRequest,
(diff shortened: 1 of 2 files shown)


plugin/http/httptrace/httptrace.go
```diff
@@ -25,15 +25,23 @@ import (
...
+// Transport is an http.RoundTripper that traces the outgoing requests.
+//
+// Use NewTransport to create new transports.
+type Transport struct {
+	// Base represents the underlying roundtripper that does the actual requests.
+	// If none is given, http.DefaultTransport is used.
+	//
+	// If base HTTP roundtripper implements CancelRequest,
+	// the returned round tripper will be cancelable.
+	Base    http.RoundTripper
```
ramonza: why does Base need to be exposed?

ramonza: Oh I see you need to set it explicitly now

rakyll:
> a) Add a mutex to protect Base in t.base()

No, holding a mutex for this is an overkill and costly. We can simply say, mutate Base before using Transport. This is the overall style in the standard library. Fields and methods are not expected to be safe for concurrent use by default in Go.

> It isn't clear to me why we need to export this type in order to provide options on it
Because I am trying to make the zero value useful rather than introducing a new type. We should also export the `formats` field and keep NewTransport around only for convenience.

rakyll:
Can you take another look?

PR #320, opened 2018-01-12
trace: Annotation and MessageEvent have suboptimal String() output Fixes: #319

ramonza:

PR #321, opened 2018-01-12
Don't embed time.Time where it should be represented as a field

### LINKED ISSUE #319: trace: Annotation and MessageEvent have suboptimal String() output
trace.Annotation and trace.MessageEvent both embed time.Time, but don't override String().
This means they inherit the String() method from time.Time, leading to misleading stringified forms.

Recommend either not embedding time.Time if it's not necessary for the structs to behave like time.Time structs, if not then redefining the String() methods on both.

rakyll:
Fixes #319.


rakyll:
Duplicate of #320.

PR #323, opened 2018-01-17
Document that trace.Exporters should be safe for concurrent use.

ramonza:


trace/export.go
```diff
@@ -21,9 +21,9 @@ import (
 
 // Exporter is a type for functions that receive sampled trace spans.
 //
-// The Export method should return quickly; if an Exporter takes a significant
-// amount of time to process a SpanData, that work should be done on another
-// goroutine.
+// The Export method should be goroutine-safe and should return quickly; if an
```
rakyll: be safe for concurrent calls and should return quickly.

"Safe for concurrent use" is the terminology we prefer.

PR #325, opened 2018-01-17
Revise annotation APIs

### LINKED ISSUE #129: trace: reduce the number of the printing APIs
The trace package currently allows users to print in a few different ways:

```
func LazyPrint(ctx context.Context, str fmt.Stringer)
func LazyPrintWithAttributes(ctx context.Context, attributes []Attribute, str fmt.Stringer)
func LazyPrintf(ctx context.Context, format string, a ...interface{})
func LazyPrintfWithAttributes(ctx context.Context, attributes []Attribute, format string, a ...interface{})
func Print(ctx context.Context, str string)
func PrintWithAttributes(ctx context.Context, attributes []Attribute, str string)
```

Instead of giving users different combinations of these options, we should provide a printing entry type and allow user to provide their options via the new type.

```
type Entry struct {
	Text       fmt.Stringer
	Attributes []Attribute
}

func LazyPrint(ctx context.Context, e Entry)
```

Then, we can provide maybe one other option for convience:

```
func LazyPrintf(ctx context.Context, format string, a ...interface{})
```

Users who doesn't want lazy formatting can always implement their own fmt.Stringer to format it at the point LazyPrint is called.

/cc @jcd2 @bogdandrutu

rakyll:
Provide Annoate and Annoatef.

We don't want to the lazy behavior of the annotation
APIs exposed to the user. Don't document the lazy
behavior.

Passing nil as attributes meets the case for annotating
only with a message. We are removing all other functions
and methods and expect user to pass nil if they
don't want to annotate with attributes.

Fixes #129.

### DIFF
--- a/exporter/trace/stackdriver/proto_test.go
+++ b/exporter/trace/stackdriver/proto_test.go
@@ -71,8 +71,8 @@ func TestExportTrace(t *testing.T) {
 		{
 			ctx2 := trace.StartSpan(ctx1, "span2")
 			trace.AddMessageSendEvent(ctx2, 0x123, 1024, 512)
-			trace.LazyPrintf(ctx2, "in span%d", 2)
-			trace.LazyPrint(ctx2, big.NewRat(2, 4))
+			trace.Annotatef(ctx2, nil, "in span%d", 2)
+			trace.Annotate(ctx2, nil, big.NewRat(2, 4).String())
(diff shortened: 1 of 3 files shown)


rakyll:
Fixing the typos...

rakyll:
@bogdandrutu and I designed this new API so he already LGTMed the APIs. @odeke-em, can you take another look?

PR #327, opened 2018-01-18
Add a test to exhibit #326.

ramonza:

Issue #328, opened 2018-01-18
closing a Span multiple times causes it to be re-submitted to Exporter


ramonza:
Instead, we should just ignore closing more than once.

odeke-em:
We can also add to each span an endOnce sync.Once field whose .Do method will only be invoked once ever during End.

ramonza:
A related issue that I'm lumping in here is that when you end a span and start a span on the same context, the second span get's added as a child of the first (ended) span.

PR #329, opened 2018-01-18
Only invoke trace.EndSpan if resp.Body.Close is invoked

### LINKED ISSUE #326: httptrace client span should only end once body is fully read
Right now, httptrace.Transport wrapper ends the client span immediately after RoundTrip. The problem is that this might be before the server span has ended if there's a response body that is still being written. The client span should only end after the response body has been fully read or if its closed.

odeke-em:
Wrap resp.Body in a proxy reader that'll only invoke
trace.EndSpan(ctx) after resp.Body.Close is invoked
otherwise currently we are erraneously invoking
trace.EndSpan(ctx) before the server has fully
served back content.

Fixes #326
Relies on test code #327

### DIFF
--- a/plugin/http/httptrace/httptrace.go
+++ b/plugin/http/httptrace/httptrace.go
@@ -97,8 +97,7 @@ func (seb *spanEndBody) Read(b []byte) (int, error) {
 	default:
 		// For all other errors, set the span status
 		trace.SetSpanStatus(seb.spanCtx, trace.Status{
-			// Code 2 is for Internal server error as per
-			// https://github.com/googleapis/googleapis/blob/f704d14a7224a140bca5cc26835fae471eaf7281/google/rpc/code.proto#L44-L51
+			// Code 2 is the error code for Internal server error.
 			Code:    2,
 			Message: err.Error(),
 		})
--- a/plugin/http/httptrace/httptrace.go
+++ b/plugin/http/httptrace/httptrace.go
@@ -24,7 +24,6 @@ import (
 
 	"go.opencensus.io/trace"
 	"go.opencensus.io/trace/propagation"
-	"google.golang.org/api/googleapi"
 )
 
 // TODO(jbd): Add godoc examples.
@@ -69,7 +68,8 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 
 	// trace.EndSpan(ctx) will be invoked after
-	// resp.Body.Close() has been invoked.
+	// a read from resp.Body returns io.EOF or when
+	// resp.Body.Close() is invoked.
 	resp.Body = &spanEndBody{rc: resp.Body, spanCtx: ctx}
 	return resp, err
 }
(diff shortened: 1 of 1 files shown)


plugin/http/httptrace/httptrace.go
```diff
@@ -59,10 +62,38 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+
+var _ io.ReadCloser = (*bodyProxyReader)(nil)
+
+func (bpr *bodyProxyReader) Read(b []byte) (int, error) {
+	return bpr.rc.Read(b)
+}
+
+func (bpr *bodyProxyReader) Close() error {
+	var err error
+	bpr.closeOnce.Do(func() {
```
ramonza: is this necessary? seems like if the caller calls close more than once they are at fault. Is this just a workaround for #328? If so I think it would be better to fix the root cause.

plugin/http/httptrace/httptrace.go
```diff
@@ -59,10 +62,38 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+type bodyProxyReader struct {
+	rc        io.ReadCloser
+	closeOnce sync.Once
+	ctxToEnd  context.Context
+}
+
+var _ io.ReadCloser = (*bodyProxyReader)(nil)
+
+func (bpr *bodyProxyReader) Read(b []byte) (int, error) {
+	return bpr.rc.Read(b)
```
ramonza: should we end the span if read returns EOF?

ramonza: or if it returns any other terminal error, should we set the span status?

plugin/http/httptrace/httptrace.go
```diff
@@ -59,10 +62,38 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		return resp, err
+	}
+
+	// trace.EndSpan(ctx) will be invoked after
+	// resp.Body.Close() has been invoked.
+	resp.Body = &bodyProxyReader{rc: resp.Body, ctxToEnd: ctx}
 	return resp, err
 }
 
+type bodyProxyReader struct {
```
ramonza: proxy is maybe a bit misleading here (might be interpreted as HTTP proxy), what about something like tracedBody?

plugin/http/httptrace/httptrace.go
odeke-em: @ramonza good questions. I wasn't thinking about #328 when I wrote that, it is just a precaution that I like to have to preserve `trace.EndSpan(ctx)` being invoked once as it was before.

plugin/http/httptrace/httptrace.go
odeke-em: >should we end the span if read returns EOF?

I also wondered about that. It might even make more sense to end the span if it returns EOF than if Close() was invoked, because for precision, we are tracing from request sent to last byte of response received, and not necessarily on when the caller invokes response.Body.Close(). Thanks for raising this.

odeke-em:
Thank you @ramonza for the first round of review, I've addressed the comments with a push, PTAL.

plugin/http/httptrace/httptrace.go
```diff
@@ -59,10 +63,75 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+
+	// Otherwise, time to end the span or set the status
+	switch err {
+	case io.EOF:
+		bpr.endSpan()
+	default:
+		// For all other errors, set the span status
+		// If the error has a status and code
+		// let's propagate those in the span
+		if ge, ok := err.(*googleapi.Error); ok {
```
rakyll: Don't import any Google API specific symbols into this package.

odeke-em: Problem is that that is contrary to the hint in https://github.com/census-instrumentation/opencensus-go/blob/38381063fc2e146b7b20f389e35f505153a89a5b/trace/basetypes.go#L114-L121

otherwise we don't have another way of finding out the Status and Message unfortunately :(

rakyll: I think there is something I cannot follow. How can a raw HTTP request can return a googleapi.Error? There should be a layer in between that takes the raw http.Request body and converts it to a googleapi.Error object.

rakyll: To be clear, we won't do it in this package.

odeke-em: Thank you @rakyll for breaking this down for me offline; I had a contrived assumption, I'll update this in a second.

odeke-em:
@rakyll and @ramonza I've updated the PR with your suggestions, PTAL.

plugin/http/httptrace/httptrace.go
```diff
@@ -59,10 +62,67 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+
+	// Otherwise, time to end the span or set the status
+	switch err {
+	case io.EOF:
+		bpr.endSpan()
+	default:
+		// For all other errors, set the span status
+		trace.SetSpanStatus(bpr.spanCtx, trace.Status{
+			// Code 2 is for Internal server error as per
+			// https://github.com/googleapis/googleapis/blob/f704d14a7224a140bca5cc26835fae471eaf7281/google/rpc/code.proto#L44-L51
```
rakyll: We can remove the link, source codes are not good source of truths because they often change, removed, reorged, etc.

plugin/http/httptrace/httptrace.go
```diff
@@ -59,10 +62,67 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	spanCtx context.Context
+
+	endSpanOnce sync.Once
+}
+
+var _ io.ReadCloser = (*spanEndBody)(nil)
+
+func (bpr *spanEndBody) Read(b []byte) (int, error) {
+	n, err := bpr.rc.Read(b)
+	if err == nil {
```
rakyll: Can't this case be a part of the switch below?

odeke-em: SGTM, updating in a second with a force push.

plugin/http/httptrace/httptrace.go
odeke-em: Hmm, but then how do we keep a reference? That link is a permalink btw so won't ever disappear unless it is force pushed or repo deleted.

plugin/http/httptrace/httptrace.go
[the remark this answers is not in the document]
rakyll: I agree that status is not required here.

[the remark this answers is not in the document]
odeke-em: @ramonza that PR code is outdated, I replaced it with a code 2 set for any non-nil error. Are y'all asking that I remove any status sets and just process a read as it would normally be done?
Don't worry about the back and forth, it allows for exchange of ideas, so please feel free to propose as many changes ;)

plugin/http/httptrace/httptrace.go
```diff
@@ -59,10 +62,65 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	switch err {
+	case nil:
+		return n, nil
+	case io.EOF:
+		seb.endSpan()
+	default:
+		// For all other errors, set the span status
+		trace.SetSpanStatus(seb.spanCtx, trace.Status{
+			// Code 2 is for Internal server error as per
+			// https://github.com/googleapis/googleapis/blob/f704d14a7224a140bca5cc26835fae471eaf7281/google/rpc/code.proto#L44-L51
```
rakyll: Remove the link to the file.

plugin/http/httptrace/httptrace.go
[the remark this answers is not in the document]
odeke-em: @ramonza and @rakyll the difference here is that if we get non2XX(*http.Response.Code) the user can handle that which is the case that you are talking about.
Whereas by the time we are in a *http.Response.Body.Read, we already have a 2XX code, the span is being recorded but we encountered an unexpected error hence we are setting the status of the span and I believe that this case is mutually exclusive of the case where we are setting by HTTP status.

odeke-em:
I've removed the link as you requested @rakyll, PTAL.

rakyll:
LGTM

PR #331, opened 2018-01-18
Prevent submitting a span to Exporters twice.

ramonza:
Fixes: #328

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -41,7 +41,7 @@ type Span struct {
 	spanContext SpanContext
 	// spanStore is the spanStore this span belongs to, if any, otherwise it is nil.
 	*spanStore
-	once sync.Once
+	exportOnce sync.Once
 }
 
 // IsRecordingEvents returns true if events are being recorded for the current span.
@@ -264,7 +264,7 @@ func (s *Span) End() {
 	if !s.IsRecordingEvents() {
 		return
 	}
-	s.once.Do(func() {
+	s.exportOnce.Do(func() {
 		// TODO: optimize to avoid this call if sd won't be used.
 		sd := s.makeSpanData()
 		sd.EndTime = time.Now()
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -274,10 +274,10 @@ func (s *Span) End() {
 		if s.spanContext.IsSampled() {
 			// TODO: consider holding exportersMu for less time.
 			exportersMu.Lock()
+			defer exportersMu.Unlock()
 			for e := range exporters {
 				e.Export(sd)
 			}
-			exportersMu.Unlock()
 		}
 	})
 }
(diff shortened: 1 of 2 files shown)


trace/trace_test.go
```diff
@@ -582,3 +582,63 @@ func TestBucket(t *testing.T) {
...
+	UnregisterExporter(&spans)
+	if len(spans) != 3 {
+		t.Fatalf("expected 3 spans, got %#v", spans)
+	}
+	var span1, span2, parent *SpanData
+	for _, span := range spans {
+		switch span.Name {
+		case "span-1":
+			span1 = span
+			break
```
rakyll: You don't have to break after case body. Go default behavior is to break.

trace/trace_test.go
```diff
@@ -582,3 +582,63 @@ func TestBucket(t *testing.T) {
...
+		case "span-2":
+			span2 = span
+			break
```
rakyll: ditto

trace/trace_test.go
```diff
@@ -582,3 +582,63 @@ func TestBucket(t *testing.T) {
...
+		case "parent":
+			parent = span
+			break
```
rakyll: ditto

trace/trace_test.go
```diff
@@ -582,3 +582,63 @@ func TestBucket(t *testing.T) {
...
+		}
+	}
+	if span1.TraceID != parent.TraceID || span2.TraceID != parent.TraceID {
+		t.Errorf("expected same TraceID, got: %q, %q, %q", parent.TraceID, span1.TraceID, span2.TraceID)
+	}
+	if span1.ParentSpanID != parent.SpanID {
+		t.Errorf("span-1 parent, expected: %q got: %q", parent.SpanID, span1.ParentSpanID)
```
rakyll: Use got, want terminology. Got first, want second.

You can omit got if you already have an identifier for what's got in the message:

"span-1 parent = %v; want %v"

trace/trace.go
```diff
@@ -41,6 +41,7 @@ type Span struct {
 	spanContext SpanContext
 	// spanStore is the spanStore this span belongs to, if any, otherwise it is nil.
 	*spanStore
+	once sync.Once
```
odeke-em: Could we make this field more desscriptive like `exportOnce`?

trace/trace_test.go
```diff
@@ -582,3 +582,63 @@ func TestBucket(t *testing.T) {
...
+	ctx1 := StartSpanWithOptions(ctx, "span-1", StartSpanOptions{Sampler: AlwaysSample()})
+	EndSpan(ctx1)
+	// Start a new span with the context containing span-1
+	// even though span-1 is ended, we still add this as a new child of span-1
+	ctx2 := StartSpanWithOptions(ctx1, "span-2", StartSpanOptions{Sampler: AlwaysSample()})
+	EndSpan(ctx2)
+	EndSpan(ctx)
+	UnregisterExporter(&spans)
+	if len(spans) != 3 {
+		t.Fatalf("expected 3 spans, got %#v", spans)
```
odeke-em: Ditto about what @rakyll said about the Go style of got and want.

PR #336, opened 2018-01-19
Don't document the Interval window

rakyll:
This window is only useful for introspection and
we will consider removing it. Don't document it
for now not to encourage users to depend on it.


rakyll:
Thanks much for the fast review :)

PR #337, opened 2018-01-20
Fix the span names generated from HTTP requests

### LINKED ISSUE #335: plugin/http/httptrace: do not replace req.URL.Scheme unless it is set
httptrace_test.go:231: invalid name: Recv./.e.x.a.m.p.l.e./.u.r.l./.p.a.t.h.?.q.p.a.r.a.m.=.v.a.l.
	httptrace_test.go:226: invalid name: Sent.://127.0.0.1:59761/example/url/path?qparam=val

rakyll:
Handle cases where there is no hostname and port name.
Avoid putting :80 and :443 in the span name.

Fixes #335.

### DIFF
--- a/plugin/http/httptrace/httptrace.go
+++ b/plugin/http/httptrace/httptrace.go
@@ -19,7 +19,7 @@ import (
 	"context"
 	"io"
 	"net/http"
-	"strings"
+	"net/url"
 	"sync"
 
 	"go.opencensus.io/trace"
@@ -48,7 +48,7 @@ type Transport struct {
 // The created span can follow a parent span, if a parent is presented in
 // the request's context.
 func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
-	name := "Sent" + strings.Replace(req.URL.String(), req.URL.Scheme, ".", -1)
+	name := spanNameFromURL("Sent", req.URL)
 	// TODO(jbd): Discuss whether we want to prefix
 	// outgoing requests with Sent.
 	ctx := trace.StartSpan(req.Context(), name)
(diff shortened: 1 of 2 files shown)


plugin/http/httptrace/httptrace.go
```diff
@@ -189,3 +189,12 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	h.handler.ServeHTTP(w, r)
 }
+
+func spanNameFromURL(prefix string, u *url.URL) string {
+	host := u.Hostname()
+	port := ":" + u.Port()
+	if port == ":" || port == ":80" || port == ":443" {
+		port = ""
+	}
+	return prefix + "." + host + port + u.Path
```
odeke-em: How come we are no longer including the query string in here?

yurishkuro: query string can be very high cardinality, like `?userId={uuid}`. But then the `Path` could contain UUIDs as well, is it a good idea to include it in the span name?

ramonza: I agree with the removal of the query string. I view the Span name as something that should be "static", akin to Tag/Measure/View names.

ramonza: If we need the query string, it could be added as a annotation to the Span.

plugin/http/httptrace/httptrace_test.go
```diff
@@ -273,3 +275,38 @@ func serveHTTP(done chan struct{}, wait chan time.Time) string {
...
+		want   string
+	}{
+		{
+			prefix: "Sent",
+			u:      "http://localhost:80/hello?q=a",
+			want:   "Sent.localhost/hello",
+		},
+		{
+			prefix: "Recv",
+			u:      "https://localhost:443/a",
```
ramonza: Could you add an example with a non-standard port?

rakyll: Path will be represented as an attributed. Added a case for a non-standard port. PTAL.

PR #341, opened 2018-01-22
DO NOT MERGE: Stats for tracing experiment.

ramonza:
This is an experiment to see how easy it would be to derive useful stats from just tracing instrumentation.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -23,6 +23,8 @@ import (
 	"runtime"
 	"sync"
 	"time"
+	"go.opencensus.io/stats"
+	"sync/atomic"
 )
 
 // Span represents a span of a trace.  It has an associated SpanContext, and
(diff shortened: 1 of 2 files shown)


trace/trace.go
```diff
@@ -23,6 +23,9 @@ import (
 	"runtime"
 	"sync"
 	"time"
+	"go.opencensus.io/stats"
```
odeke-em: Could we separate non stdlib imports from stdlib imports?

trace/trace.go
```diff
@@ -122,33 +131,106 @@ func WithSpan(parent context.Context, s *Span) context.Context {
...
+	totalLatency *stats.MeasureFloat64
+}
+
+func NewSpanStatsOptions(prefix string, messageCounts bool) (*SpanStatsOptions, error) {
+	var (
+		err error
+		ss  SpanStatsOptions
+	)
+	intMeasures := []struct {
+		mp               **stats.MeasureInt64
```
odeke-em: Clever, me likey!

trace/trace.go
ramonza: agh, I always forget about this! I configured GoLand to do it automatically so hopefully that will help.

PR #349, opened 2018-01-23
Remove Interval and Window from View, only support Cumulative views.

ramonza:
Implements: https://github.com/census-instrumentation/opencensus-specs/issues/47
This still needs discussion with other languages but there is rough consensus on removing intervals in their current form and they are broken (#183) in Go right now anyway so I'd like to move ahead with this.


exporter/stats/stackdriver/stackdriver.go
```diff
@@ -451,16 +432,7 @@ func newLabelDescriptors(keys []tag.Key) []*labelpb.LabelDescriptor {
...
-	var w stats.Window
-
-	switch md.MetricKind {
-	case metricpb.MetricDescriptor_DELTA:
-		w = stats.Interval{}
-	case metricpb.MetricDescriptor_CUMULATIVE:
-		w = stats.Cumulative{}
-	}
-
+func equalAggWindowTagKeys(md *metricpb.MetricDescriptor, agg stats.Aggregation, keys []tag.Key) error {
```
odeke-em: s/equalAggWindowTagKeys/equalAggTagKeys/g
as this PR removes window

plugin/grpc/grpcstats/client_metrics.go
```diff
@@ -149,142 +125,6 @@ func defaultClientViews() {
 		RPCClientResponseCountView,
 	)
 	// TODO(jbd): Add roundtrip_latency, uncompressed_request_bytes, uncompressed_response_bytes, request_count, response_count.
-
-	RPCClientRoundTripLatencyMinuteView, _ = stats.NewView(
```
odeke-em: What's the rationale behind us getting rid of non-cumulative views?

stats/example_test.go
```diff
@@ -39,11 +38,10 @@ func Example_view() {
 
 	view, err := stats.NewView(
 		"my.org/views/openconns",
-		"open connections distribution over one second time window",
+		"open connections",
 		nil,
 		m,
 		stats.DistributionAggregation([]float64{0, 1000, 2000}),
-		stats.Interval{Duration: time.Second},
```
odeke-em: Yeah, I used this accidentally in the example :)

stats/view.go
```diff
@@ -26,8 +26,8 @@ import (
 	"go.opencensus.io/tag"
 )
 
-// View allows users to filter and aggregate the recorded events
-// over a time window. Each view has to be registered to enable
+// View allows users to filter and aggregate the recorded events.
+// Each view has to be registered to enable
 // data retrieval. Use NewView to initiate new views.
```
odeke-em: Perhaps we can backspace and move the `data retrieval` line to the end of the previous line.

odeke-em:
Some merge conflicts or perhaps a rebase needed as the exporter structure was changed.

songy23:
Could you add an issue about this on opencensus-specs as well?

plugin/grpc/grpcstats/client_metrics.go
ramonza: https://github.com/census-instrumentation/opencensus-specs/issues/47

ramonza:
https://github.com/census-instrumentation/opencensus-specs/issues/47

ramonza:
@odeke-em updated and resolved conflicts, PTAL

exporter/stackdriver/stats.go
```diff
@@ -237,28 +227,26 @@ func (e *statsExporter) makeReq(vds []*stats.ViewData, limit int) []*monitoringp
...
 	m := vd.View.Measure()
 	agg := vd.View.Aggregation()
-	window := vd.View.Window()
 	tagKeys := vd.View.TagKeys()
 	viewName := vd.View.Name()
 
 	if md, ok := e.createdViews[viewName]; ok {
-		// Check agg, window and keys.
-		return equalAggWindowTagKeys(md, agg, window, tagKeys)
+		return equalAggWindowTagKeys(md, agg, tagKeys)
```
odeke-em: We still never changed this helper s/equalAggWindowTagKeys/equalAggTagKeys/g

exporter/stackdriver/stats_test.go
```diff
@@ -803,20 +766,3 @@ func newTestCumViewData(v *stats.View, start, end time.Time, data1, data2 stats.
 	}
 }
 
-func newTestDistViewData(v *stats.View, start, end time.Time) *stats.ViewData {
```
odeke-em: Am not sure why we removed this helper? Aren't we still using Distribution aggregations in the tests?

plugin/grpc/grpcstats/client_metrics.go
```diff
@@ -42,24 +42,6 @@ var (
 	RPCClientResponseBytesView    *stats.View
 	RPCClientRequestCountView     *stats.View
 	RPCClientResponseCountView    *stats.View
-
-	RPCClientRoundTripLatencyMinuteView *stats.View
```
odeke-em: Hmm, let's also coordinate removing these views and interval windows from the Java client /cc @songy23 @sebright

plugin/grpc/grpcstats/client_metrics.go
odeke-em: Thanks for the link @ramonza.

exporter/stackdriver/stats.go
ramonza: oops, changing

exporter/stackdriver/stats_test.go
ramonza: It was only used in one place and I removed that reference so I also removed the function. That place was basically asserting that we don't create SD time series for Interval windows, an assertion that no longer made sense after the current change. I agree that we should have a test for distribution, but it looks like we don't have this right now and I don't really have the context to write one as part of this PR. So I'm going to add back the function and just add a TODO to write a proper test for SD distribution.

plugin/grpc/grpcstats/client_metrics.go
ramonza: I would prefer not to block this PR on other languages. Interval views are currently broken in Go anyway.

ramonza:
updated PTAL

ramonza:
@bogdandrutu I'd like to get this if you're ok with it?

plugin/grpc/grpcstats/client_metrics.go
sebright: /cc @dinooliva

ramonza:
Seems like this needs more discussion, let's hold off on merging for now.

ramonza:
we have consensus on this, I will resolve conflicts and then it can be merged

ramonza:
This has been merged separately into dev

PR #356, opened 2018-01-25
Allow a vendor to provide multiple exporters from the same package

### LINKED ISSUE #312: move everything stackdriver-related under a single package
We added Stackdriver trace propagation to the Stackdriver exporter package in #310 and this prompted a discussion that even though propagation and exporting are independent, it is more convenient to have everything related to Stackdriver in one place. 

This avoid a lot of boilerplate initialization when most of the time you want to do all these things at once:
* Set up Stackdriver trace exporter
* Set up Stackdriver stats exporter

rakyll:
This fixes the boilerplate required to initiatilize exporters for
vendors that support multiple backends.

In order to make this change, trace.Exporter and stats.Exporter
interfaces had to be modified.

Also, exporter subpackage is reorginized to have a flat layout.

Fixes #312.


rakyll:
This is a major API break, please do not merge before I create fixes for github.com/rakyll/opencensus-grpc-demo and github.com/googleCloudPlatform/google-cloud-go.

exporter/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,145 @@
...
+	e.statsExporter.ExportView(vd)
+}
+
+// ExportSpan exports a SpanData to Stackdriver Trace.
+func (e *Exporter) ExportSpan(sd *trace.SpanData) {
+	e.traceExporter.ExportSpan(sd)
+}
+
+// ToRequest modifies the given request to include a Stackdriver Trace header.
+func (e *Exporter) ToRequest(sc trace.SpanContext, req *http.Request) {
```
ramonza: I don't think that propagation should be conflated with exporting. It's valid to use just the stackdriver propagation format but not the stackdriver exporters, for example if you're communicating with Google APIs from AWS and using Jaeger as your tracing backend.

examples/grpc/helloworld_server/main.go
```diff
@@ -44,8 +43,7 @@ func (s *server) SayHello(ctx context.Context, in *pb.HelloRequest) (*pb.HelloRe
 func main() {
 	// Register stats and trace exporters to export
 	// the collected data.
-	stats.RegisterExporter(&exporter.StatsLogExporter{})
-	trace.RegisterExporter(&exporter.TraceLogExporter{})
```
ramonza: why don't we register the trace exporter any longer?

exporter/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,145 @@
...
+	"go.opencensus.io/trace"
+	"google.golang.org/api/option"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+)
+
+// Options contains options for configuring the exporter.
+type Options struct {
+	// ProjectID is the identifier of the Stackdriver
+	// project the user is uploading the stats data to.
+	ProjectID string
```
ramonza: Can this default from the environment?

rakyll: We can make this configurable via http://cloud.google.com/go/compute/metadata if the user is running their code on GCP.

exporter/stackdriver/stackdriver.go
rakyll: I will send a separate CL to address this split.

examples/grpc/helloworld_server/main.go
rakyll: We register exporter.Exporter instead that implements both.

PR #361, opened 2018-01-26
Split propagation.HTTPFormat implementations

### LINKED ISSUE #358: exporters: split propagation.HTTPFormat implementations from exporters

rakyll:
Fixes #358.


rakyll:
Java call this propagation.HTTPFormat CloudTrace. I am not sure whether the package name should be cloudtrace here. Reason is Cloud Trace is the old name for Stackdriver Trace, and this is a propagation header GCP understands as a whole.

Issue #363, opened 2018-01-26
exporter/stats/prometheus: MeanAggregation should be type Summary


songy23:
Currently `MeanAggregation` is converted to a Prometheus [`UntypedValue`](https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/prometheus/prometheus.go#L238). Instead I think it should be type [`Summary`](https://github.com/prometheus/client_golang/blob/master/prometheus/summary.go#L576) with sum and count, but no quantiles.

odeke-em:
@songy23 would you like to work on this?

songy23:
@odeke-em sure I can update this - should be an easy fix.

PR #366, opened 2018-01-27
Convert MeanData to ConstSummary

songy23:
Closes https://github.com/census-instrumentation/opencensus-go/issues/363.

### DIFF
--- a/exporter/prometheus/prometheus.go
+++ b/exporter/prometheus/prometheus.go
@@ -139,6 +139,12 @@ func (o *Options) onError(err error) {
 }
 
 // ExportView exports to the Prometheus if view data has one or more rows.
+// Each OpenCensus AggregationData will be converted to
+// corresponding Prometheus Metric: SumData will be converted
+// to Untyped Metric, CountData will be Counter Metric,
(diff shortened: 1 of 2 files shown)


rakyll:
The summary type is not going to work as expected without the quantiles. The users might get confused seeing no percentiles uploaded with the data. Can we document this in the exporter?

songy23:
Sure, I added some document for `Collect()`.

exporter/prometheus/prometheus.go
```diff
@@ -196,6 +196,12 @@ func (c *collector) Describe(ch chan<- *prometheus.Desc) {
 // and delivers them as Prometheus Metrics.
 // Collect is invoked everytime a prometheus.Gatherer is run
 // for example when the HTTP endpoint is invoked by Prometheus.
+// Each OpenCensus AggregationData will be converted to
```
rakyll: This new comments section is not readable by the user given collector is not exported, we can add it to the Export's godoc.

songy23: Yeah just realized that - I've moved it to ExportView().

songy23:
Thanks!

PR #368, opened 2018-01-27
Remove duplicated documentation

jadekler:
(copy paste error presumedly; same three lines listed twice)


jadekler:
Actually, copy paste with wrong verbiage. Making new PR

PR #370, opened 2018-01-29
Refactor stats into view and measure packages.

ramonza:
stats/measure is for recording stats (instrumentation), while stats/view contains collection, aggregation and exporting logic.


odeke-em:
@ramonza IMHO we are creating more disjoint APIs to import, by separating out measure and
view, which while nice for organizing, might require more user investment as the number of imports needed also increases.

For example doing
```go
vf := view.New(foo)
view.SetRecordingPeriod(2 * time.Second)
m := measure.NewInt64(...)
measure.Record(ctx, m.M(1))
```

seems more involved than
```go
vf := stats.NewView(foo)
stats.SetRecordingPeriod(2 * time.Second)
m := stats.NewMeasureInt64(...)
stats.Record(ctx, m.M(1))
```
which are from the same package.

I don't think that we should separate views and measure from stats.

I was hoping instead that we would rename stats to metrics as you had previously suggested.

rakyll:
@odeke-em, we see some value in separation because measures and views are for separate audiences. For example, we should be able to tell the library authors only to depend on the measure package and export views. On the other hand, it is binary's responsibility to subscribe to the views and setup the exporters.

Having this breakdown in the APIs would make easier even though it is adding more indirection.

stats/view/view.go
```diff
@@ -38,14 +40,20 @@ type View struct {
...
-	m Measure
+	m measure.Measure
 
 	subscribed uint32 // 1 if someone is subscribed and data need to be exported, use atomic to access
 
 	collector *collector
 }
 
-// NewView creates a new view with the given name and description.
+type Definition struct {
```
rakyll: What is definition?

ramonza: vestigial, will remove

ramonza:
This has been merged separately into dev

PR #373, opened 2018-01-29
Use correct HTTPFormat for Stackdriver

rakyll:
As a follow up to #361.


rakyll:
I will try move our godoc snippets to godoc examples as much as possible to avoid this problem.

PR #380, opened 2018-01-29
Add B3 propagation format support

### LINKED ISSUE #362: Support Zipkin/B3 propagation format
https://github.com/openzipkin/b3-propagation

rakyll:
B3 spans have two parts: client and server. For this reason,
B3 headers can contain a client span ID and a parent ID of the
span at the same time.

OpenCensus trace context doesn't allow the same span ID to be
shared between client and server. This is why we are going
to assume the incoming span ID is the parent.

For similar reasons, outgoing requests will not set the parent
span ID and assume the receiver will accept the outgoing span
ID as the client side of the span.

Fixes #362.

### DIFF
--- a/plugin/http/propagation/b3/b3.go
+++ b/plugin/http/propagation/b3/b3.go
@@ -1,0 +1,119 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package b3 contains a propagation.HTTPFormat implementation
+// for B3 propagation. See https://github.com/openzipkin/b3-propagation
+// for more details.
+package b3
+
+import (
+	"encoding/hex"
+	"net/http"
+
+	"go.opencensus.io/trace"
+	"go.opencensus.io/trace/propagation"
+)
+
+const (
+	traceIDHeader = "X─B3─TraceId"
+	spanIDHeader  = "X─B3─SpanId"
+	sampledHeader = "X─B3─Sampled"
+)
+
+// HTTPFormat implements propagation.HTTPFormat to propagate
+// traces in HTTP headers in B3 propagation format.
+// HTTPFormat skips the X-B3-ParentId and X-B3-Flags headers
+// because there are additional fields not represented in the
+// OpenCensus span context. Spans created from the incoming
+// header will be the direct children of the client-side span.
+// Similarly, reciever of the outgoing spans should use client-side
+// span created by OpenCensus as the parent.
+type HTTPFormat struct{}
+
+var _ propagation.HTTPFormat = (*HTTPFormat)(nil)
+
+// FromRequest extracts a B3 span context from incoming requests.
+func (f HTTPFormat) FromRequest(req *http.Request) (sc trace.SpanContext, ok bool) {
+	tid, ok := parseTraceID(req.Header.Get(traceIDHeader))
+	if !ok {
+		return trace.SpanContext{}, false
+	}
+	sid, ok := parseSpanID(req.Header.Get(spanIDHeader))
+	if !ok {
+		return trace.SpanContext{}, false
+	}
+	sampled, _ := parseSampled(req.Header.Get(sampledHeader))
+	return trace.SpanContext{
+		TraceID:      tid,
+		SpanID:       sid,
+		TraceOptions: sampled,
+	}, true
+}
+
+func parseTraceID(tid string) (trace.TraceID, bool) {
+	if tid == "" {
+		return trace.TraceID{}, false
+	}
+	b, err := hex.DecodeString(tid)
+	if err != nil {
+		return trace.TraceID{}, false
+	}
+	var traceID trace.TraceID
+	if len(b) <= 8 {
+		// The lower 64-bits.
+		start := 8 + (8 - len(b))
+		copy(traceID[start:], b)
+	} else {
+		start := 16 - len(b)
+		copy(traceID[start:], b)
+	}
+
+	return traceID, true
+}
+
+func parseSpanID(sid string) (spanID trace.SpanID, ok bool) {
+	if sid == "" {
+		return trace.SpanID{}, false
+	}
+	b, err := hex.DecodeString(sid)
+	if err != nil {
+		return trace.SpanID{}, false
+	}
+	start := (8 - len(b))
+	copy(spanID[start:], b)
+	return spanID, true
+}
+
+func parseSampled(sampled string) (trace.TraceOptions, bool) {
+	switch sampled {
+	case "true", "1":
+		return trace.TraceOptions(1), true
+	default:
+		return trace.TraceOptions(0), false
+	}
+}
+
+// ToRequest modifies the given request to include B3 headers.
+func (f *HTTPFormat) ToRequest(sc trace.SpanContext, req *http.Request) {
+	req.Header.Set(traceIDHeader, hex.EncodeToString(sc.TraceID[:]))
+	req.Header.Set(spanIDHeader, hex.EncodeToString(sc.SpanID[:]))
+
+	var sampled string
+	if sc.IsSampled() {
+		sampled = "1"
+	} else {
+		sampled = "0"
+	}
+	req.Header.Set(sampledHeader, sampled)
+}
--- a/plugin/http/propagation/b3/b3_test.go
+++ b/plugin/http/propagation/b3/b3_test.go
@@ -1,0 +1,210 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


rakyll:
@adriancole,

@bogdandrutu mentioned you suggested us to avoid parent IDs both for incoming and outgoing requests. What's the implication of not parsing the incoming parent and not setting the parent  on the outgoing requests? How compatible is this with the existing users of B3?

plugin/http/propagation/b3/b3.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// ToRequest modifies the given request to include B3 headers.
+func (f *HTTPFormat) ToRequest(sc trace.SpanContext, req *http.Request) {
+	req.Header.Set(traceIDHeader, hex.EncodeToString(sc.TraceID[:]))
+	req.Header.Set(spanIDHeader, hex.EncodeToString(sc.SpanID[:]))
+
+	var sid string
+	if sc.IsSampled() {
+		sid = "1"
```
odeke-em: Minor nit:
```go
if sc.IsSampled() {
    req.Header.Set(sampledHeader, sid)
}
```
instead of the extra checks and sid declaration.

rakyll: Fixed. I don't know what I was thinking 🤦‍♀️

adriancole:
Sorry was in transit. Will review later today and answer qs

rakyll:
@adriancole, take your time. This PR can wait.

plugin/http/propagation/b3/b3.go
```diff
@@ -0,0 +1,115 @@
...
+	tid, ok := parseTraceID(req.Header.Get(traceIDHeader))
+	if !ok {
+		return trace.SpanContext{}, false
+	}
+	sid, ok := parseSpanID(req.Header.Get(spanIDHeader))
+	if !ok {
+		return trace.SpanContext{}, false
+	}
+	sampled, _ := parseSampled(req.Header.Get(sampledHeader))
+	return trace.SpanContext{
```
adriancole: ps there's a X-B3-ParentId (helps but not going to break) and X-B3-Flags: 1 (the debug flag).

Possibly worthwhile mentioning we are skipping these.

plugin/http/propagation/b3/b3.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	var traceID trace.TraceID
+	if len(b) == 16 {
+		// 64-bit trace ID.
+		copy(traceID[:], b)
+	} else {
+		// 32-bit trace ID.
+		copy(traceID[8:], b[:8])
+	}
+	return traceID, true
```
adriancole: most lenient parsers will accept 1-32 character trace ID. (ps 32-bit/8 hex is an uncommon length). 128-bit/32 char trace ID is increasingly common

plugin/http/propagation/b3/b3.go
```diff
@@ -0,0 +1,115 @@
...
+		return trace.SpanID{}, false
+	}
+	var spanID trace.SpanID
+	copy(spanID[:], b)
+	return spanID, true
+}
+
+func parseSampled(sampled string) (trace.TraceOptions, bool) {
+	switch sampled {
+	case "0":
```
adriancole: for leniency, parse "false" and "true"

plugin/http/propagation/b3/b3.go
```diff
@@ -0,0 +1,115 @@
...
+		return trace.TraceOptions(0), true
+	case "1":
+		return trace.TraceOptions(1), true
+	default:
+		return trace.TraceOptions(0), false
```
adriancole: incidentally "false" will end up in this bucket, so "true" is the only one to explicitly leniently parse

plugin/http/propagation/b3/b3.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+// ToRequest modifies the given request to include B3 headers.
+func (f *HTTPFormat) ToRequest(sc trace.SpanContext, req *http.Request) {
+	req.Header.Set(traceIDHeader, hex.EncodeToString(sc.TraceID[:]))
+	req.Header.Set(spanIDHeader, hex.EncodeToString(sc.SpanID[:]))
+
+	if sc.IsSampled() {
```
adriancole: b3 is tri-state. so if IsSampled false is an explicit decision to not sample, it should be propagated as zero. Otherwise the next hop will re-sample, which is uncommon in B3 as they trust upstream by default

plugin/http/propagation/b3/b3_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+func TestHTTPFormat_FromRequest(t *testing.T) {
+	tests := []struct {
+		name    string
+		makeReq func() *http.Request
+		wantSc  trace.SpanContext
+		wantOk  bool
+	}{
+		{
+			name: "64-bit trace ID + 32-bit span ID; sampled=1",
```
adriancole: isn't this 128-bit + 64 bit? based on 32 hex and 16 hex

plugin/http/propagation/b3/b3.go
rakyll: This is related to the issue above. For OpenCensus, zero means there is no sampling decision and the next hop might need to determine whether they should sample or not. Let's try to address https://github.com/census-instrumentation/opencensus-go/pull/380/files#r164938998 first.

plugin/http/propagation/b3/b3.go
rakyll: Done, parse now support any length trace ID and span ID.

rakyll:
GitHub messed up some of my comments above. I am going to reply here.

For OpenCensus, upon arrival of the request,
TraceOptions=1 means there is a decision whether trace is sampled and it is 1.
TraceOptions=0 means there is no decision.

For OpenCensus, after span creation,
TraceOptions=1 means there is a decision whether trace is sampled and it is 1.
TraceOptions=0 means there is a decision whether trace is sampled and it is 0.

On the other hand, in B3,
X─B3─Sampled=1 means there is a decision whether trace is sampled and it is 1.
X─B3─Sampled=0 means there is a decision whether trace is sampled and it is 0.
X─B3─Sampled= means no decision is made, it is up to the next hop to decide on the sampling.

As soon as we receive an incoming request, we will create a child span for it and decide on sampling. So, for outgoing requests, we can always set a value for the X─B3─Sampled header.

I updated the PR.

adriancole:
@bogdandrutu <https://github.com/bogdandrutu> mentioned you suggested us to
avoid parent IDs both for incoming and outgoing requests. What's the
implication of not parsing the incoming parent and not setting the parent
on the outgoing requests? How compatible is this with the existing users of
B3?

Not sending parent ID has limited impact in normal case. While the parent
is in-flight, any lookup of the trace will have incorrect hierarchy when
the next node shares the same span ID. This would also impact dependency
graph linking. This will be resolved once the parent reports data to zipkin
as after that point the parent ID will be known, repairing the data. IOTW,
this race condition hurts lookup of in-flight long running requests, or
where a parent was unable to report to zipkin.

If there is a way to allow propagation of parent ID, we should. If not,
most wont notice, and if they care they will have to choose between
configuring to always fork a new child on incoming span, or flush in-flight
spans after a threshold (like 30s) to ensure hierarchy is ok even when
parent isnt propagated.

adriancole:
Ps aws xray uses the character '?' for the sampling decision not yet made
condition.

plugin/http/propagation/b3/b3.go
```diff
@@ -0,0 +1,125 @@
...
+// OpenCensus span context. Spans created from the incoming
+// header will be the direct children of the client-side span.
+// Similarly, reciever of the outgoing spans should use client-side
+// span created by OpenCensus as the parent.
+type HTTPFormat struct{}
+
+var _ propagation.HTTPFormat = (*HTTPFormat)(nil)
+
+// FromRequest extracts a B3 span context from incoming requests.
+func (f *HTTPFormat) FromRequest(req *http.Request) (sc trace.SpanContext, ok bool) {
```
ramonza: Why use a pointer receiver? I'm just curious to know if there's a reason to prefer them in general even for trivial cases like this?

plugin/http/propagation/b3/b3.go
```diff
@@ -0,0 +1,125 @@
...
+	return spanID, true
+}
+
+func parseSampled(sampled string) (trace.TraceOptions, bool) {
+	switch sampled {
+	case "false":
+		fallthrough
+	case "0":
+		return trace.TraceOptions(0), true
+	case "true":
```
odeke-em: Am not a fan of fallthroughs, could we make it simple
```go
case "true", "1":
     return trace.TraceOptions(1), true
```

plugin/http/propagation/b3/b3.go
rakyll: I used the pointer receiver because I cannot foresee whether there will be a HTTPFormat with a field we want to mutate in its methods. I want us to be consistent, not requiring a concrete type at one place and the pointer at another place.

plugin/http/propagation/b3/b3.go
rakyll: Done, thanks.

rakyll:
@adriancole, we currently behave as expected if "?" is received.

adriancole:
@rakyll oh ? will likely break most zipkin.  I'm sorry that I added a kindof random note. I would not prepare to receive (and certainly not send) ? in B3

rakyll:
We are processing ? as 0 and immediately use the sampler to have a sampling decision. We are definitely never sending ? in the outgoing requests. I hope this is the behavior you are asking :)

adriancole:
Well you should never receive ? From B3 rather interpret lack of
x-b3-sampled as ?

At any rate not sending ? is yes the important part ;)

PR #382, opened 2018-01-30
Parse Stackdriver span ID into a 64-bit big endian byte slice

rakyll:


rakyll:
@odeke-em, you mean 1<<63? I updated the tests.

odeke-em:
@rakyll sure :), I mentioned 1<<58 because it requires more than 8 bytes if we used binary.PutUvarint.

plugin/http/propagation/google/google_test.go
```diff
@@ -27,25 +26,25 @@ func TestHTTPFormat(t *testing.T) {
...
-	binary.PutUvarint(spanID[:], 123)
+	spanID1 := [8]byte{255, 0, 0, 0, 0, 0, 0, 123}
+	spanID2 := [8]byte{0, 0, 0, 0, 0, 0, 0, 123}
 	tests := []struct {
 		incoming        string
 		wantSpanContext trace.SpanContext
 	}{
 		{
-			incoming: "105445aa7843bc8bf206b12000100000/123;o=1",
+			incoming: "105445aa7843bc8bf206b12000100000/18374686479671623803;o=1",
```
odeke-em: Awesome, thank you!

Issue #381, opened 2018-01-30
Calculate span duration using the monotonic clock


rakyll:

ramonza:
I thought Go's time package "just worked" when comparing times? https://golang.org/pkg/time/#hdr-Monotonic_Clocks

rakyll:
time uses the monotonic clocks when you use Add, Sub, etc. We export an end time retrieved from time.Now() which is the wall clock. We should calculate the end time by using the monotonic clock.

bogdandrutu:
I agree that all times should be calculated using the monotonic time but I would prefer to keep the API with endtime for consistency with all the other languages.

rakyll:
I don't have strong opinions but will prefer to keep the EndTime field and document that it is calculated by using the monotonic clock.

adriancole:
Ps we have some pretty "interesting" work in brave which tries to keep the
same monotonic+real clock offset for the whole trace. This helps the spans
from not looking crap if ntp shift happens. Locking in this offset at trace
still allows ntp adjustments between traces to be visible. If we locked
larger scope we would end up with same host skew which can happen with
multiple containers

PR #384, opened 2018-01-30
Calculate span duration using the monotonic clock

ramonza:
Fixes: #381


zpages/tracez.go
```diff
@@ -251,10 +251,10 @@ func traceRows(s *trace.SpanData) []traceRow {
 	}
 
 	firstRow := traceRow{Fields: [3]string{wholeTime(start), "", ""}, SpanContext: s.SpanContext, ParentSpanID: s.ParentSpanID}
-	if s.EndTime.IsZero() {
+	if s.Duration == time.Duration(0) {
```
rakyll: s.Duration == 0

trace/export.go
```diff
@@ -56,16 +56,22 @@ func UnregisterExporter(e Exporter) {
...
+	Annotations     []Annotation
+	MessageEvents   []MessageEvent
+	Status          Status
 	Links           []Link
 	HasRemoteParent bool
 }
+
+// EndTime calculates the Duration offset from StartTime. This is robust to wall
+// clock changes between the start and end of the span.
+func (s *SpanData) EndTime() time.Time {
```
rakyll: I don't think we need to provide this. We can simply calculate it from start+dur.

ramonza:
Updated to restore EndTime and just adjust it.

ramonza:
@rakyll PTAL

PR #386, opened 2018-01-30
Remove unnecessary key manager

rakyll:


rakyll:
Btw, we don't care about uniqueness of the keys. We return the existing key.

tag/key.go
```diff
@@ -15,9 +15,19 @@
 
 package tag
 
+import (
+	"sync"
+)
+
+var keys = struct {
```
ramonza: this seems unnecessary, why doesn't NewKey just check the string and return Key{name}?

odeke-em:
@rakyll I think there is value in the old implementation if we want to keep low latency of the library.

odeke-em:
Your change simplifies the code and makes it cleaner, credit for that. However, what's the impetus for us to get rid of the old code? Seemed to me like the purpose was to intern/memoize already seen keys and also in case later on there was a limit to the number of keys it'd be a simple addition.

rakyll:
> in case later on there was a limit to the number of keys it'd be a simple addition.

There is no such limit we are considering. I am deleting some dead code.

PR #387, opened 2018-01-30
Fix some golint warnings.

ramonza:


zpages/code_string.go
```diff
@@ -1,26 +1,12 @@
-// Copyright 2017, OpenCensus Authors
```
odeke-em: I presume the codegen is the one that erased the licenses eh?

ramonza: yeah, I think it's fine

PR #388, opened 2018-01-31
Remove NewMap, add New

### LINKED ISSUE #346: Avoid having to deal with tag.Map for common instrumentation use cases
tag.Map seems like unnecessary ceremony: it is never useful to carry around a tag.Map outside of a context.Context, since the tags cannot be used for anything outside of a context. 

Instead of:
```go
import "go.opencensus.io/tag"
m, _ := tag.NewMap(ctx, m, tag.Insert(key1, value1))
ctx = tag.NewContext(ctx, m)
```

```go
import "go.opencensus.io/tag"
ctx, err := tag.NewContext(ctx, tag.Insert(key1, value1))
```
We can do this by having tag.Map implement tag.Mutator, modifying tag.NewContext to take tag.Mutator... and internally call tag.NewMap. This should preserve the existing semantics while also allowing the new short-hand. Encode/Decode functions taking a context should also be created.

rakyll:
Common case is to propagate tags in the context. Replace NewMap
with New.
In the case of an error, the old context should be preserved.

Fixes #346.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,15 +35,13 @@ Currently, OpenCensus supports:
 
 ## Tags
 
-Tags represent propagated key-value pairs. They can be propagated using context.Context
+Tags represent propagated key-value pairs. They are propagated using context.Context
 in the same process or can be encoded to be transmitted on the wire and decoded back
 to a tag.Map at the destination.
 
 ### Getting a key by a name
 
 A key is defined by its name. To use a key, a user needs to know its name and type.
-Currently, only keys of type string are supported.
-Other types will be supported in the future.
 
 [embedmd]:# (tags.go stringKey)
 ```go
@@ -54,11 +52,15 @@ if err != nil {
 }
 ```
 
-### Creating a map of tags associated with keys
+### Creating tags and propagating them
 
-tag.Map is a map of tags. Package tags provide a builder to create tag maps.
+Package tag provides a builder to create tag maps and put it
+into the current context.
+To propagate a tag map to downstream methods and RPCs, New
+will add the produced tag map to the current context.
+If there is already a tag map in the current context, it will be replaced.
 
-[embedmd]:# (tags.go tagMap)
+[embedmd]:# (tags.go new)
 ```go
 osKey, err := tag.NewKey("my.org/keys/user-os")
 if err != nil {
@@ -69,7 +71,7 @@ if err != nil {
 	log.Fatal(err)
 }
 
-tagMap, err := tag.NewMap(ctx,
+ctx, err = tag.New(ctx,
 	tag.Insert(osKey, "macOS-10.12.5"),
 	tag.Upsert(userIDKey, "cde36753ed"),
 )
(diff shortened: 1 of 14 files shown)


rakyll:
As we discussed offline, I renamed NewTags to New to avoid stuttering. PTAL.

README.md
```diff
@@ -54,11 +52,15 @@ if err != nil {
 }
 ```
 
-### Creating a map of tags associated with keys
+### Creating tags and propagating them
 
-tag.Map is a map of tags. Package tags provide a builder to create tag maps.
+Package tags provide a builder to create tag maps and put it
```
odeke-em: Package tags provides a builder to create and put tag maps into the current context.

README.md
```diff
@@ -80,29 +82,25 @@ if err != nil {
 
 ### Propagating a tag map in a context
 
-To propagate a tag map to downstream methods and RPCs, add a tag map
-to the current context. NewContext will return a copy of the current context,
-and put the tag map into the returned one.
-If there is already a tag map in the current context, it will be replaced.
+You can also propagate a tag map in the current context:
```
odeke-em: The code in line 89 shows the contrary though, it shows retrieving the tags map instead of propagating it.

README.md
```diff
@@ -80,29 +82,25 @@ if err != nil {
...
 
 [embedmd]:# (tags.go newContext)
 ```go
-ctx = tag.NewContext(ctx, tagMap)
+m := tag.FromContext(ctx)
 ```
 
-In order to update an existing tag map, get the tag map from the current context,
-use NewMap and put the new tag map back to the context.
+In order to update an existing tags from the current context,
```
odeke-em: to update existing tags

README.md
odeke-em: Did you mean to instead use `tag.New`?

tag/map_test.go
```diff
@@ -184,20 +189,18 @@ func TestNewMapValidation(t *testing.T) {
 
 	for i, tt := range tests {
 		ctx := NewContext(context.Background(), tt.seed)
-		m, err := NewMap(ctx)
+		ctx, err := New(ctx)
+		m := FromContext(ctx)
```
odeke-em: Let's move this line after line 203 when we know there isn't an error.

README.md
ramonza: The package is also named `tag` not `tags` but I think `tags` is a better name FWIW

rakyll: In Go, there is this culture of singularizing the packages because the package name is always a part of the symbol name.

Then tag.Map reads like tag map.

Done.

README.md
rakyll: Modified the wording a bit now, it is got clearer.

README.md
rakyll: Done.

tag/map_test.go
rakyll: Done.

PR #395, opened 2018-02-02
Rename trace.TraceID to trace.ID to avoid stutter.

ramonza:


bogdandrutu:
FYI: I feel this is very confusing.

rakyll:
IMHO, in this case TraceID is not terrible even though there is stutter. It is very easy to loose the context and keep asking what ID represents.

And we shouldn't rename the field in SpanContext type.

SpanContext.ID represents span context ID.
SpanContext.TraceID represents a trace ID.

ramonza:
On Fri, Feb 2, 2018 at 2:31 PM, JBD <notifications@github.com> wrote:

> IMHO, in this case TraceID is not terrible even though there is stutter.
> It is very easy to loose the context and keep asking what ID represents.
>

It might be easy for us to lose context while we are working inside the
trace package, but users will always refer to it as "trace.ID". Just want
to make sure we aren't making things easier for ourselves at the expense of
our users?


> And we shouldn't rename the field in SpanContext type.
>
Agh, Goland refactoring was overzealous here, I didn't mean to do that.

PR #406, opened 2018-02-05
Add trace proto

ramonza:


proto/traceproto/doc.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2017, OpenCensus Authors
```
odeke-em: s/2018/g

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "proto/opencensus-proto"]
+	path = proto/opencensus-proto
+	url = https://github.com/census-instrumentation/opencensus-proto
```
odeke-em: Am not sure if we should be git submodule-ing stuff into Go as the convention is to use vendor. However this isn't a Go package plus using a git submodule will help us keep a separate git tree and update it when needed. Let's then add some instructions though for updating this submodule, if we go forward with it.

ramonza: I thought it would be okay in this case since the submodule is only required for "go generate", not when using the library.

rakyll:
We discussed earlier with @bogdandrutu not to publish generated code directly from this repo but https://github.com/census-instrumentation/opencensus-proto.

ramonza:
@rakyll that makes sense, I'll add it there and close this PR

PR #413, opened 2018-02-07
Remove in-memory reporting from trace API

ramonza:
Remove the trace reporting methods from the public trace API.

### DIFF
--- a/internal/traceinternals.go
+++ b/internal/traceinternals.go
@@ -1,0 +1,5 @@
+package internal
+
+// TraceInternal allows internal access to some trace functionality.
+// TODO(#412): remove this
+var TraceInternal interface{}
--- a/internal/traceinternals.go
+++ b/internal/traceinternals.go
@@ -1,5 +1,35 @@
 package internal
 
+import "time"
+
 // TraceInternal allows internal access to some trace functionality.
 // TODO(#412): remove this
 var TraceInternal interface{}
+
+
+// BucketConfiguration stores the number of samples to store for span buckets
+// for successful and failed spans for a particular span name.
+type BucketConfiguration struct {
+	Name                 string
+	MaxRequestsSucceeded int
+	MaxRequestsErrors    int
+}
+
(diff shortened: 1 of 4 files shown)


internal/traceinternals.go
```diff
@@ -0,0 +1,49 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+import "time"
+
+// Trace allows internal access to some trace functionality.
+// TODO(#412): remove this
+var Trace interface{}
```
rakyll: var Reporter interface {
 	ReportActiveSpans(name string) []*trace.SpanData
 	ReportSpansByError(name string, code int32) []*trace.SpanData
 	ReportSpansByLatency(name string, minLatency, maxLatency time.Duration) []*trace.SpanData
        ReportSpansPerMethod() map[string]internal.PerMethodSummary
}

trace/spanstore.go
```diff
@@ -29,8 +31,15 @@ var (
 	spanStores = make(map[string]*spanStore)
 )
 
+// This exists purely to avoid exposing internal methods used by z-Pages externally.
+type internalOnly struct{}
```
rakyll: type reporter struct{}

internal/traceinternals.go
ramonza: reason it's untyped is to avoid a circular dependency

trace/spanstore.go
ramonza: in light of the previous comment about being unable to really create a strongly typed interface, do you still think this should be renamed? I kind of prefer it as it is because it's definitely just a placeholder, not trying to pretend that it's a real abstraction (that's future work) - it really is just trying to hide the functions in the simplest way possible.

internal/traceinternals.go
rakyll: Ah sorry, cool trick :)

trace/spanstore.go
rakyll: I am fine with the things the way they are.

PR #414, opened 2018-02-07
Use microsecond precision for jaeger span StartTime and Timestamp

bpot:
Update to match the behavior of the jaeger golang client:

https://github.com/jaegertracing/jaeger-client-go/blob/b7bacc465f385658dd3586f58810226776d2a4bf/jaeger_thrift_span.go#L28

https://github.com/jaegertracing/jaeger-client-go/blob/0e91c43a6bfeb3d710aa053460c6862041046ed4/utils/utils.go#L78-L87


rakyll:
Thanks!

PR #416, opened 2018-02-07
Add HTTP attributes

### LINKED ISSUE #385: Add attributes to HTTP spans
Currently, attributes such as HTTP URL, method, request size, response size and error status are not automatically added to the HTTP spans. Add the attributes and provide support from the exporters.

rakyll:
Fixes #385.

### DIFF
--- a/exporter/stackdriver/trace_proto.go
+++ b/exporter/stackdriver/trace_proto.go
@@ -21,6 +21,7 @@ import (
 	"unicode/utf8"
 
 	"go.opencensus.io/internal"
+	"go.opencensus.io/plugin/ochttp"
 
 	timestamppb "github.com/golang/protobuf/ptypes/timestamp"
 	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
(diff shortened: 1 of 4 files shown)


rakyll:
I will also add a test, please take a look at the APIs for now.

exporter/stackdriver/trace_proto.go
```diff
@@ -163,26 +173,51 @@ func copyAttributes(out **tracepb.Span_Attributes, in map[string]interface{}) {
...
-		case bool:
-			av.Value = &tracepb.AttributeValue_BoolValue{BoolValue: value}
-		case int64:
-			av.Value = &tracepb.AttributeValue_IntValue{IntValue: value}
-		case string:
-			av.Value = &tracepb.AttributeValue_StringValue{StringValue: trunc(value, maxAttributeStringValue)}
+		av := attributeValue(value)
+		switch key {
+		case httptrace.URL:
+			(*out).AttributeMap[labelHTTPURL] = &av
```
odeke-em: Looks iffy as we are repeating the pointer creation in every case, perhaps we can reverse this and do it once e.g
```go
func attributeValue(v interface{}) *tracepb.AttributeValue
```

and then we don't have to take the address in every spot.

plugin/http/httptrace/httptrace.go
```diff
@@ -27,6 +27,18 @@ import (
...
 // TODO(jbd): Add godoc examples.
 
+// Attributes recorded on the span for the requests.
+const (
+	URL          = "http.url"
+	Host         = "http.host"
+	Method       = "http.method"
+	Path         = "http.path"
+	UserAgent    = "http.user_agent"
+	StatusCode   = "http.status"
```
odeke-em: Where are we getting the enumeration of these attributes?

odeke-em: I mention this because "http.status_code", "http.status" could even be different according to Go's HTTP API

PR #417, opened 2018-02-07
Fix gofmt issue in httpstats test

ferhatelmas:


ferhatelmas:
@odeke-em yeap, automatic check would be much better.

PR #427, opened 2018-02-08
Make minor improvements to make godoc better

rakyll:
Minor improvements such as adding some whitespace and associating godoc examples with exported names to improve the godoc.


ramonza:
I don't mind changing the examples, I previously changed them to just Example() so that they would appear at the top of the godoc but happy to go with the more specific attachments if that's more idiomatic.

rakyll:
@ramonza, I can keep the package-level example for the stackdriver package but the other package will have more examples. That's why I removed the example to its specific location. I will revert the location of the stackdriver example.

rakyll:
Made the changes, thanks!

Issue #426, opened 2018-02-08
Provide tracing and stats enabling/disabling options for ochttp Handler


rakyll:
NewHandler doesn't allow users to enable/disable tracing or stats. Give users options to enable/disable.

PR #428, opened 2018-02-08
Export ochttp.Handler to provide options on the type

rakyll:
Required for #426.

### DIFF
--- a/plugin/ochttp/example_test.go
+++ b/plugin/ochttp/example_test.go
@@ -15,9 +15,11 @@
 package ochttp_test
 
 import (
+	"log"
 	"net/http"
 
 	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/plugin/ochttp/propagation/google"
 )
 
 func ExampleTransport() {
@@ -25,4 +27,23 @@ func ExampleTransport() {
 		Transport: &ochttp.Transport{},
 	}
 	_ = client // use client to perform requests
+}
+
+var usersHandler http.Handler
+
+func ExampleHandler() {
+	// Enables OpenCensus for the default serve mux.
+	// By default, B3 propagation is used.
+	http.Handle("/users", usersHandler)
+	log.Fatal(http.ListenAndServe("localhost:8080", &ochttp.Handler{}))
+}
+
+func ExampleHandler_mux() {
+	mux := http.NewServeMux()
+	mux.Handle("/users", usersHandler)
+
+	log.Fatal(http.ListenAndServe("localhost:8080", &ochttp.Handler{
+		Handler:     mux,
+		Propagation: &google.HTTPFormat{}, // Uses Google's propagation format.
+	}))
 }
--- a/plugin/ochttp/trace.go
+++ b/plugin/ochttp/trace.go
@@ -132,34 +132,38 @@ func (t *traceTransport) CancelRequest(req *http.Request) {
 	}
 }
 
-// NewHandler returns a http.Handler from the given handler
-// that is aware of the incoming request's span.
-// The span can be extracted from the incoming request in handler
-// functions from incoming request's context:
+// Handler is a http.Handler that is aware of the incoming request's span.
+//
+// The extracted span can be accessed from the incoming request's
+// context.
 //
 //    span := trace.FromContext(r.Context())
 //
-// The span will be automatically ended by the handler.
+// The server span will be automatically ended at the end of ServeHTTP.
 //
 // Incoming propagation mechanism is determined by the given HTTP propagators.
-func NewHandler(base http.Handler, format propagation.HTTPFormat) http.Handler {
-	if format == nil {
-		format = defaultFormat
-	}
-	return &handler{handler: base, format: format}
+type Handler struct {
+	// Propagation defines how traces are propagated. If unspecified,
+	// B3 propagation will be used.
+	Propagation propagation.HTTPFormat
+
+	// Handler is the handler used to handle the incoming request.
+	Handler http.Handler
 }
 
-type handler struct {
-	handler http.Handler
-	format  propagation.HTTPFormat
-}
+// TODO(jbd): Add Handler.NoTrace and Handler.NoStats.
 
-func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	name := spanNameFromURL("Recv", r.URL)
+
+	p := h.Propagation
+	if p == nil {
+		p = defaultFormat
+	}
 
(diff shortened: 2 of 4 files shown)


plugin/ochttp/httptrace.go
```diff
@@ -132,34 +132,39 @@ func (t *traceTransport) CancelRequest(req *http.Request) {
 	}
 }
 
-// NewHandler returns a http.Handler from the given handler
-// that is aware of the incoming request's span.
-// The span can be extracted from the incoming request in handler
-// functions from incoming request's context:
+// NewHandler returns a new handler.
+func NewHandler(base http.HandlerFunc) *Handler {
```
odeke-em: Am afraid changing this signature from http.Handler to http.HandlerFunc will restrict the flexibility and interop with the outside world. It is easier to create an http.Handler than an http.HandlerFunc.
Let's make this
```go
func NewHandler(base http.Handler) *Handler
```
even the function suffix says "Handler" :)

plugin/ochttp/httptrace_test.go
```diff
@@ -139,7 +139,7 @@ func TestHandler(t *testing.T) {
 		t.Run(tt.header, func(t *testing.T) {
 			propagator := &testPropagator{}
 
-			handler := NewHandler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
odeke-em: Please see my suggestion to keep `NewHandler(http.Handler)`

plugin/ochttp/httptrace.go
ramonza: @odeke-em agreed if this is the recommended way to create a Handler then it's odd that it only takes a HandlerFunc not Handler.

plugin/ochttp/httptrace.go
```diff
@@ -169,7 +174,7 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	span.SetAttributes(requestAttrs(r)...)
 
 	r = r.WithContext(ctx)
-	h.handler.ServeHTTP(w, r)
+	h.Handler.ServeHTTP(w, r)
```
ramonza: Should we make the zero value useful by having the default behavior be to call http.DefaultMux with defaultFormat propagation? If we do this, then I think we should also rename or remove NewHandler to address @odeke-em's concern.

odeke-em:
@rakyll we finna need a rebase.

plugin/ochttp/httptrace.go
rakyll: Done

plugin/ochttp/httptrace_test.go
rakyll: Done

plugin/ochttp/trace.go
```diff
@@ -132,34 +132,39 @@ func (t *traceTransport) CancelRequest(req *http.Request) {
...
 
-type handler struct {
-	handler http.Handler
-	format  propagation.HTTPFormat
+	// Handler is the handler used to handle the incoming request.
+	Handler http.Handler
 }
 
-func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+// TODO(jbd): Add Handler.NoTrace and Handler.NoStats.
```
odeke-em: Aha cool, I was actually just looking for the link of the code that @ramonza added to the client with NoTrace and NoStats.

plugin/ochttp/trace.go
```diff
@@ -132,34 +132,39 @@ func (t *traceTransport) CancelRequest(req *http.Request) {
...
 // Incoming propagation mechanism is determined by the given HTTP propagators.
-func NewHandler(base http.Handler, format propagation.HTTPFormat) http.Handler {
-	if format == nil {
-		format = defaultFormat
-	}
-	return &handler{handler: base, format: format}
-}
+type Handler struct {
+	// Propagation defines how traces are propagated. If unspecified,
+	// B3 propagation will be used.
```
odeke-em: Where in the code do we fallback to B3 propagation if the zero value of Propagation is present?

plugin/ochttp/trace.go
rakyll: Ack

plugin/ochttp/trace.go
rakyll: Done

rakyll:
PTAL!

plugin/ochttp/example_test.go
```diff
@@ -26,3 +27,17 @@ func ExampleTransport() {
 	}
 	_ = client // use client to perform requests
 }
+
+var usersHandler http.Handler
+
+func ExampleHandler() {
+	// Enables OpenCensus for the default serve mux.
+	http.Handle("/users", usersHandler)
+	log.Fatal(http.ListenAndServe("localhost:8080", &ochttp.Handler{}))
```
odeke-em: Nice!!

plugin/ochttp/trace.go
odeke-em: Awesome, thank you!

plugin/ochttp/example_test.go
```diff
@@ -26,3 +27,17 @@ func ExampleTransport() {
...
+}
+
+func ExampleHandler_mux() {
+	mux := http.NewServeMux()
+	mux.Handle("/users", usersHandler)
+	log.Fatal(http.ListenAndServe("localhost:8080", &ochttp.Handler{Handler: mux}))
```
odeke-em: Could you add a Propagation here to exhibit that modification?

plugin/ochttp/example_test.go
odeke-em: This one looks alright without a Propagation, but let's modify the comment 
```go
// Enables OpenCensus for the default serve mux.
```
to say
```go
// Enables OpenCensus for the default serve mux and B3 Propagation, the defaults
```

plugin/ochttp/example_test.go
rakyll: Done

plugin/ochttp/example_test.go
rakyll: Done

plugin/ochttp/trace.go
```diff
@@ -132,34 +132,43 @@ func (t *traceTransport) CancelRequest(req *http.Request) {
 	}
 }
 
-// NewHandler returns a http.Handler from the given handler
-// that is aware of the incoming request's span.
-// The span can be extracted from the incoming request in handler
-// functions from incoming request's context:
+// NewHandler returns a new handler.
+func NewHandler(base http.Handler) *Handler {
```
ramonza: Is this needed?

rakyll: Removed

PR #435, opened 2018-02-09
Add godoc examples for exporters

rakyll:


exporter/jaeger/example_test.go
```diff
@@ -0,0 +1,35 @@
...
+	// Register the Jaeger exporter to be able to retrieve
+	// the collected spans.
+	exporter, err := jaeger.NewExporter(jaeger.Options{
+		Endpoint:    "http://localhost:14268",
+		ServiceName: "trace-demo",
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	trace.RegisterExporter(exporter)
```
odeke-em: defer trace.UnregisterExporter(exporter)

exporter/prometheus/example_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"go.opencensus.io/exporter/prometheus"
+	"go.opencensus.io/stats"
+)
+
+func Example() {
+	exporter, err := prometheus.NewExporter(prometheus.Options{})
+	if err != nil {
+		log.Fatal(err)
+	}
+	stats.RegisterExporter(exporter)
```
odeke-em: defer stats.UnregisterExporter(exporter)

Also am not sure why we are putting this in a goroutine in this example, let's remove this, please.

exporter/zipkin/example_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"log"
+
+	openzipkin "github.com/openzipkin/zipkin-go"
+	"github.com/openzipkin/zipkin-go/reporter/http"
+	"go.opencensus.io/exporter/zipkin"
+	"go.opencensus.io/trace"
+)
+
+func Example() {
+	// import (
```
odeke-em: How come we still have these comments in?

exporter/jaeger/example_test.go
rakyll: We don't expect users to unregister often, hence I am leaving unregistering out.

exporter/prometheus/example_test.go
rakyll: Ditto

Removed the goroutine.

exporter/zipkin/example_test.go
rakyll: I manually added them because it is not clear otherwise what package openzipkin actually is.

exporter/jaeger/example_test.go
odeke-em: Hmm, it's nonetheless a good pattern to enforce since many people will be doing copy and paste from examples.

exporter/zipkin/example_test.go
odeke-em: Oh I see, thanks for the update.

exporter/zipkin/example_test.go
```diff
@@ -0,0 +1,40 @@
...
+	//     openzipkin "github.com/openzipkin/zipkin-go"
+	//     "github.com/openzipkin/zipkin-go/reporter/http"
+	//     "go.opencensus.io/exporter/trace/zipkin"
+	// )
+
+	localEndpoint, err := openzipkin.NewEndpoint("server", "server:5454")
```
adriancole: maybe encourage using ip addresses? the local endpoint is which IP it is listening on (I know it could be many, but typically it is the link-local IP (looked up implicitly). The port part is more rarely used, but useful if known

exporter/jaeger/example_test.go
ramonza: What's the consequence of non unregistering? What is the use case for unregistering an exporter? I can't think of a reason anyone would want to do it and it complicates the implementation. I think we should consider removing the functionality wholly.

exporter/prometheus/example_test.go
ramonza: maybe add a comment saying something like "// register other handlers" because currently only the metrics handler is registered and then the program blocks forever which doesn't seem useful

exporter/jaeger/example_test.go
rakyll: I agree with this. Dynamically adding new backends and removing them might be a use case but I don't think it is a critical or common one.

exporter/prometheus/example_test.go
rakyll: I think it is expected to start a new server for the diagnostics reasons and not expose them on your default server.

exporter/jaeger/example_test.go
odeke-em: Dynamically adding exporters for a time period, rare case, for sure, but nonetheless the ability to undo a registration should still be left in. For exporters that we add and perhaps they use a lot of resources and need to be shut off, I think leaving trace.UnregisterExporter should still stay. An example is during root cause analyis or heavy traffic, your web app has an endpoint to trigger exporting for a period of time and when done, shut it off -- 2 years ago we had such a case with bugsnag in which there was a buggy exponential backoff implementation, that was triggered by the same heavy videos being transcoded. Instead of redeploying the entire app after instrumenting it, there was an endpoint to trigger exporting and then shut it off.

Issue #424, opened 2018-02-08
Fix the naming of exported measures and views from ochttp


rakyll:
For example, ClientRequest should be called ClientRequestCount, ClientRequestCount should be called ClientRequestCountView.

Name the measures and views consistently with the gRPC measures/views. https://godoc.org/go.opencensus.io/plugin/grpc/grpcstats.

There is no need to have multiple naming conventions.

ramonza:
ClientRequestCount as a view doesn't seem right. "Count" is an aggregation function so part of the view, not the measure. I will think about it some more. 

I also think that the views should maybe be all collected into a single `var DefaultViews []*View` to make it clearer that these are just defaults, not the only ones that you can define.

rakyll:
Maybe it should be called ClientRequests.

Regardless of what we are doing here, we should apply it to the grpcstats packages. It is confusing to have inconsistent naming of measures/views in the same project.

ramonza:
Agh, wrong issue number in Fixes comment

PR #449, opened 2018-02-13
Refactor gRPC integration into a single package.

ramonza:
This follows the convention established by ochttp. Trace and stats integrations are all under one package and both are enabled by default. To disable one or the other, just set NoStats or NoTrace as appropriate. 

Also renamed all views to remove the now-redundant "cumulative" suffix. Removed the "RPC" prefix from all variable names and fixed up the godoc to look like ochttp's godoc for variables.

Fixes: #424


plugin/ocgrpc/client_metrics.go
```diff
@@ -0,0 +1,91 @@
...
+// package. These are declared as a convenience only; none are subscribed by
+// default.
+var (
+	ClientErrorCountView, _ = view.New(
+		"grpc.io/client/error_count",
+		"RPC Errors",
+		[]tag.Key{KeyStatus, KeyMethod},
+		ClientErrorCount,
+		view.MeanAggregation{})
+	ClientRoundTripLatencyView, _ = view.New(
```
rakyll: Add a new line between these views please. It is easier to read them.

plugin/ocgrpc/client_metrics.go
```diff
@@ -0,0 +1,91 @@
...
+		ClientRequestCount,
+		DefaultMessageCountDistribution)
+	ClientResponseCountView, _ = view.New(
+		"grpc.io/client/response_count",
+		"Count of response messages per client RPC",
+		[]tag.Key{KeyMethod},
+		ClientResponseCount,
+		DefaultMessageCountDistribution)
+
+	DefaultClientViews = []*view.View{
```
rakyll: This can be a var group of its own and can be documented.

plugin/ocgrpc/client.go
```diff
@@ -0,0 +1,72 @@
...
+	clientTrace clientTraceHandler
+	clientStats clientStatsHandler
+)
+
+func (c *ClientHandler) HandleConn(ctx context.Context, cs stats.ConnStats) {
+	// no-op
+}
+
+func (c *ClientHandler) TagConn(ctx context.Context, cti *stats.ConnTagInfo) context.Context {
+	// no-op
```
rakyll: We might now handle tag propagation here rather than at the clientStatsHandler.

Ditto for the server-side. Maybe add a TODO?

ramonza: Isn't this called for each underlying transport connection (HTTP)? If so, I don't think this is useful for tag propagation. Going to leave as-is.

PR #452, opened 2018-02-13
Make starting a custom span more straightforward.

ramonza:
* Deprecate convenience methods for starting a span that
  are unlikely to be useful to the average user.
* Turn on the local spanstore for all spans (including custom
  spans) when the z-Pages are imported.
* Add documentation about the performance impact of z-Pages

Fixes: #424

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -102,13 +102,7 @@ func WithSpan(parent context.Context, s *Span) context.Context {
 
 // StartOptions contains options concerning how a span is started.
 type StartOptions struct {
-	// Deprecated: this option has no effect.
-	RecordEvents bool
-
 	Sampler Sampler // if non-nil, the Sampler to consult for this span.
-
-	// Deprecated: this option has no effect.
-	RegisterNameForLocalSpanStore bool
 }
 
 // TODO(jbd): Remove start options.
@@ -129,7 +123,7 @@ func StartSpan(ctx context.Context, name string) (context.Context, *Span) {
 //
 // If there is no span in the context, creates a new trace and span.
 //
-// Deprecated: use StartSpan(...), or WithSpan(ctx, NewSpan(...)).
+// Deprecated: Use StartSpan(...), or WithSpan(ctx, NewSpan(...)).
 func StartSpanWithOptions(ctx context.Context, name string, o StartOptions) (context.Context, *Span) {
 	parentSpan, _ := ctx.Value(contextKey{}).(*Span)
 	span := NewSpan(name, parentSpan, o)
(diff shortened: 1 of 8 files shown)


trace/trace.go
```diff
@@ -98,24 +102,23 @@ func WithSpan(parent context.Context, s *Span) context.Context {
 
 // StartOptions contains options concerning how a span is started.
 type StartOptions struct {
-	// RecordEvents indicates whether to record data for this span, and include
-	// the span in a local span store.
-	// Events will also be recorded if the span will be exported.
+	// Deprecated: this option has no effect.
 	RecordEvents bool
```
rakyll: We still can remove the deprecated options.

trace/trace.go
```diff
@@ -125,6 +128,8 @@ func StartSpan(ctx context.Context, name string) (context.Context, *Span) {
 // StartSpanWithOptions starts a new child span of the current span in the context.
 //
 // If there is no span in the context, creates a new trace and span.
+//
+// Deprecated: use StartSpan(...), or WithSpan(ctx, NewSpan(...)).
```
rakyll: Deprecated: Use StartSpan or WithSpan.

Typical punctuation rules. Start with capital after `:` if it is a sentence.

We refer to public identifiers with their name only.

trace/trace.go
```diff
@@ -135,6 +140,8 @@ func StartSpanWithOptions(ctx context.Context, name string, o StartOptions) (con
 //
 // If there is an existing span in ctx, it is ignored -- the returned Span is a
 // child of the span specified by parent.
+//
+// Deprecated: use WithSpan(ctx, NewSpanWithRemoteParent(...)).
```
rakyll: Ditto.

trace/trace.go
```diff
@@ -98,24 +102,23 @@ func WithSpan(parent context.Context, s *Span) context.Context {
 
 // StartOptions contains options concerning how a span is started.
 type StartOptions struct {
-	// RecordEvents indicates whether to record data for this span, and include
-	// the span in a local span store.
-	// Events will also be recorded if the span will be exported.
+	// Deprecated: this option has no effect.
```
rakyll: // Deprecated: This option has no effect.

I'd prefer we remove these, no one should currently be depending on these fields directly.

trace/trace.go
```diff
@@ -98,24 +102,23 @@ func WithSpan(parent context.Context, s *Span) context.Context {
...
 	RecordEvents bool
 
 	Sampler Sampler // if non-nil, the Sampler to consult for this span.
 
-	// RegisterNameForLocalSpanStore indicates that a local span store for spans
-	// of this name should be created, if one does not exist.
-	// If RecordEvents is false, this option has no effect.
+	// Deprecated: this option has no effect.
```
rakyll: Ditto.

trace/trace.go
ramonza: ok

trace/trace.go
ramonza: TIL!

trace/trace.go
ramonza: ok

Issue #436, opened 2018-02-09
Add a flag to gRPC and HTTP server instrumentation to indicate whether to trust the caller


ramonza:
As discussed in-person with @bogdandrutu, right now we always trust the headers/metadata passed to HTTP/gRPC server integration. In most cases this is fine (if you are a backend service or your LB strips all unknown headers) but in many cases it isn't acceptable to just accept the SpanContext from the incoming requests which is untrusted.

We should add an option in the HTTP and gRPC server integrations to trust or not trust the incoming headers. In HTTP, this might be handled by just configuring an empty slice of propagation.HTTPFormat, but we should document this.

rakyll:
We discussed offline that we are not sure if the lack of trust should be the default case.

gRPC servers are secure by default and we might need to understand more of the canonical way of exposing a gRPC server to public before finalizing this option.

rakyll:
Closed via the IsPublicEndpoint field on the plugins.

PR #454, opened 2018-02-15
Make gRPC and HTTP server plugins more configurable

ramonza:
* Add options to the gRPC and HTTP server plugins to ignore headers/metadata present on the request
* Allow specifying a Sampler to be used for Spans started by the plugins

Fixes: #436


bogdandrutu:
Please split this PR into 2:
1) which applies the "don't trust"
2) which removes old APIs.

ramonza:
@bogdandrutu sure

ramonza:
@bogdandrutu split into two commits, and added "do not squash" label

plugin/ocgrpc/server.go
```diff
@@ -36,12 +31,20 @@ type ServerHandler struct {
...
 	// gRPC method.
 	NoStats bool
+
+	// Set to true if trace metadata should be read from inbound requests.
+	// Only set this if the current service will only be invoked by
+	// services under your control. Otherwise, a malicious user might be able
+	// to trigger traces in your backend.
+	//
+	// If not set, any trace information in the inbound request will be ignored.
+	TrustCallerMetadata bool
```
rakyll: Why would anyone use the stats handler directly in a public API without taking any security considerations? The security concern here is beyond what OpenCensus propagates but applies to all metadata.

Are we expecting a public-facing server to disable this?

Why not default being the opposite?

s/TrustCallerMetadata/TrustCaller?

// TrustCallerMetadata trust the trace metadata of the incoming calls.
// ... 
TrustCallerMetadata bool

golint, godoc should start with exported identifier.

ramonza: changed to default trust for gRPC.
yes, public-facing servers should not trust tracing metadata I think. Ultimately it's up to the developer to decide but this gives them a way to not trust but still create server-side spans.

plugin/ocgrpc/client.go
```diff
@@ -38,10 +38,7 @@ type ClientHandler struct {
 	NoStats bool
 }
 
-var (
-	clientTrace clientTraceHandler
-	clientStats clientStatsHandler
-)
+var _ stats.Handler = &ClientHandler{}
```
odeke-em: var _ stats.Handler = (*ClientHandler)(nil)

plugin/ocgrpc/server.go
```diff
@@ -36,12 +31,20 @@ type ServerHandler struct {
...
 	// gRPC method.
 	NoStats bool
+
+	// NoReadTraceContext may be set to true to avoid reading trace metadata
+	// from inbound requests.
+	//
+	// You should set this if your gRPC server is a public-facing service and
+	// your load balancer does not strip metadata. Otherwise, a malicious user
+	// will be able to trigger traces in your backend.
+	NoReadTraceContext bool
```
odeke-em: NoTraceContextReads?

plugin/ocgrpc/server.go
```diff
@@ -36,12 +31,20 @@ type ServerHandler struct {
...
+
+	// Sampler to use for requests handled by this server.
+	Sampler trace.Sampler
 }
 
-var (
-	serverTrace serverTraceHandler
-	serverStats serverStatsHandler
-)
+var _ stats.Handler = &ServerHandler{}
```
odeke-em: var _ stats.Handler = (*ServerHandler)(nil)

plugin/ocgrpc/trace_common.go
```diff
@@ -55,26 +47,32 @@ func (c *clientTraceHandler) TagRPC(ctx context.Context, rti *stats.RPCTagInfo)
...
-	if s := md[traceContextKey]; len(s) > 0 {
-		if parent, ok := propagation.FromBinary([]byte(s[0])); ok {
-			ctx, _ = trace.StartSpanWithRemoteParent(ctx, name, parent, trace.StartOptions{})
-			return ctx
+	options := trace.StartOptions{Sampler: s.Sampler}
+	traceContext := md[traceContextKey]
+	if !s.NoReadTraceContext && len(traceContext) > 0 {
+		// Metadata with keys ending in -bin are actually binary. They are base64
+		// encoded before being put on the wire, see:
+		// https://github.com/grpc/grpc-go/blob/08d6261/Documentation/grpc-metadata.md#storing-binary-data-in-metadata
```
odeke-em: Where are we checking that the keys end in `-bin`? If we aren't, perhaps let's update this comment.

plugin/ocgrpc/trace_common.go
```diff
@@ -55,26 +47,32 @@ func (c *clientTraceHandler) TagRPC(ctx context.Context, rti *stats.RPCTagInfo)
...
+		traceContextBinary := []byte(traceContext[0])
+		if parent, ok := propagation.FromBinary(traceContextBinary); ok {
+			span := trace.NewSpanWithRemoteParent(name, parent, trace.StartOptions{})
+			return trace.WithSpan(ctx, span)
 		}
 	}
-	ctx, _ = trace.StartSpan(ctx, name)
```
odeke-em: Why this change? Don't they do the same thing? Moreover it seems like now are making two calls instead of just one.

plugin/ochttp/default_format.go
```diff
@@ -19,4 +19,6 @@ import (
 	"go.opencensus.io/trace/propagation"
 )
 
-var defaultFormat propagation.HTTPFormat = &b3.HTTPFormat{}
+var (
+	defaultOutboundFormat propagation.HTTPFormat = &b3.HTTPFormat{}
```
odeke-em: Minor nit: perhaps we don't need the brackets and new indentation?

plugin/ochttp/trace.go
```diff
@@ -143,41 +143,65 @@ func (t *traceTransport) CancelRequest(req *http.Request) {
...
+	// http.DefaultServeMux will be used.
 	Handler http.Handler
+
+	// Sampler to use to decide whether a new trace should be started with the
+	// server span as the root span.
+	//
+	// Sampler will be consulted if either no inbound trace information was read
+	// from the request, or if the request was not sampled.
+	//
+	// If not set, the default sampler will be used (see
```
odeke-em: Please keep the `(see` and rest of the comment on the same line.

plugin/ochttp/trace.go
```diff
@@ -143,41 +143,65 @@ func (t *traceTransport) CancelRequest(req *http.Request) {
...
+		span = trace.NewSpanWithRemoteParent(name, sc, trace.StartOptions{Sampler: h.Sampler})
 	} else {
-		ctx, span = trace.StartSpan(ctx, name)
+		span = trace.NewSpan(name, nil, trace.StartOptions{Sampler: h.Sampler})
 	}
+	ctx = trace.WithSpan(ctx, span)
+	r = r.WithContext(ctx)
 	defer span.End()
 
 	span.SetAttributes(requestAttrs(r)...)
```
odeke-em: Does it matter that we invoke SetAttributes after trace.WithSpan?

adriancole:
Food for thought in general as I don't have specific code comments.

I aimed to do something similar in Brave (java zipkin tracer) for http
abstraction, but havent commited (literally or figuratively) here's why..

Sensitivity of trace data vs tags is a thing, and why is different. Ex
limiting incoming trace to prevent abuse (like always using the same trace
id) or making it easier to map the system. Tags for also abuse even if by
accident.

It seemed like trust is a configuration which is both direction and
endpoint specific (both incoming and outgoing unless you rely on transport
for this). Ex I may distrust or disable outbound propagation to a third
party api.

I can solve this with layering.. ex if i abstract the propagation keys, a
custom filter can make a security policy of any kind.. simply by choosing
to clean an inbound or outbound request. Has an N framework implication as
such a filter needs to be made for potentially incompatible apis.

I can also solve this by scoping (ex make a trust=false decision then
making a request scoped instrumentation object). More work to propagate a
request scoped tracer and also has N framework implication.

I can solve this by making propagation policy a function of the remote
endpoint. This might be enough for most cases, but no experience trying.

Anyway this stuff was tabled as I am working on other things, but since you
are working on exact same problem, figured I would share what is in my head.

ramonza:
It's clear that there is a lot complexity and nuance here. My worry is that right now we don't provide a way for users to opt out of trusting the headers at all. So either they will not install the instrumentation or they will install it an be insecure. Essentially, there is no correct way to instrument public-facing servers using our gRPC or HTTP integrations.

With this PR, at least there is a way to configure public-facing servers that is both useful (starts root spans) and not insecure. On the other hand, if we prefer to punt on this for now I would be ok replacing this PR with just some warnings in comments not to install instrumentation on public-facing servers. What do you say about this @rakyll @bogdandrutu ?

We don't propagate tags for now so I am less concerned about outbound propagation. We will definitely need to tackle this when we start implementing that.

plugin/ocgrpc/server.go
ramonza: That makes it sounds like "there are not trace context reads" where what I want to say is "don't read the trace context"

plugin/ocgrpc/trace_common.go
ramonza: `traceContextKey` ends in `-bin` - it is a constant

plugin/ochttp/trace.go
ramonza: no, it mutates the existing span through the `*Span`

ramonza:
@odeke-em updated

plugin/ocgrpc/server.go
odeke-em: Gotcha gotcha, ACK.

plugin/ocgrpc/trace_common.go
odeke-em: It's just a little confusing and without context for me as a reader that looked at this PR, without the mention of what uses `-bin`.

ramonza: Oh I see where you're coming from. The reason I added it was that I was confused by why we were just assuming that the value of the metadata (which is typed as string) is actually the raw bytes and not hex encoded (as, I would argue, the type would suggest). So I added this documentation to clarify the situation. I didn't introduce the use of -bin in this PR.

ramonza:
@odeke-em updated PTAL. I renamed the field in the gRPC ServerHandler to emphasize that setting it to true means that we start a new trace and leaving if false means you trust the metadata

odeke-em:
Thanks @ramonza, sorry for the late response and for blocking the merge of this. Btw, this PR needs a rebase.

ramonza:
I am in the process of splitting this PR up

odeke-em:
Some rebase might be needed here if this PR is still active.

ramonza:
I ended up splitting this into separate PRs.

Issue #402, opened 2018-02-04
Race in prometheus exporter


rakyll:
See the log:

```
fatal error: concurrent map iteration and map write

goroutine 198 [running]:
runtime.throw(0x1587a35, 0x26)
	/Users/jbd/go/src/runtime/panic.go:619 +0x81 fp=0xc4205d4e08 sp=0xc4205d4de8 pc=0x102be41
runtime.mapiternext(0xc4205d4f28)
	/Users/jbd/go/src/runtime/hashmap.go:747 +0x55c fp=0xc4205d4e98 sp=0xc4205d4e08 pc=0x100aacc
go.opencensus.io/exporter/prometheus.(*collector).Collect(0xc4201a86c0, 0xc420324120)
	/Users/jbd/src/go.opencensus.io/exporter/prometheus/prometheus.go:206 +0x92 fp=0xc4205d4f98 sp=0xc4205d4e98 pc=0x13f1db2
github.com/prometheus/client_golang/prometheus.(*Registry).Gather.func2(0xc4203ec0d0, 0xc420324120, 0x15da7e0, 0xc4201a86c0)
	/Users/jbd/src/github.com/prometheus/client_golang/prometheus/registry.go:382 +0x61 fp=0xc4205d4fc0 sp=0xc4205d4f98 pc=0x13ef191
runtime.goexit()
	/Users/jbd/go/src/runtime/asm_amd64.s:2361 +0x1 fp=0xc4205d4fc8 sp=0xc4205d4fc0 pc=0x10573a1
created by github.com/prometheus/client_golang/prometheus.(*Registry).Gather
	/Users/jbd/src/github.com/prometheus/client_golang/prometheus/registry.go:380 +0x302

goroutine 1 [sleep]:
time.Sleep(0x5f5e100)
	/Users/jbd/go/src/runtime/time.go:102 +0x166
main.main()
	/Users/jbd/src/github.com/rakyll/opencensus-grpc-demo/client/main.go:81 +0x598

goroutine 20 [select]:
go.opencensus.io/stats.(*worker).start(0xc4200ae180)
	/Users/jbd/src/go.opencensus.io/stats/worker.go:221 +0x11d
created by go.opencensus.io/stats.init.0
	/Users/jbd/src/go.opencensus.io/stats/worker.go:29 +0x54

goroutine 38 [select]:
google.golang.org/grpc.(*ccResolverWrapper).watcher(0xc4201f5ad0)
[...]

PR #456, opened 2018-02-16
prom: fixes dynamic registration of views bug

rdallman:
prometheus's default registry appears to block registration of any other
metrics based on there being >= 1 duplicate in the set of metrics returned
from Describe to register. While Describe's API seems to indicate that a
collector should return a superset of metrics, and while this makes perfect
sense with the Unregister logic, it does not seem to play nicely with us
running register on the same collector repeatedly with different metrics, nor
with calling unregister and then attempting to register everything once again.
that is to say, prometheus client seems to explicitly disallow the thing we're
seemingly trying to do. upstream: prometheus/client_golang#47

this patch will:
* no longer call Unregister and then Register, as this is a case handled by
prometheus: https://github.com/prometheus/client_golang/blob/fcc130e101e76c5d303513d0e28f4b6d732845c7/prometheus/registry.go#L112-L117
* only return metrics that are not yet registered to prometheus registry

since ExportView appears to be called in a single thread, I am not sure
the locks are required, but have abided here. that also ruled out this
being a race of any kind.

It's also a perfectly reasonable explanation that I've not properly registered
views. fwiw am doing `v.Subscribe()` for every view in init() atm, and the
intent appears to be to be able to toggle these on/off during runtime.

I was able to generate these errors consistently on the every sweep of
metrics that would be done:

```
cannot register the collector: descriptor Desc{fqName: "redacted", help:
"redacted", constLabels: {}, variableLabels: []} already exists with the same
fully-qualified name and const label values
```

to be clear, this is not due to my attempting to register 2 views / metrics
under the same name, but from registerViews repeatedly registering all metrics
to prometheus registry (confirmed) which will [trivially] contain duplicates.

I apologize that I do not have a reduced reproduction. I do however have a
patch in an OSS project that will reliably reproduce this, though I hate
pointing at that as it'll take some orienting to run.

perhaps these were simply an annoyance and to be ignored, in which case we can
just remove the `onError` bits, I am seeing all metrics in the prometheus
endpoint even though I get this error spat out a lot (and, it's not clear that
this should work this way after reading the prometheus register code). there
may be other solutions :)

thanks for this library, your efforts are appreciated, am looking forward to
using it!

### DIFF
--- a/exporter/prometheus/prometheus.go
+++ b/exporter/prometheus/prometheus.go
@@ -104,7 +104,7 @@ func (c *collector) registerViews(views ...*view.View) {
 				nil,
 			)
 			c.registeredViewsMu.Lock()
-			c.registeredViews[sig] = desc
+			c.unregisteredViews[sig] = desc
 			c.registeredViewsMu.Unlock()
 			count++
 		}
@@ -113,7 +113,6 @@ func (c *collector) registerViews(views ...*view.View) {
 		return
 	}
 
-	c.reg.Unregister(c)
 	if err := c.reg.Register(c); err != nil {
 		c.opts.onError(fmt.Errorf("cannot register the collector: %v", err))
 	}
@@ -163,6 +162,10 @@ type collector struct {
 	registeredViewsMu sync.Mutex
 	// registeredViews maps a view to a prometheus desc.
 	registeredViews map[string]*prometheus.Desc
+
+	// unregisteredViews is the set of views that are
+	// queued to be registered with prometheus.
+	unregisteredViews map[string]*prometheus.Desc
 }
 
 func (c *collector) addViewData(vd *view.Data) {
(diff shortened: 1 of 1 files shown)


exporter/prometheus/prometheus.go
```diff
@@ -163,6 +162,10 @@ type collector struct {
 	registeredViewsMu sync.Mutex
 	// registeredViews maps a view to a prometheus desc.
 	registeredViews map[string]*prometheus.Desc
+
+	// unregisteredViews is the set of views that are
```
rakyll: s/unregisteredViews/newViews

Unregistered sounds like they have been registered and unregistered at some point.

exporter/prometheus/prometheus.go
```diff
@@ -163,6 +162,10 @@ type collector struct {
...
+	// queued to be registered with prometheus.
```
rakyll: s/with prometheus/with Prometheus

exporter/prometheus/prometheus.go
```diff
@@ -174,11 +177,19 @@ func (c *collector) addViewData(vd *view.Data) {
 	c.mu.Unlock()
 }
 
+// NOTE that this will not work properly if calling
```
rakyll: Can you move this comment inside the function block?

exporter/prometheus/prometheus.go
```diff
@@ -104,7 +104,7 @@ func (c *collector) registerViews(views ...*view.View) {
 				nil,
 			)
 			c.registeredViewsMu.Lock()
```
rakyll: Can we call registeredViewsMu to viewsMu and document is guards both fields?

rdallman:
addressed, thank you for the review

odeke-em:
> since ExportView appears to be called in a single thread, I am not sure
the locks are required, but have abided here. that also ruled out this
being a race of any kind.

c.viewData can be read/written concurrently by multiple goroutines during a Collect cycle when a user visits /metrics concurrently while ExportView is invoked internally, hence why we need the lock. One can easily reproduce such races as with test https://github.com/census-instrumentation/opencensus-go/issues/402#issuecomment-362905246 and the bug linked there was a real bug reported by @rakyll

PR #463, opened 2018-02-21
Fix stats new measure error handling style

odeke-em:
instead of

```go
_, err := foo()
if err != nil {
    return nil, err
} else {
    return m, err
}
```

simplify that to the idiomatic style
```go
if _, err := foo(); err != nil {
    return nil, err
}
return m, nil
```
and notice that in the return we no longer do `return m, err`
where err is nil to begin with, so be more explicit and return
nil for better readability and consistency.

### DIFF
--- a/stats/measure.go
+++ b/stats/measure.go
@@ -43,6 +43,8 @@ var (
 	mu           sync.RWMutex
 	measures     = make(map[string]Measure)
 	errDuplicate = errors.New("duplicate measure name")
+
+	errMeasureNameTooLong = fmt.Errorf("measure name cannot be longer than %v", internal.MaxNameLength)
 )
 
 func FindMeasure(name string) Measure {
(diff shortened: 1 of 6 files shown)


stats/measure.go
```diff
@@ -43,15 +43,15 @@ var (
 	mu           sync.RWMutex
 	measures     = make(map[string]Measure)
 	errDuplicate = errors.New("duplicate measure name")
+
+	errMeasureNameTooLong = fmt.Errorf("measure name cannot be larger than %v", internal.MaxNameLength)
```
ramonza: s/larger/longer/

odeke-em: Thanks @ramonza, updating in a second.

odeke-em:
The test failures are from unrelated code not passing the linting test.

PR #464, opened 2018-02-21
Reduce the pressure on event channel and avoid tag map alloc on the critical path

rakyll:
The event channel was using a non-buffered channel that blocks until each
event is processed. Switch to a buffered channel to not the block the
writes.

tag.FromContext currently allocates a new tag map if there is none
in the current context. This creates an extra allocation on
the critical path.

Benchmarks before the change:
BenchmarkRecord0-8       2000000               786 ns/op             120 B/op          3 allocs/op
BenchmarkRecord1-8       2000000               950 ns/op             192 B/op          6 allocs/op
BenchmarkRecord8-8       1000000              1647 ns/op             472 B/op         13 allocs/op

Benchmarks after the change:
BenchmarkRecord0-8               5000000               361 ns/op              64 B/op          1 allocs/op
BenchmarkRecord1-8               5000000               366 ns/op             136 B/op          4 allocs/op
BenchmarkRecord8-8               3000000               526 ns/op             416 B/op         11 allocs/op
BenchmarkRecord8_Parallel-8      2000000               881 ns/op             416 B/op         11 allocs/op
BenchmarkRecord8_8Tags-8         3000000               540 ns/op             416 B/op         11 allocs/op

### DIFF
--- a/tag/context.go
+++ b/tag/context.go
@@ -22,7 +22,7 @@ func FromContext(ctx context.Context) *Map {
 	// The returned tag map shouldn't be mutated.
 	ts := ctx.Value(mapCtxKey)
 	if ts == nil {
-		return newMap(0)
+		return nil
 	}
 	return ts.(*Map)
 }
--- a/plugin/ocgrpc/client_stats_handler.go
+++ b/plugin/ocgrpc/client_stats_handler.go
@@ -45,8 +45,10 @@ func (h *clientStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo)
 
 	d := &rpcData{startTime: startTime}
 	ts := tag.FromContext(ctx)
-	encoded := tag.Encode(ts)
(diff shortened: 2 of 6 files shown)


tag/context.go
```diff
@@ -22,7 +22,7 @@ func FromContext(ctx context.Context) *Map {
 	// The returned tag map shouldn't be mutated.
 	ts := ctx.Value(mapCtxKey)
 	if ts == nil {
-		return newMap(0)
```
odeke-em: I thought that the purpose of FromContext is to create an empty map if none exists, and for convenience that way we don't have to always check `if ts != nil`?

rakyll: FromContext should return nil. This gives the user the signal, for example, whether they should Encode and propagate the tags.

OTOH, I made the map methods defensive and they handle the nil value.

rakyll:
PTAL

PR #467, opened 2018-02-21
Add ochttp.Transport benchmarks

ramonza:
```
$ git describe && go test  -run=^$ -bench=.
v0.1.0-101-g68ae3ab
goos: darwin
goarch: amd64
pkg: go.opencensus.io/plugin/ochttp
BenchmarkTransportNoInstrumentation-8              	   20000	     74474 ns/op	    4857 B/op	      64 allocs/op
BenchmarkTransportTraceOnly-8                      	   20000	     81016 ns/op	    6649 B/op	      98 allocs/op
BenchmarkTransportStatsOnly-8                      	   20000	     87528 ns/op	    6754 B/op	      95 allocs/op
BenchmarkTransportAllInstrumentation-8             	   20000	     92464 ns/op	    8554 B/op	     129 allocs/op
BenchmarkTransportParallel64NoInstrumentation-8    	  100000	     16560 ns/op	    4912 B/op	      62 allocs/op
BenchmarkTransportParallel64TraceOnly-8            	  100000	     18791 ns/op	    6732 B/op	      96 allocs/op
BenchmarkTransportParallel64StatsOnly-8            	  100000	     18601 ns/op	    6859 B/op	      93 allocs/op
BenchmarkTransportParallel64AllInstrumentation-8   	  100000	     20737 ns/op	    8681 B/op	     127 allocs/op
--- BENCH: BenchmarkTransportParallel64AllInstrumentation-8
	client_test.go:218: Get: Get http://127.0.0.1:64470: read tcp 127.0.0.1:64475->127.0.0.1:64470: read: connection reset by peer
PASS
ok  	go.opencensus.io/plugin/ochttp	18.405s
```

### DIFF
--- a/plugin/ochttp/client_test.go
+++ b/plugin/ochttp/client_test.go
@@ -16,6 +16,7 @@ package ochttp_test
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"strings"
@@ -118,4 +119,55 @@ func TestClient(t *testing.T) {
 			t.Fatalf("%s = %d; want %d", viewName, got, reqCount)
 		}
 	}
+}
+
+func BenchmarkTransportNoInstrumentation(b *testing.B) {
+	benchmarkClientServer(b, &ochttp.Transport{NoStats: true, NoTrace: true})
+}
+
+func BenchmarkTransportTraceOnly(b *testing.B) {
+	benchmarkClientServer(b, &ochttp.Transport{NoStats: true})
+}
+
+func BenchmarkTransportStatsOnly(b *testing.B) {
+	benchmarkClientServer(b, &ochttp.Transport{NoTrace: true})
+}
+
+func BenchmarkTransportAllInstrumentation(b *testing.B) {
+	benchmarkClientServer(b, &ochttp.Transport{})
+}
+
+// Copied from net/http/serve_test.go
+func benchmarkClientServer(b *testing.B, transport http.RoundTripper) {
+	b.ReportAllocs()
+	b.StopTimer()
+	ts := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(rw, "Hello world.\n")
+	}))
+	defer ts.Close()
+	var client http.Client
+	client.Transport = transport
+	b.StartTimer()
+
+	for i := 0; i < b.N; i++ {
+		get, err := http.NewRequest("GET", ts.URL, nil)
+		if err != nil {
+			b.Fatalf("NewRequest: %v", err)
+		}
+		res, err := client.Do(get)
+		if err != nil {
+			b.Fatalf("Get: %v", err)
+		}
+		all, err := ioutil.ReadAll(res.Body)
+		res.Body.Close()
+		if err != nil {
+			b.Fatal("ReadAll:", err)
+		}
+		body := string(all)
+		if body != "Hello world.\n" {
+			b.Fatal("Got body:", body)
+		}
+	}
+
+	b.StopTimer()
 }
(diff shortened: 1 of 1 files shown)


plugin/ochttp/client_test.go
```diff
@@ -119,3 +120,115 @@ func TestClient(t *testing.T) {
...
+
+func BenchmarkTransportParallel64StatsOnly(b *testing.B) {
+	benchmarkClientServerParallel(b, 64, &ochttp.Transport{NoTrace: true})
+}
+
+func BenchmarkTransportParallel64AllInstrumentation(b *testing.B) {
+	benchmarkClientServerParallel(b, 64, &ochttp.Transport{})
+}
+
+func benchmarkClientServerParallel(b *testing.B, parallelism int, transport *ochttp.Transport) {
```
rakyll: Want to call parallelism concurrency or c? We have n number of concurrent goroutines that might be only scheduled on one thread.

plugin/ochttp/client_test.go
```diff
@@ -119,3 +120,115 @@ func TestClient(t *testing.T) {
...
+func benchmarkClientServer(b *testing.B, transport http.RoundTripper) {
+	b.ReportAllocs()
+	b.StopTimer()
+	ts := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(rw, "Hello world.\n")
+	}))
+	defer ts.Close()
+	var client http.Client
+	client.Transport = transport
+	b.StartTimer()
```
odeke-em: Why not just b.ResetTimer() and remove the b.StopTimer() above?

plugin/ochttp/client_test.go
```diff
@@ -119,3 +120,115 @@ func TestClient(t *testing.T) {
...
+		if err != nil {
+			b.Fatal("ReadAll:", err)
+		}
+		body := string(all)
+		if body != "Hello world.\n" {
+			b.Fatal("Got body:", body)
+		}
+	}
+
+	b.StopTimer()
```
odeke-em: Unnecessary

plugin/ochttp/client_test.go
ramonza: I don't know, this code was copied from the standard library HTTP tests so didn't really want to change it

plugin/ochttp/client_test.go
ramonza: I agree that concurrency is probably a better fit but the testing package and net/http where this was copied from calls it parallelism so I'm inclined to leave it as-is

plugin/ochttp/client_test.go
ramonza: Ugh, don't know where my previous reply went but just to clarify again this is copied from the net/http test so I want to leave it as-is

plugin/ochttp/client_test.go
odeke-em: I mean but we can fix that though, even stdlib code could use such fixes :)

ramonza:
@odeke-em not worth making changes since the code was copied from somewhere else

plugin/ochttp/client_test.go
ramonza: I guess I'm not convinced that ResetTimer is better? I don't really care TBH but I also don't want people to see the comment above that says "copied from ..." and wonder why I changed this - it might cause them to think there was some reason for it.

ramonza:
@odeke-em PTAL, I changed to ResetTimer

plugin/ochttp/client_test.go
```diff
@@ -119,3 +120,111 @@ func TestClient(t *testing.T) {
...
+		MaxIdleConns:        parallelism,
+		MaxIdleConnsPerHost: parallelism,
+	}
+	c.Transport = transport
+
+	b.ResetTimer()
+
+	// TODO(ramonza): replace with b.RunParallel (it didn't work when I tried)
+
+	var wg sync.WaitGroup
```
odeke-em: Why didn't we just use a channel? I believe we can radically simplify the code below like this:

```go
semaphore := make(chan bool, concurrency)
for i := 0; i < b.N; i++ {
     <-semaphore
     go func() {
          defer func() { semaphore <- true }()
          res, err := c.Get(...)
     }()
}
```

PR #468, opened 2018-02-21
 Collect float64 in Measurement

ramonza:
This reduces allocations per stats.Record operation substantially:
```
git describe && go test  -run=^$ -benchmem -bench=.
v0.1.0-103-g23a8069
goos: darwin
goarch: amd64
pkg: go.opencensus.io/stats
BenchmarkRecord-8   	  500000	      2239 ns/op	     392 B/op	       5 allocs/op
PASS
ok  	go.opencensus.io/stats	1.163s
```
Compare: https://github.com/census-instrumentation/opencensus-go/wiki/stats-Benchmarks

### DIFF
--- a/stats/benchmark_test.go
+++ b/stats/benchmark_test.go
@@ -6,7 +6,7 @@ import (
 	"testing"
 
 	"go.opencensus.io/stats"
-	_ "go.opencensus.io/stats/view"
+	_ "go.opencensus.io/stats/view" // enable collection
 )
 
 var m = makeMeasure()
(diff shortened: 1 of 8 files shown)


odeke-em:
@ramonza perhaps in the future, it might help to include the benchmarks from before and after and then using https://godoc.org/golang.org/x/perf/cmd/benchstat for comparisons. Thank you for getting this in.

ramonza:
@odeke-em thanks for pointing that out, I will definitely do that next time!

PR #472, opened 2018-02-22
Revert Prometheus PR #456 but instead register only once

### LINKED ISSUE #469: Prometheus exporter example returning errors to /metrics endpoint
**What did you do?**
Downloaded and executed [this example](https://github.com/census-instrumentation/opencensus-go/blob/master/examples/stats/prometheus/main.go), then pulled up the localhost:9999/metrics endpoint in Chrome.

**What did you expect to see?**
There should be output along the lines of:
```
opencensus_video_count 33
opencensus_video_cum_bucket{le="0"} 0
opencensus_video_cum_bucket{le="65536"} 0
opencensus_video_cum_bucket{le="4.294967296e+09"} 0
opencensus_video_cum_bucket{le="+Inf"} 17
opencensus_video_cum_sum 6.671402338140933e+19
opencensus_video_cum_count 17
```

**What did you see instead? Under which circumstances?**
```
An error has occurred during metrics gathering:

2 error(s) occurred:
* collected metric opencensus_video_count counter:<value:225 >  was collected before with the same name and label values
* collected metric opencensus_video_cum histogram:<sample_count:225 sample_sum:1.0183722549263512e+21 bucket:<cumulative_count:0 upper_bound:0 > bucket:<cumulative_count:0 upper_bound:65536 > bucket:<cumulative_count:0 upper_bound:4.294967296e+09 > >  was collected before with the same name and label values
```

**More context**
I tried commenting out either of the stats.Record statements at the bottom of the file, and it seems like each one works individually. The problem seems to happen when measurements are recorded for more than one measure.

I followed the example to include OpenCensus in a Go project of my own and also encountered the same thing. Recording measurements of a single measure look fine, but then when a second measure's measurement is recorded I see errors.

**Environment**

* System information:
Darwin 17.4.0 x86_64

* Package versions:
go.opencensus.io v0.3.0
prometheus client_golang v0.8.0

odeke-em:
Fixes #469
Reverts #456

Also ensures that we only register the collector exactly
once, of which we erraneously had assumed Unregister would
do but as documented at
https://github.com/prometheus/client_golang/blob/fcc130e101e76c5d303513d0e28f4b6d732845c7/prometheus/registry.go#L89-L101
which specifies that invoking Register again even after Unregister
will just spit out an error because a collector should only
be registered once.

PR #456 attempted to fix this but didn't maintain uniqueness of
registeredViews hence doubly or more registered Prometheus metrics.
Nonetheless it was a great attempt.

With this PR, we should never get back the error that prompted
the PR in revision.

### DIFF
--- a/examples/stats/prometheus/main.go
+++ b/examples/stats/prometheus/main.go
@@ -75,7 +75,7 @@ func main() {
 		"processed video size over time",
 		nil,
 		videoSize,
-		view.DistributionAggregation([]float64{1 << 10, 1<<15, 1 << 20, 1 << 25}),
+		view.DistributionAggregation([]float64{0, 1 << 16, 1 << 32}),
 	)
 	if err != nil {
 		log.Fatalf("Cannot create view: %v", err)
@@ -92,10 +92,9 @@ func main() {
 
 	// Record some data points...
 	go func() {
-		rand.Seed(time.Now().Unix())
 		for {
 			stats.Record(ctx, videoCount.M(1))
-			stats.Record(ctx, videoSize.M(int64(rand.ExpFloat64() * (1<<24))))
+			stats.Record(ctx, videoSize.M(rand.Int63()))
 			<-time.After(time.Millisecond * time.Duration(1+rand.Intn(400)))
 		}
 	}()
--- a/examples/stats/prometheus/main.go
+++ b/examples/stats/prometheus/main.go
@@ -75,7 +75,7 @@ func main() {
 		"processed video size over time",
 		nil,
 		videoSize,
-		view.DistributionAggregation([]float64{0, 1 << 16, 1 << 32}),
+		view.DistributionAggregation([]float64{1 << 10, 1<<15, 1 << 20, 1 << 25}),
 	)
 	if err != nil {
 		log.Fatalf("Cannot create view: %v", err)
@@ -92,9 +92,10 @@ func main() {
 
 	// Record some data points...
 	go func() {
+		rand.Seed(time.Now().Unix())
 		for {
 			stats.Record(ctx, videoCount.M(1))
-			stats.Record(ctx, videoSize.M(rand.Int63()))
+			stats.Record(ctx, videoSize.M(int64(rand.ExpFloat64() * (1<<24))))
 			<-time.After(time.Millisecond * time.Duration(1+rand.Intn(400)))
 		}
 	}()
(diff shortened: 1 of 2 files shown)


rdallman:
that makes sense, thanks. it was hard to distill the prometheus collector contract (still is). I've tested with this patch and see all metrics I expect sans the errors I was previously, thank you for fixing this and sorry for breaking things!

examples/stats/prometheus/main.go
```diff
@@ -75,7 +75,7 @@ func main() {
 		"processed video size over time",
 		nil,
 		videoSize,
-		view.DistributionAggregation([]float64{0, 1 << 16, 1 << 32}),
+		view.DistributionAggregation([]float64{1 << 10, 1<<15, 1 << 20, 1 << 25}),
```
rakyll: Can you send this changes in a different PR? They are not related to the bug fix.

exporter/prometheus/prometheus.go
```diff
@@ -103,19 +103,32 @@ func (c *collector) registerViews(views ...*view.View) {
...
 
-	if err := c.reg.Register(c); err != nil {
-		c.opts.onError(fmt.Errorf("cannot register the collector: %v", err))
-	}
+	c.ensureRegisteredOnce()
+}
+
+// ensureRegisteredOnce invokes reg.Register on the collector itself
+// exactly once to ensure that we don't get errors such as
+//    cannot register the collector: descriptor Desc{fqName: *}
```
rakyll: Bad indentation in the comments.

exporter/prometheus/prometheus.go
```diff
@@ -270,10 +271,10 @@ func tagsToLabels(tags []tag.Tag) []string {
 
 func newCollector(opts Options, registrar *prometheus.Registry) *collector {
 	return &collector{
+		registerOnce:    sync.Once{},
```
rakyll: You can remove this line.

examples/stats/prometheus/main.go
odeke-em: ACK, thanks.

exporter/prometheus/prometheus.go
odeke-em: That was sort of intentional, but perhaps I overdid the indentation ;)

exporter/prometheus/prometheus.go
odeke-em: Oops, I left that in by mistake, thanks, I'll remove it.

odeke-em:
> that makes sense, thanks. it was hard to distill the prometheus collector contract (still is). I've tested with this patch and see all metrics I expect sans the errors I was previously, thank you for fixing this and sorry for breaking things!

@rdallman, thanks for the investigation and for the patch, it helped me narrow down the issue and it's cool that you are using OpenCensus!

odeke-em:
@rakyll I've addressed your feedback, PTAL.

PR #474, opened 2018-02-22
Add microbenchmarks for span creation and annotation

rakyll:
BenchmarkStartEndSpan-8                   10000000  185 ns/op   128 B/op    2 allocs/op
BenchmarkSpanWithAnnotations_3-8   5000000   338 ns/op   224 B/op   5 allocs/op
BenchmarkSpanWithAnnotations_6-8   3000000 . 475 ns/op   320 B/op   8 allocs/op

### DIFF
--- a/trace/benchmark_test.go
+++ b/trace/benchmark_test.go
@@ -1,0 +1,63 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


rakyll:
Just FYI, the trace package is impacting the performance more significantly on the critical path than the stats package right now.

ramonza:
These number look good but this is without linking the zpages so not recording events

PR #475, opened 2018-02-22
Add missing license headers and lint

rakyll:


rakyll:
Haha, we have tons of lint errors coming from internal packages and the autogenerated thrift code. I suggest using VSCode because it underlines the lint errors and you can selectively fix :P

PR #476, opened 2018-02-23
Add ochttp examples

rakyll:


examples/grpc/helloworld_client/main.go
```diff
@@ -20,7 +20,7 @@ import (
 	"os"
 	"time"
 
-	"go.opencensus.io/examples/grpc/exporter"
+	"go.opencensus.io/examples/exporter"
```
odeke-em: Wait, does this code compile?

examples/grpc/helloworld_server/main.go
```diff
@@ -21,7 +21,7 @@ import (
 	"net"
 	"net/http"
 
-	"go.opencensus.io/examples/grpc/exporter"
```
odeke-em: Ditto, was this perhaps a stray edit?

examples/http/helloworld_client/main.go
```diff
@@ -0,0 +1,53 @@
...
+	view.RegisterExporter(exporter)
+	trace.RegisterExporter(exporter)
+
+	view.SetReportingPeriod(1 * time.Second)
+
+	client := &http.Client{
+		Transport: &ochttp.Transport{},
+	}
+
+	for {
```
odeke-em: Please remove this for loop, this is a tight loop, why not just one get
```go
res, err := client.Get(url)
...
```

examples/grpc/helloworld_client/main.go
rakyll: Yes, I moved the exporter package.

odeke-em: Oh I see, sorry for the false alarm
screen shot 2018-02-22 at 4 37 58 pm

examples/http/helloworld_client/main.go
rakyll: Done.

rakyll:
PTAL

PR #479, opened 2018-02-23
ochttp: OpenCensus stats for net/http serving.

fluffle:


fluffle:
Hello! I noticed this feature was missing a few days ago and reached out to @rakyll internally :-)

fluffle:
Ooo. I did not run go vet before sending this PR. My apologies, I'll fix that next week!

plugin/ochttp/stats_transport.go
```diff
@@ -79,12 +79,44 @@ func (t statsTransport) CancelRequest(req *http.Request) {
 	}
 }
 
+// StatsHandler is an http.Handler that updates OpenCensus HTTP stats for
+// requests served by a net/http server.
+type StatsHandler struct {
```
ramonza: For gRPC we moved from having a separate stats and tracing handler to having a combined one (with the ability to opt-out of stats or tracing using booleans). We reasoned that in almost all cases, there's no reason not to configure both stats and tracing. The reason both stats and tracing are part of OpenCensus is that there are some really cool  things you can do when you have both, like exemplars: https://github.com/census-instrumentation/opencensus-specs/issues/46 - so we're kind of pushing people to use both together. What do you think?

plugin/ochttp/stats_transport.go
```diff
@@ -79,12 +79,44 @@ func (t statsTransport) CancelRequest(req *http.Request) {
...
+	if r.Body == nil {
+		// TODO: Handle cases where ContentLength is not set.
+		track.reqSize = -1
+	} else if r.ContentLength > 0 {
+		track.reqSize = r.ContentLength
+	}
+	stats.Record(ctx, ServerRequestCount.M(1))
+
+	// Perform request.
+	h.Handler.ServeHTTP(track, r)
```
ramonza: We use the convention that a zero-valued http.Handler should mean http.DefaultServeMux, see the existing ochttp.Handler

plugin/ochttp/stats_transport.go
```diff
@@ -103,7 +135,22 @@ func (t *tracker) end() {
...
 }
 
+func (t *tracker) endServe() {
+	t.endOnce.Do(func() {
+		m := []stats.Measurement{
+			ServerLatency.M(float64(time.Since(t.start)) / float64(time.Millisecond)),
+			ServerResponseBytes.M(t.respSize),
+		}
+		if t.reqSize >= 0 {
+			m = append(m, ServerRequestBytes.M(t.reqSize))
```
ramonza: Minor nit but the asymmetry between when we record the response size (always, even if there is not response body as in a 204 response) and when we record the request body (only if size>0) seems a little odd.

plugin/ochttp/stats_transport.go
```diff
@@ -79,12 +79,44 @@ func (t statsTransport) CancelRequest(req *http.Request) {
...
+	track.endServe()
+}
+
 type tracker struct {
 	respSize   int64
 	reqSize    int64
 	ctx        context.Context
 	start      time.Time
 	body       io.ReadCloser
+	writer     http.ResponseWriter
```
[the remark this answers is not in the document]
fluffle: I originally thought there would be a load of overlap in functionality. A few of the struct fields are shared, but they are used in different ways, so it's probably safer to separate out the functionality.

plugin/ochttp/stats_transport.go
fluffle: Not a concern post-refactor.

plugin/ochttp/stats_transport.go
fluffle: My contrary opinion was: I don't understand why you would coupled them :-)

The http.Handler interface means that if you want both pieces of functionality it is trivial to wrap one in the other. But if you're incentivising use by opting users into all functionality as standard, it makes sense to have everything in one place. I'll ditch the separate handler.

I didn't want to move the stats functionality into trace.go because that seemed counter-intuitive, so instead I've done a minor bit of refactoring. Hope that's ok?

plugin/ochttp/stats_transport.go
fluffle: I don't disagree, but this was cargo-culted from the roundtripper above, so it at least has the virtue of consistency ;-)

I don't see a meaningful difference between not recording a size at all and recording a size of zero.

plugin/ochttp/stats_test.go
```diff
@@ -19,9 +23,111 @@ func TestVarsInitialized(t *testing.T) {
 			t.Errorf("measure not initialized: %d", i)
 		}
 	}
+	for i, m := range []stats.Measure{ServerRequestCount, ServerResponseBytes, ServerRequestBytes, ServerLatency} {
```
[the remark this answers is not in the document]
fluffle: Uh. They are. And so are the ones above. I guess I should delete both of these then!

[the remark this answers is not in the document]
fluffle: No, wait. These are measures, not views. But the views that use them are in DefaultViews, so there's that.

fluffle:
PR updated, please LMK if there's anything else.

plugin/ochttp/stats_test.go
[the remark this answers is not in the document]
odeke-em: @fluffle if the goal here is to ensure that those measures are always set for a user, if we forget to run the tests
and something haywire occurs, we'll be in bigger trouble. Thus, perhaps we should instead make a small helper
```go
func mustMeasureInt64(name, description, unit string) *stats.Int64 {
      m, err := stats.Int64(name, description, unit)
      if err != nil {
          log.Fatalf("Failed to create view with name=%q description=%q unit=%q err: %v", name, description, unit)
      }
}
```
ditto for the Float64Measure

plugin/ochttp/stats_test.go
```diff
@@ -19,9 +23,112 @@ func TestVarsInitialized(t *testing.T) {
...
+		if got, want := len(rows), 1; got != want {
+			t.Errorf("len(%q) = %d; want %d", viewName, got, want)
+			continue
+		}
+		data := rows[0].Data
+		var count int
+		var sum float64
+		switch data := data.(type) {
+		case *view.CountData:
+			count = int(*(*int64)(data))
```
odeke-em: Spurious conversion, let's make this
```go
count = int(*data)
```

plugin/ochttp/stats_test.go
```diff
@@ -19,9 +23,112 @@ func TestVarsInitialized(t *testing.T) {
...
+		case *view.DistributionData:
+			count = int(data.Count)
+			sum = data.Sum()
+		default:
+			t.Errorf("don't know how to handle data type: %v", data)
+		}
+		if got, want := count, totalCount; got != want {
+			t.Fatalf("%s = %d; want %d", viewName, got, want)
+		}
+		if got, want := sum, meanReqSize*float64(totalCount); viewName == "opencensus.io/http/server/request_bytes" && got != want {
```
odeke-em: I think we can make this code easier to follow, with:
```go
switch viewName {
case "opencensus.io/http/server/request_bytes":
     if got, want := sum, meanReqSize*float64(totalCount); got != want {...}

case "opencensus.io/http/server/response_bytes":
     if got, want := sum, meanRespSize*float64(totalCount); got != want {...}

default:
      t.Errorf("Stray view %q", viewName)
}
```

plugin/ochttp/trace.go
```diff
@@ -157,31 +157,48 @@ type Handler struct {
...
 	handler.ServeHTTP(w, r)
+	if traceEnd != nil {
+		traceEnd.End()
+	}
+	if statsEnd != nil {
+		statsEnd.End()
+	}
+}
+
+func (h *Handler) wrapForTrace(w http.ResponseWriter, r *http.Request) (http.ResponseWriter, *http.Request, ender) {
```
odeke-em: Perhaps registerReceiverSpan?

I couldn't make out what wrapForTrace intended to convey.

plugin/ochttp/stats_transport.go
```diff
@@ -126,3 +126,75 @@ func (t *tracker) Close() error {
 	t.end()
 	return t.body.Close()
 }
+
+func (h *Handler) wrapForStats(w http.ResponseWriter, r *http.Request) (http.ResponseWriter, *http.Request, ender) {
```
odeke-em: Ditto here about the name `wrapForStats`

plugin/ochttp/trace.go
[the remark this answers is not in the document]
odeke-em: Nice!

plugin/ochttp/stats_test.go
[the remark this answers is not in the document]
fluffle: Again, you're reading a little too much into what is effectively a copy/paste operation. I can certainly refactor as you suggest, though. Arguably these are generally useful functions for opencensus users, so I will put them in their various packages as MustVariants.

plugin/ochttp/stats_test.go
fluffle: Oops, thanks!

plugin/ochttp/trace.go
fluffle: The incoming Request and ResponseWriter are being wrapped in a new context and the tracking writer. eh, naming is hard.

I'll go with startTrace per @rakyll

plugin/ochttp/stats_transport.go
```diff
@@ -126,3 +126,75 @@ func (t *tracker) Close() error {
...
+		stats.Record(ctx, m...)
+	})
+}
+
+func (t *trackingResponseWriter) Header() http.Header {
+	return t.writer.Header()
+}
+
+func (t *trackingResponseWriter) Write(data []byte) (int, error) {
+	if t.statusCode == "" {
```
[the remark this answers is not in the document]
fluffle: It was a string in the tracker, so I just copied that. It needs to be a string eventually because it's inserted into a tag. No reason not to do the conversion up-front.

plugin/ochttp/stats_transport.go
```diff
@@ -126,3 +126,75 @@ func (t *tracker) Close() error {
...
+		t.statusCode = "200"
+	}
+	n, err := t.writer.Write(data)
+	if err == nil {
```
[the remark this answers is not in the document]
fluffle: This mimics the behaviour of Read() but I guess the semantics are different there.

plugin/ochttp/stats_transport.go
```diff
@@ -126,3 +126,75 @@ func (t *tracker) Close() error {
...
+
+func (t *trackingResponseWriter) End() {
+	t.endOnce.Do(func() {
+		m := []stats.Measurement{
+			ServerLatency.M(float64(time.Since(t.start)) / float64(time.Millisecond)),
+			ServerResponseBytes.M(t.respSize),
+		}
+		if t.reqSize >= 0 {
+			m = append(m, ServerRequestBytes.M(t.reqSize))
+		}
```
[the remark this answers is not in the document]
fluffle: Done.

plugin/ochttp/trace.go
```diff
@@ -157,31 +157,48 @@ type Handler struct {
 	Handler http.Handler
 }
 
+type ender interface {
```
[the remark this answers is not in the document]
fluffle: That's definitely neater, thanks!

plugin/ochttp/trace.go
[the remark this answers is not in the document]
fluffle: Ack.

plugin/ochttp/stats_transport.go
fluffle: Done.

plugin/ochttp/stats_test.go
fluffle: We can't have the default case be an error, since we look at more than these two views. But otherwise I agree this is neater, thanks :-)

fluffle:
PTAL? I think I've addressed everything.

fluffle:
I've rebased onto latest master.

stats/measure_int64.go
```diff
@@ -44,3 +46,13 @@ func Int64(name, description, unit string) (*Int64Measure, error) {
 	}
 	return m, nil
 }
+
+// MustInt64 creates a new measure of type Int64Measure. It will log a
+// fatal error if a measure with the same name already exists.
```
ramonza: Should we panic rather than fatal like regexp.MustCompile, template.Must?

tag/key.go
```diff
@@ -15,20 +15,33 @@
...
 	return Key{name: name}, nil
 }
 
+// MustKey creates or retrieves a string key identified by name.
+// Calling MustKey (or NewKey) with the same name returns the same key.
+// It will log a fatal error if the key name is invalid.
+func MustKey(name string) Key {
+	k, err := NewKey(name)
+	if err != nil {
+		log.Fatalf("Failed to create tag key with name=%q: %v", name, err)
```
ramonza: Ditto

stats/measure_int64.go
odeke-em: @ramonza for sure let's panic then, thanks for this catch. My bad, I was the one who suggested https://github.com/census-instrumentation/opencensus-go/pull/479#discussion_r171699498 log.Fatalf but that was for the internal helper functions.

plugin/ochttp/stats.go
```diff
@@ -22,25 +22,33 @@ import (
 
 // The following client HTTP measures are supported for use in custom views:
 var (
-	ClientRequestCount, _  = stats.Int64("opencensus.io/http/client/request_count", "Number of HTTP requests started", stats.UnitNone)
-	ClientRequestBytes, _  = stats.Int64("opencensus.io/http/client/request_bytes", "HTTP request body size if set as ContentLength (uncompressed)", stats.UnitBytes)
-	ClientResponseBytes, _ = stats.Int64("opencensus.io/http/client/response_bytes", "HTTP response body size (uncompressed)", stats.UnitBytes)
-	ClientLatency, _       = stats.Float64("opencensus.io/http/client/latency", "End-to-end latency", stats.UnitMilliseconds)
+	ClientRequestCount  = stats.MustInt64("opencensus.io/http/client/request_count", "Number of HTTP requests started", stats.UnitNone)
```
rakyll: I don't think we should increase the API surface with stats.MustX and tag.MustX.

For hardcoded values, it is ok to avoid the error. We do it all the time for the http.NewRequest call.

You will often see, error is avoided for hardcoded values that are not going to fail:

```
req, err := http.NewRequest("GET", "http://example.com", nil)
```

If we don't do that, we are almost deprecating NewX functions because they are in practice useful in rare cases, e.g. if you are dynamically creating keys dynamically.

odeke-em: I concur about not increasing the API surface layer; I had suggested those functions as internal helper functions.

@rakyll as for the example
> You will often see, error is avoided for hardcoded values that are not going to fail:
> req, err := http.NewRequest("GET", "http://example.com", nil)
req, err := http.NewRequest("GET", uri, nil)

requires an error because there are lots of cases where you'll pass in a URL that isn't known at compile time and if that URL fails to parse you'll get an error for example when downloading URLs constructed at runtime, in a proxy, on a backend etc.

rakyll: Btw, what I meant above was actually:

```
req, _ := http.NewRequest("GET", "http://example.com", nil)
```

odeke-em: Gotcha gotcha, in deed, thanks.

rakyll:
I am taking over this PR and addressing the issues pointed above.

rakyll:
Please review https://github.com/census-instrumentation/opencensus-go/pull/534.

odeke-em:
@rakyll took over it with https://github.com/census-instrumentation/opencensus-go/pull/534. Thank you @fluffle for your work on it and @rakyll for finishing it up!

PR #486, opened 2018-02-23
Add Sampler to ochttp.Transport

ramonza:
Add an explicit Sampler to ochttp.Transport. Use this Sampler in
benchmarks to ensure we are measuring the full cost of sampled
traces.


plugin/ochttp/trace.go
```diff
@@ -40,7 +40,7 @@ const (
 )
 
 type traceTransport struct {
-	base   http.RoundTripper
+	*Transport
```
rakyll: Can we instead have a Sampler field? Circular dependencies between traceTransport and Transport and statsTransport and Transport are not ideal.

ramonza:
Updated, @rakyll PTAL

Issue #470, opened 2018-02-22
exporter/prometheus: make an end-to-end test that parses wire output from /metrics


odeke-em:
https://github.com/census-instrumentation/opencensus-go/issues/469 shows that we need an end-to-end test that parses the output from the /metrics endpoint or so, and successfully parses it to ensure that we never regress when we make changes to the Prometheus exporter as happened in https://github.com/census-instrumentation/opencensus-go/pull/456

odeke-em:
@freddygv would like to work on this issue: @freddygv, all yours please feel free to take it and no pressure if you need to ask questions or defer it to any of us, we value your contributions, thank you!

odeke-em:
Fixed by https://github.com/census-instrumentation/opencensus-go/pull/488

PR #488, opened 2018-02-25
Issue #470 e2e test prometheus exporter

freddygv:
#469 surfaced a regression in the prometheus exporter. 

This PR adds an end to end test for the exporter by parsing the output of the /metrics endpoint.

It tests for:
* Non-empty output in /metrics
* Duplicated name/labels like in #469
* Any other error posted from the Prometheus registry


exporter/prometheus/prometheus_test.go
```diff
@@ -174,3 +179,94 @@ func TestCollectNonRacy(t *testing.T) {
...
+	}
+
+	vc := &vCreator{}
+	for _, m := range measures {
+		vc.createAndSubscribe(m.Name(), m.Description(), nil, m, view.CountAggregation{})
+	}
+	if vc.err != nil {
+		t.Fatalf("failed to create views: %v", err)
+	}
+	view.SetReportingPeriod(time.Second)
```
ramonza: Since this is a test, why not set this much shorter so the test runs more quickly? How about using time.Millisecond?

exporter/prometheus/prometheus_test.go
```diff
@@ -174,3 +179,94 @@ func TestCollectNonRacy(t *testing.T) {
...
+			t.Fatalf("failed to get /metrics: %v", err)
+		}
+		defer resp.Body.Close()
+
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatalf("failed to read body: %v", err)
+		}
+		output = string(body)
+		time.Sleep(time.Second)
```
ramonza: Ditto

exporter/prometheus/prometheus_test.go
freddygv: Sure thing!

exporter/prometheus/prometheus_test.go
```diff
@@ -15,12 +15,17 @@
 package prometheus
```
rakyll: Missing license header.

exporter/prometheus/prometheus_test.go
```diff
@@ -174,3 +179,94 @@ func TestCollectNonRacy(t *testing.T) {
...
+		defer resp.Body.Close()
+
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatalf("failed to read body: %v", err)
+		}
+		output = string(body)
+		time.Sleep(time.Millisecond)
+	}
+
```
rakyll: I'd also check whether the measure names are in the results and their types.

ramonza:
looks good but I'm not that familiar with the original issue so @odeke-em would you mind taking a look?

exporter/prometheus/prometheus_test.go
```diff
@@ -15,12 +15,17 @@
 package prometheus
 
 import (
+	"context"
```
odeke-em: "golang.org/x/net/context" and please move it an import a line after the first stdlib import.

exporter/prometheus/prometheus_test.go
freddygv: Not sure what you mean here. Is it just a matter of updating the year from 2017 to 2018?

exporter/prometheus/prometheus_test.go
freddygv: Thanks! Will do.

exporter/prometheus/prometheus_test.go
```diff
@@ -174,3 +179,94 @@ func TestCollectNonRacy(t *testing.T) {
...
+	if vc.err != nil {
+		t.Fatalf("failed to create views: %v", err)
+	}
+	view.SetReportingPeriod(time.Millisecond)
+
+	for _, m := range measures {
+		stats.Record(context.Background(), m.M(1))
+	}
+
+	addr := "localhost:9999"
```
odeke-em: Let's make a server manually i.e.
```go
import "net/http/httptest"

srv := httptest.NewServer(exporter)
defer srv.Close()

for ... {
    resp, err := http.Get(srv.URL)
}
```

No need for the http.ListenAndServe in the goroutine below.

exporter/prometheus/prometheus_test.go
```diff
@@ -174,3 +179,94 @@ func TestCollectNonRacy(t *testing.T) {
...
+		if i == 10000 {
+			t.Fatal("no output at /metrics (10s wait)")
+		}
+		i++
+
+		resp, err := http.Get("http://localhost:9999/metrics")
+		if err != nil {
+			t.Fatalf("failed to get /metrics: %v", err)
+		}
+		defer resp.Body.Close()
```
odeke-em: Let's close this body right after reading instead of in defers.

exporter/prometheus/prometheus_test.go
odeke-em: In deed, what @rakyll said^

exporter/prometheus/prometheus_test.go
odeke-em: @rakyll it exists already, this isn't a new file :)
@freddygv it looked like a new file, and Github's code review doesn't let us expand
files easily.

exporter/prometheus/prometheus_test.go
freddygv: Oooh, nice. Was wondering how else I could that. Thanks!

exporter/prometheus/prometheus_test.go
freddygv: Why not use the context in the stdlib?

freddygv: @odeke-em ^

odeke-em: @freddygv we still need to support gRPC which uses go1.6,go1.7 which use "golang.org/x/net/context"

ramonza: @odeke-em we support 1.8+ only, we use "context" in a number of places so I think it's fine

odeke-em: Oh cool, thank you @ramonza for the update. I'll also page @rakyll to ask if this intentional.

freddygv:
Signed the CLA again, git was pointing at a different email

exporter/prometheus/prometheus_test.go
```diff
@@ -174,3 +180,102 @@ func TestCollectNonRacy(t *testing.T) {
...
+	if vc.err != nil {
+		t.Fatalf("failed to create views: %v", err)
+	}
+	view.SetReportingPeriod(time.Millisecond)
+
+	for _, m := range measures {
+		stats.Record(context.Background(), m.M(1))
+	}
+
+	http.Handle("/metrics", exporter)
```
odeke-em: Please remove this as it is no longer necessary.

exporter/prometheus/prometheus_test.go
```diff
@@ -174,3 +180,102 @@ func TestCollectNonRacy(t *testing.T) {
...
+	if strings.Contains(output, "error(s) occurred") {
+		t.Fatal("error reported by prometheus registry")
+	}
+
+	for _, name := range names {
+		if !strings.Contains(output, "opencensus_tests_"+name+" 1") {
+			t.Fatalf("measurement missing in output: %v", name)
+		}
+	}
+
```
odeke-em: Spurious newline, please remove.

exporter/prometheus/prometheus_test.go
```diff
@@ -174,3 +180,102 @@ func TestCollectNonRacy(t *testing.T) {
...
+		}
+
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatalf("failed to read body: %v", err)
+		}
+		resp.Body.Close()
+
+		output = string(body)
+		time.Sleep(time.Millisecond)
```
odeke-em: As soon as we get the output, let's break
```go
for {
    if i == 10000 {
        ...
    }
    resp, err := http.Get(srv.URL)
    ...
    if output != "" {
        break
    }
    time.Sleep(time.Millisecond)
}
```

exporter/prometheus/prometheus_test.go
```diff
@@ -183,3 +189,102 @@ func TestCollectNonRacy(t *testing.T) {
...
+	for _, m := range measures {
+		stats.Record(context.Background(), m.M(1))
+	}
+
+	srv := httptest.NewServer(exporter)
+	defer srv.Close()
+
+	var i int
+	var output string
+	for output == "" {
```
odeke-em: From the last review, we should have made this a plain `for`
as below we check `if output != "" { break }` but that's okay :)

freddygv: Ah, thanks for spotting that. Fixing now.

odeke-em:
Perfect, thank you very much @freddygv!

PR #492, opened 2018-02-26
stats/view ergonomics improvements

ramonza:
* Make Subscribe a function that takes a variable number of views
* Add View.WithName() to rename a view
* Defer errors from view.New() to subscription time
* Deprecate: view functions Register, Unregister, and Find


examples/stats/prometheus/main.go
```diff
@@ -39,51 +39,30 @@ func main() {
...
 		view.DistributionAggregation([]float64{0, 1 << 16, 1 << 32}),
 	)
-	if err != nil {
-		log.Fatalf("Cannot create view: %v", err)
-	}
 
 	// Subscribe will allow view data to be exported.
 	// Once no longer needed, you can unsubscribe from the view.
-	if err := viewSize.Subscribe(); err != nil {
+	if err := view.Subscribe(viewCount, viewSize); err != nil {
```
rakyll: With this new API, it is hard to programmatically handle the associate which view the error is associated with.

stats/view/worker.go
```diff
@@ -104,6 +87,24 @@ func (v *View) Subscribe() error {
...
+// you may call Unsubscribe to stop collecting data for the view.
+func Subscribe(views ...*View) error {
+	var errstr []string
+	for _, v := range views {
+		err := v.Subscribe()
+		if err != nil {
+			errstr = append(errstr, fmt.Sprintf("%s: %v", v.name, err))
+		}
+	}
+	if len(errstr) > 0 {
```
rakyll: This is programmatically impossible to handle.

One of the main reason why we have the subscription model is to enable/disable collection in prod.

How can we programmatically handle subscription errors?

stats/view/worker.go
```diff
@@ -50,8 +51,7 @@ var defaultWorker *worker
 
 var defaultReportingDuration = 10 * time.Second
 
-// Find returns a registered view associated with this name.
-// If no registered view is found, nil is returned.
+// Deprecated: No replacement.
```
rakyll: Find is useful if you need a pointer to an existing View by name and unsubscribe from it. We shouldn't remove it.

stats/view/example_test.go
```diff
@@ -27,17 +27,15 @@ func Example_view() {
 		log.Fatal(err)
 	}
 
-	v, err := view.New(
+	v := view.New(
```
rakyll: Why do we need view.New? We should instead allow users to depend on the struct literal.

ramonza:
Updated with your recommendations @rakyll, want to take a look at https://github.com/census-instrumentation/opencensus-go/pull/492/commits/512c6d3c686e8df8cd125beb909ac3128cc12502

exporter/stackdriver/stats_test.go
```diff
@@ -384,14 +377,7 @@ func TestExporter_makeReq_batching(t *testing.T) {
...
 
-	v, err := view.New("view", "desc", []tag.Key{key}, m, view.CountAggregation{})
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err := view.Register(v); err != nil {
-		t.Fatal(err)
-	}
-	defer view.Unregister(v)
+	v, _ := view.New("view", "desc", []tag.Key{key}, m, view.CountAggregation{})
```
odeke-em: Since view.New is deprecated let's use &view.View{...}

README.md
```diff
@@ -239,6 +199,17 @@ func (e *exporter) ExportView(vd *view.Data) {
 
 ```
 
+Configure the default interval between reports of collected data.
+This is a system wide interval and impacts all views. The default
+interval duration is 10 seconds. Trying to set an interval with
+a duration less than a certain minimum (maybe 1s) should have no effect.
```
odeke-em: Currently there is no limit. Let's remove the  sentence about trying to set an internal with maybe 1s.

README.md
```diff
@@ -119,23 +120,29 @@ if err != nil {
...
 
-Currently 4 types of aggregations are supported. The CountAggregation is used to count
-the number of times a sample was recorded. The DistributionAggregation is used to
-provide a histogram of the values of the samples. The SumAggregation is used to
-sum up all sample values. The MeanAggregation is used to calculate the mean of
-sample values.
+Views are how Measures are aggregated. You can think of them as queries over the
+set of recorded data points (Measurements).
+
+Views have two parts: the tags to group by and the aggregation function used.
```
odeke-em: `aggregation method` used.

README.md
```diff
@@ -145,74 +152,27 @@ sumAgg := view.SumAggregation{}
 meanAgg := view.MeanAggregation{}
 ```
 
-### Creating, registering and unregistering a view
-
-Create and register a view:
+Here we create a view with the DistributionAggregation over our Measure.
+Setting tags to nil (as in this example) means that Measurements with all tags
```
odeke-em: `that all Measurements will be aggregated together, irrespective of their tags (i.e. no grouping by tag)`

stats/view/view.go
```diff
@@ -28,112 +27,109 @@ import (
...
-	// tagKeys to perform the aggregation on.
-	tagKeys []tag.Key
+	// GroupByTags are the tag keys describing the grouping of this view.
+	// A single Row will be produced for each combination of associated tag values.
+	GroupByTags []tag.Key
 
-	// Examples of measures are cpu:tickCount, diskio:time...
-	m stats.Measure
+	// MeasureName is the name of the stats.Measure to aggregate in this view.
+	MeasureName string
```
odeke-em: Let's still use Measure, I don't see the utility of the user invoking m.Name() and then when registering, we also look for the Measure by name. In fact exposing this field as a name `MeasureName` means that some users will accidentally not create measures assuming we'll automagically do it for them, at least I'd fall for this.

stats/view/worker.go
```diff
@@ -104,25 +84,46 @@ func (v *View) Subscribe() error {
...
+	}
+	if len(errstr) > 0 {
+		return errors.New(strings.Join(errstr, "\n"))
+	}
+	return nil
+}
+
 // Unsubscribe unsubscribes a previously subscribed view.
 // Data will not be exported from this view once unsubscription happens.
 func (v *View) Unsubscribe() error {
```
odeke-em: This is now not reflexive, we have a struct with  a deprecated Subscribe method yet an Unsubscribe method, and a global Subscribe function, this is also inconsistent. Let's keep *view.View.Subscribe and still advertise Subscribe as a convenience function for subscribing many views.

stats/view/worker.go
```diff
@@ -187,41 +188,52 @@ func (w *worker) stop() {
...
-		return nil
+		return x, nil
 	}
 
-	if v.Measure() == nil {
-		return fmt.Errorf("cannot register view %q: measure not defined", v.Name())
+	if v.MeasureName == "" {
+		return nil, fmt.Errorf("cannot subscribe view %q: measure not set", v.Name)
 	}
+	m := stats.FindMeasure(v.MeasureName)
```
odeke-em: My point about v.MeasureName applies here given that a user invoke m.Name() to create *view.View.MeasureName and now here we are looking for Measure again

stats/view/view.go
```diff
@@ -28,112 +27,109 @@ import (
...
-	name        string // name of View. Must be unique.
-	description string
+	Name        string // Name of View. Must be unique.
+	Description string // Description is a human-readable description for this view.
 
-	// tagKeys to perform the aggregation on.
-	tagKeys []tag.Key
+	// GroupByTags are the tag keys describing the grouping of this view.
+	// A single Row will be produced for each combination of associated tag values.
+	GroupByTags []tag.Key
```
odeke-em: Too verbose. We've already defined tags as what groups Measurements, we use tag.Keys to group only one way.

```go
TagKeys []tag.Key
```

zpages/rpcz.go
```diff
@@ -58,10 +58,13 @@ var (
 )
 
 func init() {
-	for view := range viewType {
-		if err := view.Subscribe(); err != nil {
-			log.Printf("error subscribing to view %q: %v", view.Name(), err)
-		}
+	views := make([]*view.View, 0, len(viewType))
+	for k := range viewType {
```
odeke-em:
```go
for view := range viewType
```
as it was before is nicer IMHO than `k`

stats/view/view.go
ramonza: Good point. I was wanting to move this further in the direction of allowing views to be defined as configuration rather than code but that is really a separate feature so I'll revert to Measure.

stats/view/worker.go
ramonza: ok, although I think Subscribe as a method (as well as Unsubscribe) is weird because it's really an operation on the package global state rather than on the view on which the method is called

stats/view/view.go
ramonza: What about `GroupBy`? TagKeys just repeats the type name so doesn't really add anything?

ramonza: The reason  I wanted to emphasize grouping is that actually before we had some incorrect documentation that said that these tags were being used to filter, which is obviously wrong.

odeke-em: I like `GroupBy`, cool and more descriptive suggestion!

README.md
```diff
@@ -145,74 +152,27 @@ sumAgg := view.SumAggregation{}
...
-)
-if err != nil {
-	log.Fatalf("cannot create view: %v", err)
-}
-if err := view.Register(v); err != nil {
-	log.Fatal(err)
+v := &view.View{
+	Name:        "my.org/video_size_distribution",
+	Description: "distribution of processed video size over time",
+	MeasureName: videoSize.Name(),
```
rakyll: I'd take the Measure here even though I will internally use the measure name. API should work with the high-level APIs if possible.

zpages/rpcz.go
ramonza: agreed with the name, the reason I changed it was (a) `Subscribe` is deprecated and (b) `view` variable shadows `view` package but I agree `k` is not right

README.md
```diff
@@ -145,74 +152,27 @@ sumAgg := view.SumAggregation{}
...
-```
-
 ### Retrieving collected data for a view
 
 Users need to subscribe to a view in order to retrieve collected data.
 
 [embedmd]:# (stats.go subscribe)
 ```go
-if err := v.Subscribe(); err != nil {
+if err := view.Subscribe(v); err != nil {
```
rakyll: This feels very odd to me now.

view.Subscribe(a view).

I don't know if view.Subscribe(...) adds much.

stats/view/view.go
rakyll: I don't like the use of group by. This is a SQL terminology and not common in other language implementation or in any of the specs or docs. It is more confusing to me than having TagKeys (consistence with other languages) and having godoc explaining what it does.

stats/view/view.go
rakyll: Ditto.

plugin/ochttp/example_test.go
```diff
@@ -20,9 +20,31 @@ import (
...
 	"go.opencensus.io/plugin/ochttp"
 	"go.opencensus.io/plugin/ochttp/propagation/google"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
 )
 
 func ExampleTransport() {
+
+	err := view.Subscribe(
+
```
rakyll: Remove line.

plugin/ochttp/example_test.go
```diff
@@ -20,9 +20,31 @@ import (
 
 	"go.opencensus.io/plugin/ochttp"
 	"go.opencensus.io/plugin/ochttp/propagation/google"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
 )
 
 func ExampleTransport() {
+
```
rakyll: Remove line.

plugin/ochttp/example_test.go
```diff
@@ -20,9 +20,31 @@ import (
...
+	err := view.Subscribe(
+
+		// Subscribe to a few default views, with renaming
```
rakyll: I don't think we should ever have an example of WithName at ochttp, this is such as edge-case feature.

An example can be in the stats package.

stats/example_test.go
```diff
@@ -16,40 +16,11 @@ package stats_test
...
-		log.Fatal(err)
-	}
-	if err := view.Subscribe(); err != nil {
-		log.Fatal(err)
-	}
-
-	// Use stats.RegisterExporter to export collected data.
+func ExampleRecord() {
+	openConns, _ := stats.Int64("my.org/measure/openconns", "open connections", stats.UnitNone)
+	stats.Record(context.TODO(), openConns.M(124)) // Record 124 open connections.
```
rakyll: Can we replace context.TODO() with ctx while you are here.

stats/view/view.go
```diff
@@ -28,112 +27,109 @@ import (
...
+	vNew := *v
+	vNew.Name = name
+	return &vNew
 }
 
-// Description returns the name of the view.
-func (v *View) Description() string {
-	return v.description
+// Equal compares two views and returns true if they represent the same aggregation.
+func (v *View) Equal(other *View) bool {
```
rakyll: Is this useful to the end-user?

examples/stats/prometheus/main.go
```diff
@@ -39,51 +39,28 @@ func main() {
...
-		"video_count",
-		"number of videos processed over time",
-		nil,
-		videoCount,
-		view.CountAggregation{},
-	)
-	if err != nil {
-		log.Fatalf("Cannot create view: %v", err)
-	}
+	videoCount, _ := stats.Int64("my.org/measures/video_count", "number of processed videos", "")
```
odeke-em: As I had mentioned offline, examples are our face of style and in my experience I've found
lots of users copied and pasted the examples so let's do some error handling here to
avoid future bug reports in which users can tell where the nil dereference came from
yet handling the error could have helped them.

stats/view/worker.go
```diff
@@ -104,25 +84,46 @@ func (v *View) Subscribe() error {
...
+		v:   v.Name,
 		err: make(chan error),
 	}
 	defaultWorker.c <- req
 	return <-req.err
 }
 
 // RetrieveData returns the current collected data for the view.
+// Deprecated: Use RetrieveData(name)
 func (v *View) RetrieveData() ([]*Row, error) {
```
rakyll: I think we should remove this API. No one is using RetrieveData.

stats/view/view.go
odeke-em: I defer to @rakyll's suggestion.

stats/view/worker.go
ramonza: What I'm going to do is convert Unsubscribe to a function as well and deprecate the existing method.

stats/view/view.go
ramonza: In java they are called "Columns" (in the View class) FTR I am: +0 on TagKeys (just repeats the type) +1 on GroupBy (I like SQL, many people know SQL) and I don't really understand where Columns comes from? I am happy to disagree and commit though so @dinooliva @bogdandrutu what's the suggestion here?

README.md
ramonza: I don't know what you mean here. Do you want views to be "automatically" subscribed? Or do you want to keep the method as we previously had.

This situation (a single view argument) is not the strong suite of this current proposed approach. It works much better when there are multiple views (see ochttp/example_test.go)

plugin/ochttp/example_test.go
ramonza: To be honest, I think it will not be an edge-case feature. Our view names for HTTP and gRPC are so long I would almost certainly rename them if I were an using OC in an application. I feel like we haven't got this part right quite yet. I'll remove it but I think we need to get some practical experience with what such long view names look like in the UI of common metrics backends.

examples/stats/prometheus/main.go
ramonza: ack

README.md
ramonza: Take a look at 8674a4473ec6bfe28077dc45ab4843daa053feb4

README.md
```diff
@@ -119,23 +120,29 @@ if err != nil {
 }
 ```
 
-Retrieve measure by name:
+### Recording Measurements
 
-[embedmd]:# (stats.go findMeasure)
+Measurements are data points associated with Measures.
+Recording implicitly tags the set of Measurements with the tags from the
```
odeke-em: Hmm, given this mention here, perhaps it would help if in the example we inserted a tag into a context right before recording.

examples/stats/prometheus/main.go
```diff
@@ -93,8 +76,7 @@ func main() {
 	// Record some data points...
 	go func() {
 		for {
-			stats.Record(ctx, videoCount.M(1))
-			stats.Record(ctx, videoSize.M(rand.Int63()))
+			stats.Record(ctx, videoCount.M(1), videoSize.M(rand.Int63()))
```
odeke-em: This is great, thank you!

examples/stats/prometheus/main.go
```diff
@@ -43,47 +43,30 @@ func main() {
...
-	}
-
-	// 2. Create view to see the amount of video processed
-	viewSize, err := view.New(
-		"video_cum",
-		"processed video size over time",
-		nil,
-		videoSize,
-		view.DistributionAggregation([]float64{0, 1 << 16, 1 << 32}),
+	err = view.Subscribe(
```
odeke-em: This new subscribe feels awesome for user experience!

exporter/stackdriver/stats_test.go
```diff
@@ -83,15 +83,8 @@ func TestExporter_makeReq(t *testing.T) {
...
-	v, err := view.New("testview", "desc", []tag.Key{key}, m, view.CountAggregation{})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	distView, err := view.New("distview", "desc", nil, m, view.DistributionAggregation([]float64{2, 4, 7}))
-	if err != nil {
-		t.Fatal(err)
-	}
+	v, _ := view.New("testview", "desc", []tag.Key{key}, m, view.CountAggregation{})
```
odeke-em: Why not just `&view.View{...}`?

exporter/stackdriver/stats_test.go
```diff
@@ -573,10 +565,7 @@ func TestExporter_createMeasure(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	v, err := view.New("testview", "desc", []tag.Key{key}, m, view.CountAggregation{})
-	if err != nil {
-		t.Fatal(err)
-	}
+	v, _ := view.New("testview", "desc", []tag.Key{key}, m, view.CountAggregation{})
```
odeke-em: Ditto, eventually we'll phase view.New out ;)

internal/readme/stats.go
```diff
@@ -57,36 +57,17 @@ func statsExamples() {
...
+		Name:        "my.org/video_size_distribution",
+		Description: "distribution of processed video size over time",
+		Measure:     videoSize,
+		Aggregation: view.DistributionAggregation([]float64{0, 1 << 32, 2 << 32, 3 << 32}),
+	})
 	if err != nil {
-		log.Fatalf("cannot create view: %v", err)
-	}
-	if err := view.Register(v); err != nil {
 		log.Fatal(err)
```
odeke-em:
```go
log.Fatalf("Cannot create and subscribe view: %v", err)
```

README.md
ramonza: There's a separate section on tags. This statement just says why we take a context parameter to Record.

ramonza:
@odeke-em issues addressed, PTAL

README.md
odeke-em: Roger that, thanks for the clarification.

odeke-em:
Also please go-fmt your files.

PR #496, opened 2018-02-27
Record on a separate buffer

rakyll:
This change separates the recording of measurement events from
all other events.
We prefer to use a slice and a mutex because this combination is
more performant than a buffered channel.

Even though there is still significant time spent in runtime.growslice,
we don't have a better growing strategy at the moment.

```
name                old time/op    new time/op    delta
Record0-8             4.08ns ± 0%    3.37ns ± 0%   
Record1-8              359ns ± 0%     244ns ± 0%   
Record8-8              460ns ± 0%     443ns ± 0%   
Record8_Parallel-8     855ns ± 0%     553ns ± 0%  
Record8_8Tags-8        450ns ± 0%     447ns ± 0% 
```

### DIFF
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -18,6 +18,7 @@ package view
 import (
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"go.opencensus.io/stats"
@@ -44,6 +45,9 @@ type worker struct {
 	timer      *time.Ticker
 	c          chan command
 	quit, done chan bool
+
+	recordsMu sync.Mutex
+	records   []*recordReq
 }
 
 var defaultWorker *worker
@@ -131,11 +135,14 @@ func (v *View) RetrieveData() ([]*Row, error) {
 }
 
 func record(tags *tag.Map, ms interface{}) {
-	req := &recordReq{
+	w := defaultWorker
+
+	w.recordsMu.Lock()
+	w.records = append(w.records, &recordReq{
 		tm: tags,
 		ms: ms.([]stats.Measurement),
-	}
-	defaultWorker.c <- req
+	})
+	w.recordsMu.Unlock()
 }
 
 // SetReportingPeriod sets the interval between reporting aggregated views in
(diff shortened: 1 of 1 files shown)


rakyll:
We eventually will move to a model where stats.Record contention is minified and the new ticker is collecting from multiple buffers.

stats/view/worker.go
```diff
@@ -131,11 +135,14 @@ func (v *View) RetrieveData() ([]*Row, error) {
 }
 
 func record(tags *tag.Map, ms interface{}) {
-	req := &recordReq{
+	w := defaultWorker
+
+	w.recordsMu.Lock()
+	w.records = append(w.records, &recordReq{
```
ramonza: It worries me that this slice can grow unbounded.

stats/view/worker.go
```diff
@@ -171,6 +179,13 @@ func (w *worker) start() {
 			if cmd != nil {
 				cmd.handleCommand(w)
 			}
+		case <-time.Tick(10 * time.Millisecond):
+			w.recordsMu.Lock()
```
ramonza: This mutex doesn't protect the other parts of the worker (like the previous channel approach did). I think that means you'll have a race when recording and e.g. adding views concurrently.

ramonza:
It seems odd that this is faster than a buffered channel. I worry that the benchmark is better because work is being done in much larger batches that wouldn't be as big outside of a tight loop. It would be interesting to see how large the batches become in the benchmark.

rakyll:
We chatted with @ramonza  and decided to not to proceed with this improvement without understanding the real life scenarios. Lots of decisions require us the to tweak the constants we have selected in this PR such as the initial buffer size, buffer growing strategy and how often we will consume the buffer.

stats/view/worker.go
rakyll: Since it is in the same select, everything else will wait until records buffer is consumed. So recordReq.handleCommand will be able to access worker fields without holding a mutex.

PR #499, opened 2018-02-28
Fix the status code attribute key

rakyll:
We recently switch to "status_code" from "status" as
it is documented on the spec at
https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTPAttributes.md.

### DIFF
--- a/plugin/ochttp/trace.go
+++ b/plugin/ochttp/trace.go
@@ -33,7 +33,7 @@ const (
 	MethodAttribute     = "http.method"
 	PathAttribute       = "http.path"
 	UserAgentAttribute  = "http.user_agent"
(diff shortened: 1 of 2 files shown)


plugin/ochttp/trace.go
```diff
@@ -33,7 +33,7 @@ const (
 	MethodAttribute     = "http.method"
 	PathAttribute       = "http.path"
 	UserAgentAttribute  = "http.user_agent"
-	StatusCodeAttribute = "http.status"
+	StatusCodeAttribute = "http.status_code"
```
odeke-em: Aha nice, great to see that this is finally changed after PR https://github.com/census-instrumentation/opencensus-go/pull/416

PR #500, opened 2018-02-28
Deprecate zpages.TracezHandler and zpages.RpczHandler

### LINKED ISSUE #477: No easy way to register zpages.Handler on the default mux
It is not clear how to register the zpages.Handler on the default mux.

zpages.Handler is a mux named as a handler. Users typically assume that they can register it on the default mux on a custom path such as:

```
http.Handle("/debug", zpages.Handler)
```

But /rpcz won't be accessible either at http://localhost:port/debug/rpcz or http://localhost:port/rpcz.

I think we should bring back the AddDefaultHTTPHandlers and also export the static file handler.

/cc @ramonza

rakyll:
Also modify the godoc examples to provide a reference
to the newly recommended API.

Fixes #477.


rakyll:
I will remove the old APIs instead given zpages are not a commonly-used feature at this point.

Issue #497, opened 2018-02-28
Make Stackdriver exporter default to using DefaultCredentials


ramonza:
```go
if options.ProjectID == "" {
			creds, err := google.FindDefaultCredentials(context.Background(), "")
			if err != nil {
				return err
			}
			options.ProjectID = creds.ProjectID
		}
```

PR #501, opened 2018-02-28
Application default credentials for Stackdriver

ramonza:
Also added a convenience `Register` method to register as both a
stats and trace Exporter.

Fixes: #497


exporter/stackdriver/stackdriver.go
```diff
@@ -73,9 +80,34 @@ type Exporter struct {
 	statsExporter *statsExporter
 }
 
+// Register sets up both stats and tracing exporting.
+func Register(o Options) (done func(), err error) {
```
odeke-em: Seems more like a convenience method that a user should make on their own than something we want to expose in our API, could we perhaps make it a helper for the test as you had done?

ramonza: My goal here was to make it as simple as possible for people to "turn on" OpenCensus for GCP client libraries. Adding this method makes this a one-liner as opposed to a 6-liner. @rakyll what do you think?

rakyll: I agree that done might be giving the wrong impression that it should be deferred right away. What about calling it unregister and document that it is more likely never be called but can be called to unregister and flush?

exporter/stackdriver/stackdriver.go
```diff
@@ -73,9 +80,34 @@ type Exporter struct {
...
+		trace.UnregisterExporter(e)
+		e.Flush()
+	}, nil
+}
+
 // NewExporter creates a new Exporter that implements both stats.Exporter and
 // trace.Exporter.
 func NewExporter(o Options) (*Exporter, error) {
+	if o.ProjectID == "" {
+		creds, err := google.FindDefaultCredentials(context.Background(), traceapi.DefaultAuthScopes()...)
```
rakyll: Do we need to pass the auth scopes here? We are not using creds to create a new RoundTripper. We can omit it.

ramonza: true, but does it not make sense to defensively pass them anyway? it seems conceivable that someone might add a way to override default credentials by scope, in which case we want to pick up the ones associated with Stackdriver.

ramonza:
@odeke-em removed the Register helper function, PTAL

Issue #502, opened 2018-02-28
OpenTracing implementation on top of OpenCensus trace


ramonza:
Many frameworks have already been instrumented with OpenTracing. Building an OpenTracing implementation on top of OpenCensus would give us compatibility with these frameworks.

savaki:
I've written a few open tracing implementations so I'd be happy to do this.  There are a couple of incompatibilities that I'd love some guidance on:

* OT has a concept of baggage, strings key values that are set on a span and passed to all child spans.  I'm assuming we would just document this is not supported.
* OT has logging integrated.  My suggestion here is that the OC OT implementation allow for an optional logger interface to be provided.
* OT tags are string => interface{} which is much broader than an attribute.  Presumably this would be another one to document.

The primary behavior that will be odd for people using the OC OT implementation imho is that log messages which previously contained baggage would now just contain the log message (plus attributes).  I know we put things like tenantID and userID into the baggage so losing these would be a showstopper.

savaki:
If I put my marketing hat on.  If you really want OC to be widely adopted, perhaps a parallel approach would be to write native OC implementations for those other frameworks.  Otherwise, there's really no reason to use OC directly if you can always use the OT adapter.

ramonza:
We are worried that having two trace recording APIs in this repo will cause confusion for users. We are looking for an appropriate place to put this work.

savaki:
Fair enough.  Let me know what you'd like me to do.

ramonza:
@savaki we are going to create an "opencensus-contrib" organization to put all the integrations (including exporters) and I think this can go there. You can keep the PR open for now if you like or publish it to a repo in your account that we can then clone to the new org once it has been created. It should be created this week.

someone1:
Would love to see this (the OpenTracing vs OpenCensus debate is making my head spin) - is there a connector somewhere for this today? I can't seem to find anything anywhere, did the previous work just disappear?

smartheye:
hi, happy to see this issue. is there possible that opencensus provide a opentraing api. a bridge. I am using opencensus, you guys are doing great!

sagikazarmark:
I'm looking for a solution as well. Although it would be great to see OC adoption instead of using bridges, when using third party stuff (eg. Traefik), we don't really have a choice.

rghetia:
This will be addressed in upcoming OC - OT merging. keeping the issue open untill we haver merged api available.

PR #505, opened 2018-02-28
initial take at opencensus opentracing implementation

savaki:
Implementation allows users to provide external Logger implementation.  Tags and baggage will be published to logger, but will not interact with opencensus trace.

### DIFF
--- a/contrib/opentracer/logger.go
+++ b/contrib/opentracer/logger.go
@@ -1,0 +1,43 @@
+package opentracer
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"time"
+
+	"github.com/opentracing/opentracing-go/log"
+)
+
+type Logger interface {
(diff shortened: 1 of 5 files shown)


ramonza:
Yeah, I hadn't considered the API question very deeply when I filed #502. 

I think having this bridge would be super useful to have as many libraries are already instrumented with OpenTracing. 

Whether it lives here or in a the opentracing-contrib repo I don't really have a strong opinion. I guess we are already getting flak for the number of dependencies this repo is pulling in so that's another argument against adding contrib here.

rakyll:
I am suggesting to merge this to https://github.com/opentracing-contrib given we don't have a contrib organization at the moment.

ramonza:
@savaki if you have some time, would you mind moving this to a separate repo that we can then pull into the newly created census-ecosystem org?

Issue #508, opened 2018-03-01
trace: calculate all children span start times using the monotonic clock


rakyll:
We initially made an improvement to calculate the span duration from the monotonic clock.

We also have to make sure that children start times are calculated with the monotonic clock to have a more precise overview of a trace.

odeke-em:
After https://github.com/census-instrumentation/opencensus-go/pull/509 is merged, then we can easily work on this too.

PR #509, opened 2018-03-01
Add an internal utility MonotonicTimeSince

### LINKED ISSUE #506: trace: make clear that the computation in setting Span.EndTime in Span.End() is using the monotonic clock
While editing a bunch of code I noticed this
https://github.com/census-instrumentation/opencensus-go/blob/6e008f8971f53840fe7b528ab3d733a38e54a4bb/trace/trace.go#L231

That code is an unnecessary computation, because:
a) time.Since(startTime) = time.Now().Sub(startTime)
b) startTime.Add(time.Since(startTime)) = startTime.Add(time.Now().Sub(startTime)) == time.Now()

that line should just be time.Now()

I'd send a PR but am deep in edits on my branch.

odeke-em:
Fixes #506

We added a monotonic time calculation but without
much context as to why, so by just looking at the code
the calculation seemed unnecessary but after a few
discussions and godoc readings, calculation of time
using the monotonic clock is achievable by subtractions
hence the startTime.Add(time.Now(startTime).
Extract that code into an internal helper for clarity
and since we'll be re-using that code later on, such as
for #508

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -14,6 +14,16 @@
 
 package internal
 
+import "time"
+
 // UserAgent is the user agent to be added to the outgoing
 // requests from the exporters.
 const UserAgent = "opencensus-go-v0.1.0"
+
+func MonoticTimeSince(start time.Time) time.Time {
+	// The monotonic clock is used in subtractions hence
+	// the duration since start added back to start gives
+	// end as a monotonic time.
+	// See https://golang.org/pkg/time/#hdr-Monotonic_Clocks
+	return start.Add(time.Now().Sub(start))
+}
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -20,10 +20,10 @@ import "time"
 // requests from the exporters.
 const UserAgent = "opencensus-go-v0.1.0"
 
+// The monotonic clock is used in subtractions hence
+// the duration since start added back to start gives
+// end as a monotonic time.
+// See https://golang.org/pkg/time/#hdr-Monotonic_Clocks
 func MonoticTimeSince(start time.Time) time.Time {
-	// The monotonic clock is used in subtractions hence
-	// the duration since start added back to start gives
-	// end as a monotonic time.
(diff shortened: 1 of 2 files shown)


internal/internal.go
```diff
@@ -14,6 +14,16 @@
...
 package internal
 
+import "time"
+
 // UserAgent is the user agent to be added to the outgoing
 // requests from the exporters.
 const UserAgent = "opencensus-go-v0.1.0"
+
+func MonoticTimeSince(start time.Time) time.Time {
+	// The monotonic clock is used in subtractions hence
```
rakyll: Want to move these comments to the function godoc?

odeke-em: Roger that, great idea.

odeke-em:
All addressed, thank you @rakyll, PTAL.

internal/internal.go
```diff
@@ -14,6 +14,16 @@
...
+
 // UserAgent is the user agent to be added to the outgoing
 // requests from the exporters.
 const UserAgent = "opencensus-go-v0.1.0"
+
+// The monotonic clock is used in subtractions hence
+// the duration since start added back to start gives
+// end as a monotonic time.
+// See https://golang.org/pkg/time/#hdr-Monotonic_Clocks
+func MonoticTimeSince(start time.Time) time.Time {
```
ramonza: name here seems a little strange. "since" seems to imply that it returns a duration. What about "MonotonicEndTime"?

odeke-em: In deed, thanks for catching that. However, the new name deosn't convey how we calculate the EndTime as being from time.Now, how about an update to your suggestion `MonotonicTimeNow`?

odeke-em: Actually, never mind, I'll go with your suggestion, I withdraw my previous reservations.

odeke-em:
@ramonza I've updated the PR with your suggestions, PTAL!

ramonza:
LGTM

odeke-em:
Thank you!

PR #513, opened 2018-03-02
ensures span names conform to aws xray spec

savaki:


exporter/xray/segment_test.go
```diff
@@ -137,3 +137,79 @@ func TestParseAmazonSpanID(t *testing.T) {
...
+			if actual := fixSegmentName(tc.Name); tc.Expected != actual {
+				t.Errorf("expected %v; got %v", tc.Expected, actual)
+			}
+		})
+	}
+}
+
+// makeStringN returns a string of the specified length
+func makeStringN(length int) string {
+	var content []byte
```
odeke-em: This could be simplified to a single and more efficient call:
```go
return strings.Repeat(" ", length)
```

PR #517, opened 2018-03-03
created datadog stats exporter

savaki:
publishes stats via dogstatsd agent

### DIFF
--- a/exporter/datadog/datadog.go
+++ b/exporter/datadog/datadog.go
@@ -1,0 +1,313 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package datadog contains an OpenCensus stats exporter for Datadog.
+package datadog // import "go.opencensus.io/exporter/datadog"
+
+import (
+	"io"
+	"log"
+	"os"
+	"reflect"
+	"regexp"
+
+	"github.com/DataDog/datadog-go/statsd"
+	"go.opencensus.io/stats/view"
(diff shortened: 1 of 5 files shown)


exporter/datadog/datadog.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+type optionFunc func(c *config)
+
+func (fn optionFunc) apply(c *config) {
+	fn(c)
+}
+
+// WithOutput - optional writer for error messages
+func WithOutput(w io.Writer) Option {
```
ramonza: why not accept a log.Logger?

exporter/datadog/datadog.go
```diff
@@ -0,0 +1,313 @@
...
+		var count int64
+		if index < length {
+			count = value.CountPerBucket[index]
+		}
+		if index < lastLength {
+			count = count - lastValue.CountPerBucket[index]
+		}
+
+		for i := int64(0); i < count; i++ {
+			e.client.Histogram(name, bucket, tags, rate)
```
ramonza: Seems odd that we make the same call repeatedly

exporter/datadog/datadog_test.go
```diff
@@ -0,0 +1,413 @@
...
+		count        = makeView(key, m, "opencensus.sample.count", view.CountAggregation{})
+		mean         = makeView(key, m, "opencensus.sample.mean", view.MeanAggregation{})
+		distribution = makeView(key, m, "opencensus.sample.distribution", view.DistributionAggregation{1, 2, 3, 4, 5})
+	)
+
+	sum.Subscribe()
+	defer sum.Unsubscribe()
+
+	count.Subscribe()
+	defer count.Unsubscribe()
```
ramonza: in general I would not worry too much about unsubscribing from views in tests. as long as you ensure that your view names and measure names are unique to the test, it should not be necessary and doesn't really add anything. we don't expect that in practice people will use Unsubscribe much.

exporter/datadog/datadog.go
savaki: Yep, I thought so too.  statsd and Views are conceptually very similar.  They take raw events and roll them up into something interesting.  What I was looking for was effectively a way to undo what View had done.  

In an ideal world, I could supply a statsd view and allow statsd to do this.  As it is, I'm try to reconstruct the originating measures from the view.

ramonza:
Created a spec issue to discuss how best to support this type of local aggregation agent: https://github.com/census-instrumentation/opencensus-specs/issues/63

rakyll:
What's the status of this PR?

PR #522, opened 2018-03-07
Don't force download all dependencies

rakyll:
With https://github.com/census-instrumentation/opencensus-go/pull/85,
it became possible to go get the package without `...`.

Without the `...`, the dependencies is not going to be downloaded.

### DIFF
--- a/internal/readme/source.md
+++ b/internal/readme/source.md
@@ -16,7 +16,7 @@ rapidly, vendoring is recommended.
 ## Installation
 
 ```
-$ go get -u go.opencensus.io/...
+$ go get -u go.opencensus.io
 ```
 
 ## Prerequisites


internal/readme/source.md
```diff
@@ -16,7 +16,7 @@ rapidly, vendoring is recommended.
 ## Installation
 
 ```
-$ go get -u go.opencensus.io/...
```
ramonza: What is the point of this? There is nothing in the root. I don't think we need a generic installation section but maybe above each example we can install the package that it uses.

rakyll: There is a Go file in the root. It will clone the repo on your GOPATH.

rakyll: I mean, there is no concept of installing a library in Go other than cloning things to your GOPATH and documenting the import path.

We have a vanity URL. This instruction only tells the user what import path we use and is a good practice to do in READMEs.

PR #526, opened 2018-03-07
Add LastAggregation and LastData

rakyll:
This aggregation keeps the last value recorded.

LastValue is explained at
https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/DataAggregation.md.


rakyll:
We are working for another API for view.Data, we shouldn't work on this PR until that is not resolved.

ramonza:
I think this is a reasonable approach and distinct from the "no aggregation" case since it provides us a way to limit memory usage (keeping only the most recent value). I am +1 on merging this. 

/cc @bogdandrutu @groob

bogdandrutu:
Based on the discussion I had today with @rakyll I think we decided to support this for aggregated data with custom (runtime) tags.

The other interface that we will provide will be for things with static tags like gc-stats or cpu-usage.

Any thoughts on this?

ramonza:
@bogdandrutu I don't understand what you mean by custom (runtime) tags

rakyll:
We will recommend using LastValue for measurements that need to be broken by tags. That's what he meant by custom (runtime) tags. If you don't need to break down by tags, users should use the new interface.

rakyll:
PTAL for the APIs. Once we agree, I will work on the exporter support.

PR #528, opened 2018-03-07
Hide Measurement fields.

ramonza:
We don't want to expose Measurement fields or allow direct construction
of Measurement instances because we want to keep open the possibility
of adding new Measurement representations and also encourage users to
use the .M() method.

Also saw that we make some performance promises about Measure that are
not really true. Looks like there was the beginnings of an implementation
of the fast path for skipping recording but it was never fully hooked
up.

### DIFF
--- a/stats/measure_test.go
+++ b/stats/measure_test.go
@@ -26,6 +26,11 @@ func TestCheckMeasureName(t *testing.T) {
 			view:    "my.org/measures/\007",
 			wantErr: true,
 		},
+		{
+			name:    "no emoji for you!",
+			view:    "💩",
+			wantErr: true,
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/stats/record.go
+++ b/stats/record.go
@@ -24,20 +24,8 @@ import (
 
 // Record records one or multiple measurements with the same tags at once.
 // If there are any tags in the context, measurements will be tagged with them.
-//
-// Record returns fast and recorded measurements will have little overhead
-// if there are no views assicaiated with them.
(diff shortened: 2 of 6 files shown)


rakyll:
Can you fix the title saying Measurement?

stats/measure_test.go
```diff
@@ -26,6 +26,11 @@ func TestCheckMeasureName(t *testing.T) {
 			view:    "my.org/measures/\007",
 			wantErr: true,
 		},
+		{
+			name:    "perfectly reasonable name",
+			view:    "💩",
+			wantErr: true, // 😞
```
rakyll: Can you remove the comment here? It is confusing.

stats/record.go
```diff
@@ -24,20 +24,8 @@ import (
 
 // Record records one or multiple measurements with the same tags at once.
 // If there are any tags in the context, measurements will be tagged with them.
-//
-// Record returns fast and recorded measurements will have little overhead
-// if there are no views assicaiated with them.
 func Record(ctx context.Context, ms ...Measurement) {
 	if internal.DefaultRecorder != nil {
-		var record bool
```
odeke-em: Why are we removing this code? It doesn't hurt to skip recording if there is no Measure attached,
plus this seems to revert https://github.com/census-instrumentation/opencensus-go/pull/468 whose purpose was to improve performance

odeke-em: and in particular https://github.com/census-instrumentation/opencensus-go/pull/468/files#diff-8f16a4ad787d3d7d83413c0f7415e357

ramonza: Nothing ever set measure.views so this code path was never taken

PR #533, opened 2018-03-07
prometheus: make registry configurable

tecbot:
This allows users to use their own prometheus registry for exporting metrics.


rakyll:
Thanks much!

PR #534, opened 2018-03-07
Add stats support for net/http servers

rakyll:


plugin/ochttp/stats.go
```diff
@@ -25,20 +25,31 @@ var (
 	ClientRequestCount, _  = stats.Int64("opencensus.io/http/client/request_count", "Number of HTTP requests started", stats.UnitNone)
 	ClientRequestBytes, _  = stats.Int64("opencensus.io/http/client/request_bytes", "HTTP request body size if set as ContentLength (uncompressed)", stats.UnitBytes)
 	ClientResponseBytes, _ = stats.Int64("opencensus.io/http/client/response_bytes", "HTTP response body size (uncompressed)", stats.UnitBytes)
-	ClientLatency, _       = stats.Float64("opencensus.io/http/client/latency", "End-to-end latency", stats.UnitMilliseconds)
+	ClientLatency, _       = stats.Int64("opencensus.io/http/client/latency", "End-to-end latency", stats.UnitMilliseconds)
```
ramonza: Looks like this was accidentally changed from Float64 to Int64

plugin/ochttp/stats_test.go
```diff
@@ -1,27 +1,116 @@
...
+		var count int
+		var sum float64
+		switch data := data.(type) {
+		case *view.CountData:
+			count = int(*data)
+		case *view.DistributionData:
+			count = int(data.Count)
+			sum = data.Sum()
+		default:
+			t.Errorf("don't know how to handle data type: %v", data)
```
odeke-em: continue

tag/key.go
```diff
@@ -21,7 +21,7 @@ type Key struct {
 }
 
 // NewKey creates or retrieves a string key identified by name.
-// Calling NewKey consequently with the same name returns the same key.
+// Calling NewKey (or MustKey) with the same name returns the same key.
```
odeke-em: Looks like a stray comment, the `MustKey` or do we already have `MustKey`? And I believe we were talking of not exposing MustKey, right?

plugin/ochttp/stats_transport.go
```diff
@@ -126,3 +126,73 @@ func (t *tracker) Close() error {
 	t.end()
 	return t.body.Close()
 }
+
+func (h *Handler) startStats(w http.ResponseWriter, r *http.Request) (http.ResponseWriter, *http.Request, func()) {
+	ctx, _ := tag.New(r.Context(),
```
ramonza: I don't think we should add these tags by default until we have a way to control propagation. It means that every outgoing gRPC call will have a HTTP host tag added to its metadata that will be propagated to the gRPC server and set there as well. This doesn't seem like an expected behavior. Instead, we should add the tags just around the Record call, not the context used on the Request.

rakyll:
Done, PTAL

plugin/ochttp/stats_transport.go
ramonza: it's also a little strange that his method is in the _transport.go file

rakyll:
I will send a cleanup PR for this package.

PR #536, opened 2018-03-08
Cleanup the ochttp package

rakyll:
Several reorganization related and readability improvements.


plugin/ochttp/server.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Handler is the handler used to handle the incoming request.
+	Handler http.Handler
+}
+
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if !h.NoTrace {
+		var end func()
+		r, end = h.startTrace(w, r)
+		defer end()
```
odeke-em: Aha, I like this, I read the previous code on my way back home and had the urge for this kind of change, so thank you!

plugin/ochttp/server.go
```diff
@@ -0,0 +1,160 @@
...
+	ctx := r.Context()
+	var span *trace.Span
+	if sc, ok := p.SpanContextFromRequest(r); ok {
+		ctx, span = trace.StartSpanWithRemoteParent(ctx, name, sc, trace.StartOptions{})
+	} else {
+		ctx, span = trace.StartSpan(ctx, name)
+	}
+
+	span.SetAttributes(requestAttrs(r)...)
+	return r.WithContext(trace.WithSpan(r.Context(), span)), span.End
```
odeke-em: Nice, thanks for changing it from `func() { span.End }` to `span.End`!

plugin/ochttp/server.go
```diff
@@ -0,0 +1,160 @@
...
+	ctx        context.Context
+	reqSize    int64
+	respSize   int64
+	start      time.Time
+	statusCode int
+	endOnce    sync.Once
+	writer     http.ResponseWriter
+}
+
+var _ http.ResponseWriter = (*trackingResponseWriter)(nil)
```
odeke-em: Cool, thanks for this cleanup.

plugin/ochttp/server.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+type trackingResponseWriter struct {
+	ctx        context.Context
+	reqSize    int64
+	respSize   int64
+	start      time.Time
+	statusCode int
+	endOnce    sync.Once
+	writer     http.ResponseWriter
```
odeke-em: writer seems quite ambigious for the field of http.ResponseWriter.

How about either of?
* rw
* responseWriter

PR #545, opened 2018-03-09
Add benchmark to internal stats worker record command

ramonza:
The current stats benchmarks don't examine the recording work that
happens in the worker goroutine directly. Added this benchmark to
attempt to capture that.

```
goos: darwin
goarch: amd64
pkg: go.opencensus.io/stats
BenchmarkRecord0-8            	 5000000	       226 ns/op	      32 B/op	       1 allocs/op
BenchmarkRecord1-8            	 5000000	       304 ns/op	      96 B/op	       3 allocs/op
BenchmarkRecord8-8            	 5000000	       382 ns/op	     256 B/op	       3 allocs/op
BenchmarkRecord8_Parallel-8   	 2000000	       659 ns/op	     256 B/op	       3 allocs/op
BenchmarkRecord8_8Tags-8      	 3000000	       390 ns/op	     256 B/op	       3 allocs/op
PASS
goos: darwin
goarch: amd64
pkg: go.opencensus.io/stats/view
BenchmarkRecordReqCommand-8   	 1000000	      1839 ns/op	     362 B/op	      25 allocs/op
```

### DIFF
--- a/stats/view/benchmark_test.go
+++ b/stats/view/benchmark_test.go
@@ -1,0 +1,90 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package view
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
+)
+
+var (
+	m, _  = stats.Float64("m", "", "")
+	k1, _ = tag.NewKey("k1")
+	k2, _ = tag.NewKey("k2")
+	k3, _ = tag.NewKey("k3")
+	k4, _ = tag.NewKey("k4")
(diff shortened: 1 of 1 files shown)


stats/view/benchmark_test.go
```diff
@@ -0,0 +1,95 @@
...
+		Measure:     m,
+		Aggregation: DistributionAggregation{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
+		TagKeys:     []tag.Key{k1, k2},
+	}
+)
+
+// BenchmarkRecordReqCommand benchmarks calling the internal recording machinery
+// directly.
+func BenchmarkRecordReqCommand(b *testing.B) {
+	var (
```
rakyll: Remove var grouping.

stats/view/benchmark_test.go
```diff
@@ -0,0 +1,95 @@
...
+		w         = newWorker()
+		errchan   = make(chan error, 1)
```
rakyll: errChan

stats/view/benchmark_test.go
```diff
@@ -0,0 +1,95 @@
...
+		subscribe = &subscribeToViewReq{views: []*View{view}, err: errchan}
+	)
+	subscribe.handleCommand(w)
+	err := <-errchan
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	var (
```
rakyll: Remove var grouping.

stats/view/benchmark_test.go
```diff
@@ -0,0 +1,95 @@
...
+		ctx  = context.Background()
+		ctxs = make([]context.Context, 0, 10)
```
rakyll: Either convert size into a const.

```
const size = 10
```

Or initiate this slice without a capacity. We don't care about the performance of this section of the test.

It makes it harder to refactor to hardcode size in multiple places.

Issue #544, opened 2018-03-09
Creating measures and checking errors is inconvenient


ramonza:
The issue of having Must* variants to create Measures has come up before. Previously, the argument against having such variants was that the caller in most cases knows that the measure name is valid and so can simply ignore the error.

This is true for tag.Key where the only validation is that the key name is valid syntactically. With stats.Int64 and stats.Float64, you also need to know that the name is *unique*. This is not something that the caller can know just from the text of the program. I have actually run into this multiple times, especially in tests where you create measures inline. 

I think we should either add a Must variant for stats.Int64 and stats.Float64 or make duplicating the measure name not an error.

/cc @rakyll

rakyll:
I agree that there is a compelling reason to add this for measures but the other question how effective the uniqueness check is. Although it gives us integrity within a single process, it doesn't help us with uniqueness beyond a single process. IMHO, we can prefer to stop retuning an error if from stats.Int64 and stats.Float64 if returning errors doesn't add significant value.

rakyll:
We chatted with @bogdandrutu offline and decided that we should keep returning errors. We might implement a global uniqueness check later on and want to be facilitate the current API.

Bogdan also agrees in-process uniqueness check doesn't add much value and we can remove it.

ramonza:
A global uniqueness check happening implicitly in package init (which is where these functions are mostly called) sounds unlikely to be implementable. 

The way I can imagine this playing out (if we really wanted to check global consistency of measures) is that there would be some initialization function which would do a single remote call to validate all the defined methods that the user could call from their main. This function would be able to return errors. 

Therefore, I think there's no reason to return an error in this function and I would prefer that we remove it from the function signature. 

@rakyll @bogdandrutu

PR #552, opened 2018-03-09
Make stats.{Int,Float}64 never return errors

ramonza:
I left the error in the signature for now since it's used externally. We should remove it in the next round of deprecation removals.

Updates: #544


stats/measure_float64.go
```diff
@@ -37,21 +29,15 @@ func (f *Float64Measure) M(v float64) Measurement {
 	return Measurement{m: f, v: v}
 }
 
-// Float64 creates a new measure of type Float64Measure. It returns
-// an error if a measure with the same name already exists.
+// Float64 creates a new measure of type Float64Measure.
+// It never returns an error.
+//TODO(ramonza): remove error
```
rakyll: Move this in the block.

stats/measure_int64.go
```diff
@@ -17,15 +17,7 @@ package stats
...
-	measure
-}
-
-func (i *Int64Measure) subscribe() {
-	i.measure.subscribe()
-}
-
-func (i *Int64Measure) subscribed() bool {
-	return i.measure.subscribed()
+	Measure
```
rakyll: Why are we changing this?

stats/measure_int64.go
```diff
@@ -37,21 +29,15 @@ func (i *Int64Measure) M(v int64) Measurement {
 	return Measurement{m: i, v: float64(v)}
 }
 
-// Int64 creates a new measure of type Int64Measure. It returns an
-// error if a measure with the same name already exists.
+// Int64 creates a new measure of type Int64Measure.
+// It never returns an error.
+//TODO(ramonza): remove error
```
rakyll: Remove this in the block.

ramonza:
@rakyll removed the embedding, PTAL

Issue #549, opened 2018-03-09
Rename plugin/ocgrpc/testdata


zombiezen:
As per https://golang.org/cmd/go/#hdr-Test_packages, the go tool will ignore directories called "testdata" when evaluating packages to build.  However, this actually holds a build-able go package that trace_common_test.go depends on.  This causes various issues in tools that try to do import graph analysis, so please consider renaming the package (perhaps `internal/testpb`?).

/cc @rakyll

PR #554, opened 2018-03-09
Rename plugin/ocgrpc/testdata to plugin/internal/testpb

ramonza:
Fixes: #549


plugin/ocgrpc/trace_common_test.go
```diff
@@ -21,7 +21,7 @@ import (
 	"testing"
 	"time"
 
-	testpb "go.opencensus.io/plugin/ocgrpc/testdata"
+	testpb "go.opencensus.io/plugin/ocgrpc/internal/testpb"
```
rakyll: You can rename the proto package name to testpb.

PR #558, opened 2018-03-12
Allow customizing ocgrpc.ServerHandler

ramonza:
* Added ServerHandler.AlwaysStartNewTraces to allow a configuration
  suitable for a public-facing server.
* Added ServerHandler.StartOptions to allow customizing the sampler
  used for new spans.

### DIFF
--- a/plugin/ocgrpc/server.go
+++ b/plugin/ocgrpc/server.go
@@ -15,6 +15,7 @@
 package ocgrpc
 
 import (
+	"go.opencensus.io/trace"
 	"golang.org/x/net/context"
 
 	"google.golang.org/grpc/stats"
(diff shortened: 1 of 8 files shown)


plugin/ocgrpc/server.go
```diff
@@ -28,34 +29,65 @@ func NewServerStatsHandler() stats.Handler {
...
+
+	// AlwaysStartNewTrace may be set to true to start a new trace, wrapping each
+	// RPC with a root Span.
+	//
+	// You should set this if your gRPC server is a public-facing service.
+	//
+	// Be aware that if you leave this false (the default) on a public-facing
+	// server, callers will be able to send tracing metadata in gRPC headers
+	// and trigger traces in your backend.
+	AlwaysStartNewTrace bool
```
ramonza: We are going to rename this option

Issue #557, opened 2018-03-11
Usually exporters can generate more traces than the real traffic.


bogdandrutu:
Every time when we upload to SD (for example) we start a new trace in the gRPC client, which uses the default sampler (and sometimes for low traffic services 1qps for example) using the default sampler as always sample we have a lot of traces for the exporter.

I think we have few options here:
* Wrap the "uploadToStackdriver" (or equivalent for other exporters) in a Span that uses "never sample" sampler.
* Wrap the "uploadToStackdriver" (or equivalent for other exporters) in a Span that uses "1/1000 sample" sampler (this is implemented in Java).
* We can expose the sampler that we use in the option with a default of 1/1000, but I don't think these traces are that useful.
* Recommend users to set a larger BundleDelayThreshold (the current default is 2 seconds) and make probably the default 5 or 10 seconds. (We may want to do this anyway)

/cc @groob who reported similar problem on gitter channel.

ramonza:
Took a cursory look and from what I can tell this should only happen if you explicitly pass ocgrpc.ClientHandler to stackdriver.NewExporter (via ClientOptions). Is that what's happening here?

ramonza:
Confirmed by running stackdriver.TestExport with a dummy project that at least with a very vanilla configuration this should not happen

ramonza:
Actually, I think what's happening here is that the grpc client used by the Stackdriver exporter is also instrumented because of the instrumentation of `google.golang.org/api/transport/grpc`.

PR #560, opened 2018-03-12
Avoid creating spans around Stackdriver submit

ramonza:
The Stackdriver exporter uses a gRPC client that's instrumented with
OpenCensus. This means that there's normally a chance that traces will
be started from the span submission method. In the worst case, with
AlwaysSample, this adds an unbounded number of useless Spans.

This PR follows @bogdandrutu's advice and wraps the upload method in
a Span with NeverSample.

Fixes: #557

### DIFF
--- a/exporter/stackdriver/trace.go
+++ b/exporter/stackdriver/trace.go
@@ -35,7 +35,7 @@ import (
 type traceExporter struct {
 	projectID string
 	bundler   *bundler.Bundler
-	// uploadFn defaults to uploadToStackdriver; it can be replaced for tests.
+	// uploadFn defaults to uploadSpans; it can be replaced for tests.
 	uploadFn func(spans []*trace.SpanData)
 	overflowLogger
 	client *tracingclient.Client
@@ -75,7 +75,7 @@ func newTraceExporterWithClient(o Options, c *tracingclient.Client) *traceExport
 	bundler.BufferedByteLimit = bundler.BundleCountThreshold * 2000
 
 	e.bundler = bundler
-	e.uploadFn = e.uploadToStackdriver
+	e.uploadFn = e.uploadSpans
 	return e
 }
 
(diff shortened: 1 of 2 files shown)


exporter/stackdriver/trace.go
```diff
@@ -116,8 +116,12 @@ func (e *traceExporter) uploadToStackdriver(spans []*trace.SpanData) {
 	for _, span := range spans {
 		req.Spans = append(req.Spans, protoFromSpanData(span, e.projectID))
 	}
-	err := e.client.BatchWriteSpans(context.Background(), &req)
+	// create a never-sampled span to prevent traces associated with exporter
+	span := trace.NewSpan("ignored: uploadToStackdriver", nil, trace.StartOptions{Sampler: trace.NeverSample()})
```
bogdandrutu: Please use a better name because in case of debugging we would look at this spans using the zpages.

exporter/stackdriver/trace.go
```diff
@@ -116,8 +116,12 @@ func (e *traceExporter) uploadToStackdriver(spans []*trace.SpanData) {
...
+	ctx := trace.WithSpan(context.Background(), span) // TODO: add timeouts
+	err := e.client.BatchWriteSpans(ctx, &req)
```
bogdandrutu: Not sure if this can happen (I cannot understand based on the documentation), but if 0 spans are exported we should just immediately return. So I suggest to check that as the first thing in the uploadToStackdriver

exporter/stackdriver/trace.go
```diff
@@ -116,8 +116,12 @@ func (e *traceExporter) uploadToStackdriver(spans []*trace.SpanData) {
 	for _, span := range spans {
 		req.Spans = append(req.Spans, protoFromSpanData(span, e.projectID))
 	}
-	err := e.client.BatchWriteSpans(context.Background(), &req)
+	// create a never-sampled span to prevent traces associated with exporter
+	span := trace.NewSpan("ignored: uploadToStackdriver", nil, trace.StartOptions{Sampler: trace.NeverSample()})
+	ctx := trace.WithSpan(context.Background(), span) // TODO: add timeouts
```
bogdandrutu: Add some useful debugging infos:
* number of exported spans. As an attribute probably.
* log the error as status.

exporter/stackdriver/trace.go
ramonza: I don't think this is possible

ramonza:
Issues addressed, @bogdandrutu PTAL

PR #562, opened 2018-03-12
ochttp Handler and Transport StartOptions support

ramonza:
Users might need to configure the StartOptions used to create spans
by the instrumentation.


ramonza:
@rakyll I don't get that? If we had functional options, we could just as easily have added a `func WithStartOptions(startOptions []StartOption) ochttp.Option`.  Sure, it's a little more boilerplate for the ochttp library, but not really any more public surface area.

rakyll:
> If we had functional options, we could just as easily have added a func WithStartOptions(startOptions []StartOption) ochttp.Option

That's what we are not advocating because it adds duplicated code and exposes different API surface to the end user.

PR #564, opened 2018-03-12
Make Sampler a function

### LINKED ISSUE #490: trace: Sampler interface should just return a boolean instead of a struct with one boolean field
Sampler is defined as
https://github.com/census-instrumentation/opencensus-go/blob/e2fe6beaf0d96b99036db72e68f97157c140cde7/trace/sampling.go#L43-L47

where SamplingDecision is a struct with a single field that's the boolean Sample
https://github.com/census-instrumentation/opencensus-go/blob/e2fe6beaf0d96b99036db72e68f97157c140cde7/trace/sampling.go#L58-L61

I propose we just define Sampler as
```go
type Sampler interface {
   Sample(p SamplingParameters) bool 
}
```

Unless we plan on expanding SamplingDecision to include other fields, I don't see the need for us to have to an extra level of indirection

rakyll:
Fixes #490.


rakyll:
I don't think this change adds much value. I am not going to merge it if it sounds good to you.

PR #565, opened 2018-03-12
Add imports for packages known to depend on OpenCensus API

ramonza:
Stackdriver / GCP is an early adopter and even though we don't yet
support stable APIs, we want to try to avoid breaking them. Added
imports for packages we know are using our APIs to the Stackdriver
tests so that "go vet" PR check will pick up any incompatibilities.

### DIFF
--- a/exporter/stackdriver/stackdriver_test.go
+++ b/exporter/stackdriver/stackdriver_test.go
@@ -27,6 +27,13 @@ import (
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/trace"
 	"golang.org/x/net/context/ctxhttp"
+
+	// Import some other packages known to integrate OpenCensus to check for compilation errors.
+	_ "cloud.google.com/go/datastore"
+	_ "cloud.google.com/go/pubsub"
+	_ "cloud.google.com/go/spanner"
+	_ "google.golang.org/api/transport/grpc"
+	_ "google.golang.org/api/transport/http"
 )
 
 func TestExport(t *testing.T) {


rakyll:
This would create a circular dependency and block our merge queue. When we block cloud.google.com/go, we cannot merge until their fix is merged. Their fix cannot be merged until we merge the breaking API.

We need a build that is not going to block our merge queue.

ramonza:
Discussed in person and agreed that we are in a state where we mostly will be able to deprecate, update dependent code and then remove rather than do a "two-phase" commit. @rakyll

odeke-em:
How about we make a simple CI project that gets run after every commit to master? That is add a webhook to this repo that'll trigger a simple run for this project otherwise in deed, we are creating circular dependencies.

All that it'll contain is:

* imports.go
```go
package main

import (
       // Import some other packages known to integrate OpenCensus to check for compilation errors.
	_ "cloud.google.com/go/datastore"
	_ "cloud.google.com/go/pubsub"
	_ "cloud.google.com/go/spanner"
	_ "google.golang.org/api/transport/grpc"
	_ "google.golang.org/api/transport/http"
)

func main() {}
```


* script
```shell
go get -u opencensus-go-dependencies
```

On every commit hook, run the script and if there are any problems, email contributors or post on the offending PR.

ramonza:
@odeke-em what's the advantage of this over the current PR? as far as I can tell it would have the same effect (tuning checks red on breakages)?

odeke-em:
@ramonza, @rakyll talks about the circular dependency at https://github.com/census-instrumentation/opencensus-go/pull/565#issuecomment-372472549

>This would create a circular dependency and block our merge queue. When we block cloud.google.com/go, we cannot merge until their fix is merged. Their fix cannot be merged until we merge the breaking API.

and I do too in https://github.com/census-instrumentation/opencensus-go/pull/565#issuecomment-372761329
>otherwise in deed, we are creating circular dependencies.

Doing it out out of OpenCensus would solve that problem or at least each project can be fixed independently of the other.

ramonza:
@odeke-em there are already many circular dependencies between repositories (not between packages). yeah in an ideal world, we'd be able to get away without these. but I don't think that this is the most egregious one (it's only in test code) and it's much simpler than creating a separate repo. let's just try this out, we can always revert if it causes problems. i just want to avoid breaking them again, the cost is very small to us to add this.

exporter/stackdriver/stackdriver_test.go
```diff
@@ -27,6 +27,13 @@ import (
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/trace"
 	"golang.org/x/net/context/ctxhttp"
+
+	// Import some other packages known to integrate OpenCensus to check for compilation errors.
```
rakyll: I'd prefer us to have them in an internal package not to confuse any reader.

ramonza:
Seems like their build is broken right now. I'm going to hold off on merging this until we get a better sense for how stable their build is.

PR #566, opened 2018-03-12
Don't propagate method and canonical_status for gRPC requests

### LINKED ISSUE #541: plugin/ocgrpc: do not insert "method" during tag propagation
The ocgrpc plugin inserts the "method" key during tag propagation. I found this in real life usage and I had to manually get around this with
https://github.com/orijtech/opencensus-experiments/blob/0ac9edcefcf4273c0d4f6dfa85a53cfc8e527f66/integration/go-grpc-client/interop_test.go#L103
screen shot 2018-03-08 at 5 49 41 pm

and @ramonza encouraged me to file an issue

This is where we insert it
https://github.com/census-instrumentation/opencensus-go/blob/2ecd8d94fb0d89f945d4911143a0d640bce45936/plugin/ocgrpc/client_stats_handler.go#L52-L54

rakyll:
These tags should be created at the node and applied when recording
measurements. They shouldn't be propagated.

Fixes #541.

### DIFF
--- a/plugin/ocgrpc/server_stats_handler.go
+++ b/plugin/ocgrpc/server_stats_handler.go
@@ -39,9 +39,12 @@ func (h *ServerHandler) statsTagRPC(ctx context.Context, info *stats.RPCTagInfo)
 		}
 		return ctx
 	}
-	d := &rpcData{startTime: startTime}
-	ctx, _ = h.createTags(ctx, info.FullMethodName)
-	ocstats.Record(ctx, ServerStartedCount.M(1))
+	d := &rpcData{
+		startTime: startTime,
+		method:    info.FullMethodName,
+	}
+	ctx, _ = h.createTags(ctx)
+	record(ctx, d, "", ServerStartedCount.M(1))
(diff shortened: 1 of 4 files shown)


plugin/ocgrpc/server_stats_handler.go
```diff
@@ -108,31 +111,27 @@ func (h *ServerHandler) handleRPCEnd(ctx context.Context, s *stats.End) {
...
 }
 
 // createTags creates a new tag map containing the tags extracted from the
 // gRPC metadata.
-func (h *ServerHandler) createTags(ctx context.Context, fullinfo string) (context.Context, error) {
-	mods := []tag.Mutator{
-		tag.Upsert(KeyMethod, methodName(fullinfo)),
+func (h *ServerHandler) createTags(ctx context.Context) (context.Context, error) {
+	buf := stats.Tags(ctx)
+	if buf == nil {
```
odeke-em:
```go
if len(buf) == 0
```

PR #567, opened 2018-03-13
Set status based on HTTP status code

### LINKED ISSUE #457: Set the error code based on the HTTP status code
plugin/ochttp need to set the error code based on the HTTP response code.

Below, there is a reference how HTTP status codes maps to error codes:

```
	case 200:
		return int32(code.Code_OK)
	case 499:
		return int32(code.Code_CANCELLED)
	case 500:
		return int32(code.Code_UNKNOWN) // Could also be Code_INTERNAL, Code_DATA_LOSS
	case 400:
		return int32(code.Code_INVALID_ARGUMENT) // Could also be Code_OUT_OF_RANGE
	case 504:
		return int32(code.Code_DEADLINE_EXCEEDED)
	case 404:
		return int32(code.Code_NOT_FOUND)
	case 409:
		return int32(code.Code_ALREADY_EXISTS) // Could also be Code_ABORTED
	case 403:
		return int32(code.Code_PERMISSION_DENIED)
	case 401:
		return int32(code.Code_UNAUTHENTICATED)
	case 429:
		return int32(code.Code_RESOURCE_EXHAUSTED)
	case 501:
		return int32(code.Code_UNIMPLEMENTED)
	case 503:
		return int32(code.Code_UNAVAILABLE)
	default:
		return int32(code.Code_UNKNOWN)
```

rakyll:
Setting status code only in cases we are confident enough
to map the HTTP status code to error codes.

Fixes #457.

### DIFF
--- a/plugin/ochttp/trace.go
+++ b/plugin/ochttp/trace.go
@@ -18,7 +18,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"sync"
 
 	"go.opencensus.io/plugin/ochttp/propagation/b3"
 	"go.opencensus.io/trace"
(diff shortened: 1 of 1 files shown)


plugin/ochttp/trace.go
```diff
@@ -71,37 +70,36 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 
 	span.SetAttributes(responseAttrs(resp)...)
+	span.SetStatus(status(resp.StatusCode))
```
ramonza: This means that for unknown codes we set zero status. Instead, I think we should set status "2" for any HTTP response code >= 400. It's better to indicate an error vaguely than to indicate success. We should also add the exact error code as an attribute.

plugin/ochttp/trace.go
```diff
@@ -158,3 +149,30 @@ func responseAttrs(resp *http.Response) []trace.Attribute {
 		trace.Int64Attribute(StatusCodeAttribute, int64(resp.StatusCode)),
 	}
 }
+
+func status(statusCode int) trace.Status {
+	st := trace.Status{}
```
ramonza: This only sets the code. Why not also set the message to the text of the status line?

plugin/ochttp/trace.go
rakyll: We decided to set errors for anything outside of [200, 400). I will update the PR.

rakyll:
All addressed, PTAL.

plugin/ochttp/trace.go
```diff
@@ -158,3 +149,72 @@ func responseAttrs(resp *http.Response) []trace.Attribute {
 		trace.Int64Attribute(StatusCodeAttribute, int64(resp.StatusCode)),
 	}
 }
+
+func status(statusCode int) trace.Status {
+	var code int32
+	if code < 200 || code >= 400 {
+		code = codeUnknown
```
ramonza: HTTP 1XX are not errors

rakyll: HTTP 100s are intermediate status codes, and they shouldn't be final.

For example, HTTP 100 behavior as explained at https://tools.ietf.org/html/rfc7231#section-5.1.1:

A server that sends a 100 (Continue) response MUST ultimately send
      a final status code, once the message body is received and
      processed, unless the connection is closed prematurely.

jadekler:
@rakyll Sorry I'm late (been traveling); is it possible to export the `status(..)` function so that we can get rid of our own `status(..)` impl?

rakyll:
@jadekler, we need to discuss whether we want to export this API from the ochttp package. Can you copy paste the mapping section for now if you want to reuse?

jadekler:
@rakyll Certainly happy to - I'll make a note to check in with y'all when I'm back! 👍

PR #574, opened 2018-03-13
Use HTTP and gRPC guides as the entry point

rakyll:
Users are confused by the low-level API due to the significant
highlight on the README. Document the HTTP and gRPC guides as
the main entry points and reduce the number of low-level API
snippets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,26 +34,28 @@ Currently, OpenCensus supports:
 * [Jaeger][exporter-jaeger] for traces
 * [AWS X-Ray][exporter-xray] for traces
 
+
+## Overview
+
+![OpenCensus Overview](https://i.imgur.com/cf4ElHE.jpg)
+
+In a microservices environment, a user request may go through
+multiple services until there is a response. OpenCensus allows
+you to instrument your services and collect diagnostics data all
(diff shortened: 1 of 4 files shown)


rakyll:
@bogdandrutu, this structure is what I was suggesting.

README.md
```diff
@@ -34,26 +34,28 @@ Currently, OpenCensus supports:
...
-// Get a key to represent user OS.
-key, err := tag.NewKey("my.org/keys/user-os")
-if err != nil {
-	log.Fatal(err)
-}
-```
+Start with instrumenting HTTP and gRPC clients and servers,
+then add additional custom instrumentation.
+
+* [HTTP guide](https://github.com/census-instrumentation/opencensus-go/tree/master/examples/http)
```
odeke-em: Would relative URLs like `(./examples/http)` suffice instead of the absolute hard coded one?

rakyll: IMO, it is better to keep them as they are. They still work from preview pages such as https://github.com/rakyll/opencensus-go/blob/200da9489b3e88e6494c0376ab758462214a826c/README.md.

PR #578, opened 2018-03-13
Update internal UserAgent version to 0.4.0 to match latest release

odeke-em:
Noticed while examining some Stackdriver traces that we are
on release 0.4.0 but the useragent field shows `opencensus-go-v0.1.0`

screen shot 2018-03-13 at 4 12 50 pm


odeke-em:
Perhaps later on once the API has been stabilized and we have free time, we could go generate and retrieve the tag from git directly on compilation of the project.

odeke-em:
Thank you!

PR #581, opened 2018-03-14
Convert Attribute to a struct

ramonza:
This avoid about 30% of allocations in benchmarks with 3 or more
attributes.

```
name                     old time/op    new time/op    delta
StartEndSpan-8              522ns ± 0%     502ns ± 0%   ~     (p=1.000 n=1+1)
SpanWithAnnotations_3-8    1.22µs ± 0%    1.08µs ± 0%   ~     (p=1.000 n=1+1)
SpanWithAnnotations_6-8    1.66µs ± 0%    1.29µs ± 0%   ~     (p=1.000 n=1+1)
TraceID_DotString-8         191ns ± 0%     185ns ± 0%   ~     (p=1.000 n=1+1)
SpanID_DotString-8          172ns ± 0%     174ns ± 0%   ~     (p=1.000 n=1+1)

name                     old alloc/op   new alloc/op   delta
StartEndSpan-8               576B ± 0%      576B ± 0%   ~     (all equal)
SpanWithAnnotations_3-8    1.37kB ± 0%    1.27kB ± 0%   ~     (p=1.000 n=1+1)
SpanWithAnnotations_6-8    1.49kB ± 0%    1.30kB ± 0%   ~     (p=1.000 n=1+1)
TraceID_DotString-8         80.0B ± 0%     80.0B ± 0%   ~     (all equal)
SpanID_DotString-8          56.0B ± 0%     56.0B ± 0%   ~     (all equal)

name                     old allocs/op  new allocs/op  delta
StartEndSpan-8               4.00 ± 0%      4.00 ± 0%   ~     (all equal)
SpanWithAnnotations_3-8      13.0 ± 0%      10.0 ± 0%   ~     (p=1.000 n=1+1)
SpanWithAnnotations_6-8      18.0 ± 0%      12.0 ± 0%   ~     (p=1.000 n=1+1)
TraceID_DotString-8          3.00 ± 0%      3.00 ± 0%   ~     (all equal)
SpanID_DotString-8           3.00 ± 0%      3.00 ± 0%   ~     (all equal)
```

### DIFF
--- a/trace/basetypes.go
+++ b/trace/basetypes.go
@@ -41,46 +41,26 @@ type Annotation struct {
 	Attributes map[string]interface{}
 }
 
-// Attribute is an interface for attributes;
-// it is implemented by BoolAttribute, IntAttribute, and StringAttribute.
-type Attribute interface {
-	isAttribute()
+// Attribute represents a key-value pair on a span, link or annotation.
+// Construct with one of: BoolAttribute, Int64Attribute, or StringAttribute.
+type Attribute struct {
+	key   string
+	value interface{}
 }
 
 // BoolAttribute returns a bool-valued attribute.
 func BoolAttribute(key string, value bool) Attribute {
-	return boolAttribute{key: key, value: value}
+	return Attribute{key: key, value: value}
 }
-
-type boolAttribute struct {
-	key   string
-	value bool
-}
-
-func (b boolAttribute) isAttribute() {}
-
-type int64Attribute struct {
-	key   string
-	value int64
-}
-
-func (i int64Attribute) isAttribute() {}
 
 // Int64Attribute returns an int64-valued attribute.
 func Int64Attribute(key string, value int64) Attribute {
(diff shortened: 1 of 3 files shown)


ramonza:
Not sure why the `delta` column displays just "~" but you can see that the numbers did change. This test compares the numbers after making everything sampled (before, we didn't benchmark the sampled case, I also added sampled/not-sampled benchmarks).

odeke-em:
>Not sure why the delta column displays just "~" but you can see that the numbers did change

If you didn't already, could you please run the benchmarks say 10 times or so e.g
```shell
go test -run=^$ -bench=. -count=10
```

ramonza:
Unfortunately it isn't easy to reproduce the above table with "count=10" because I changed the benchmarks themselves in this PR. In any case, the performance gain is not the only reason for this change - it is more idiomatic (returning structs rather than interfaces) and it's a net code reduction (except for more test code).

Issue #547, opened 2018-03-09
Double-negative NoTrace, NoStats options in ocgrpc and ochttp


ramonza:
We should try to find a way to avoid these double-negatives.

rakyll:
We don't have to. It is an explicit request from user and it is little cost if both are disabled.

ramonza:
what do you mean "little cost if both are disabled"? Do you mean if neither option is specified, or if they are specified (meaning that trace and stats are disabled)? 

What was the user requested feature?

rakyll:
If the user both disables stats and traces (NoStats=true, NoTrace=true), it doesn't add additional performance overhead to the Transport or Handler.

OTOH, having (NoStats=true, NoTrace=true) will be a valid case once we have tag propagation over HTTP implemented. User might just want to use ochttp for tag propagation.

ramonza:
I wasn't really thinking about performance, I was thinking about how confusing it sounds having a `NoTrace bool` parameter rather than `Trace bool` just so that the zero value is what we want.

rakyll:
If we have `Trace bool` and `Stats bool`, they are most likely to be set by every user all the time. I expect users coming to this plugin to enable at least trace or stats. Making it optional to enable them sounds like the opposite of why this package exists in the first place.

Most of the time, users shouldn't care about disabling but they have the option to disable.

ramonza:
Yeah I know, I don't want people to type that either. For stats, I think we can get to the point where if there are not views defined there is very little cost, so we can simply remove the option. For trace, we can maybe get to the same point if we allow the user to customize the StartOptions and replace the Sampler with something that never samples. To avoid tag/trace propagation, they can set the propagation format (for HTTP) or we can provide a toggle (for gRPC).

rakyll:
We briefly chatted about this issue offline and also removed NoTrace from ochttp and octrace. I assume renaming NoStats to DisableStats is the last thing left to do. Once the rename is in, we can close this issue.

rakyll:
Closed via 283e9037189fd258e9fffc6263e9aa0f0890ab95.

PR #585, opened 2018-03-14
Remove NoTrace from ochttp and ocgrpc integrations

ramonza:
NoTrace was previously used to disable trace integration. This can
now mostly be achieved by customizing the sampler that each integration
takes as an argument.

The one thing that can no longer be achieved after this PR is
disabling trace propagation while still using the plugin. I think
this is really a good thing: trace propagation should not be optional.

### DIFF
--- a/plugin/ocgrpc/client.go
+++ b/plugin/ocgrpc/client.go
@@ -15,6 +15,7 @@
 package ocgrpc
 
 import (
+	"go.opencensus.io/trace"
 	"golang.org/x/net/context"
 
 	"google.golang.org/grpc/stats"
@@ -23,13 +24,12 @@ import (
 // ClientHandler implements a gRPC stats.Handler for recording OpenCensus stats and
 // traces. Use with gRPC clients only.
 type ClientHandler struct {
-	// NoTrace may be set to disable recording OpenCensus Spans around
-	// gRPC methods.
-	NoTrace bool
-
 	// NoStats may be set to disable recording OpenCensus Stats around each
 	// gRPC method.
 	NoStats bool
+
+	// StartOptions allows configuring the StartOptions used to create new spans.
+	StartOptions trace.StartOptions
 }
 
 func (c *ClientHandler) HandleConn(ctx context.Context, cs stats.ConnStats) {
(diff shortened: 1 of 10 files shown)


ramonza:
> Should we be giving clients the ability to change the samplers they'd like to use? 

Yes, who else could possibly make this decision?

> Perhaps we should be doing more of setting the default tracer once and then either turning it off entirely or letting tracing continue. 

I don't understand this statement, could you clarify what you mean by "turning it off entirely or letting tracing continue"?

> How will the transport's Sampler and the internal defaultSampler interact?

The default sampler will be used if no more specific sampler is provided. In most cases I hope that the default will be enough. If users have configured OpenCensus, I think it is safe to assume that they would like to get tracing. The option NoTrace existed to allow people to disable tracing specifically in the rare case that they want to only get stats. This is still possible, they just need to fiddle with the StartOptions. Since it's expected to be pretty uncommon, I am ok with that.

odeke-em:
>Yes, who else could possibly make this decision?

Not let them dynamically change the sampling decisions. Let them use the same sampling decision of the default sampler. If they'd like to turn off tracing entirely, use the NoTrace or DisableTracing field.

odeke-em:
*Not let them dynamically change the sampling decisions
Not let them dynamically change the sampling decisions per transport

ramonza:
If you're arguing against the StartOptions.Sampler field, I think that is a larger discussion than this PR. Given that it exists and that it can be used to disable tracing, we can rely on it and avoid the NoTrace toggle.

FWIW, I hope to also eventually get rid of NoStats/DisableStats by making stats overhead so low that there is no compelling reason to disable it. See #547. This will require a little more perf work and coordination with Google Cloud SDK who depend on NoStats right now.

ramonza:
I would like to handle NoStats separately since it is used externally.

odeke-em:
>If you're arguing against the StartOptions.Sampler field, I think that is a larger discussion than this PR. Given that it exists and that it can be used to disable tracing, we can rely on it and avoid the NoTrace toggle.

Gotcha, I'll think more on this and ping you to ask more questions. Thank you for bearing with me. For now I don't have other objections.

PR #588, opened 2018-03-15
Ensure rpcData fields are aligned properly on 32-bit architectures

### LINKED ISSUE #587: Atomic access to unaligned 64 bit causes exception on ARM processors
I am using pubsub with go on a Raspberry Pi and I have an exception in opencensus code:

```
panic: runtime error: invalid memory address or nil pointer dereference
...
goroutine 1 [running]:
sync/atomic.loadUint64(0x127660a4, 0xbea27cbc, 0x178bf9a)
        /usr/local/go/src/sync/atomic/64bit_arm.go:10 +0x3c
go.opencensus.io/plugin/ocgrpc.(*ClientHandler).handleRPCEnd(0x1268c614, 0x4a3400, 0x1275c2a0, 0x12762280)
        /home/pi/go/src/go.opencensus.io/plugin/ocgrpc/client_stats_handler.go:109 +0x9c
go.opencensus.io/plugin/ocgrpc.(*ClientHandler).statsHandleRPC(0x1268c614, 0x4a3400, 0x1275c2a0, 0x4a17e8, 0x12762280)
```

The exception seems to be caused by an unaligned atomic access to a 64bit value. In the function `handleRPCEnd`, you have the following code:

```
reqCount := atomic.LoadInt64(&d.reqCount)
respCount := atomic.LoadInt64(&d.respCount)
```   

and the variable `d` is a `rpcData struct` defined as:

```
type rpcData struct {
	startTime           time.Time
	reqCount, respCount int64 // access atomically
	method              string
}
```

The `reqCount` and `respCount` fields are after a `time.Time` field defined as:

```
type Time struct {
	wall uint64
	ext  int64
	loc *Location
}
```

`loc` is a pointer to a `Location` and is 32bit on a Raspberry Pi, so the size of the `Time` structure is 20 Bytes (2\*8 + 1\*4). It seems that the ARM processor can't access a 64bit value atomically if its address is not aligned at a multiple of 8.

I was able to solve this issue by putting `reqCount` and `respCount` at the top of the structure like this:

 ```
type rpcData struct {
	reqCount, respCount int64 // access atomically
	startTime           time.Time
	method              string
}
```

and with this change, everything seems to work as expected.

rakyll:
On 32-bit architectures, it is the caller's responsibility
to arrange for 64-bit alignment for atomic access.

Fixes #587.

### DIFF
--- a/plugin/ocgrpc/stats_common.go
+++ b/plugin/ocgrpc/stats_common.go
@@ -31,12 +31,15 @@ type grpcInstrumentationKey string
 // and end of an call. It holds the info that this package needs to keep track
 // of between the various GRPC events.
 type rpcData struct {
+	// reqCount and respCount has to be the first words
+	// in order to be 64-aligned on 32-bit architectures.
+	reqCount, respCount int64 // access atomically
+
 	// startTime represents the time at which TagRPC was invoked at the
 	// beginning of an RPC. It is an appoximation of the time when the
 	// application code invoked GRPC code.
-	startTime           time.Time
-	reqCount, respCount int64 // access atomically
-	method              string
+	startTime time.Time
+	method    string
 }
 
 // The following variables define the default hard-coded auxiliary data used by


odeke-em:
Cool!

PR #589, opened 2018-03-15
Fix zpages

ramonza:


zpages/tracez.go
```diff
@@ -392,3 +392,57 @@ func writeTextTraces(w io.Writer, data traceData) {
...
+}
+
+type summaryPageRow struct {
+	Name    string
+	Active  int
+	Latency []int
+	Errors  int
+}
+
+func (s *summaryPageData) Len() int           { return len(s.Rows) }
```
odeke-em: We use Go1.8+ and can use `sort.Slice` which is available https://github.com/golang/go/blob/cceee685be4436c8fd417e9c248429b8625b978d/api/go1.8.txt#L243

zpages/rpcz_test.go
```diff
@@ -0,0 +1,117 @@
+// Copyright 2017, OpenCensus Authors
```
odeke-em: It's 2018 :)

rakyll:
Not familiar with the summary page and adding @songy23.

ramonza:
This summary page code is just moved between files. It was only used from tracez, not rpcz, so it makes not sense to have it in rpcz.go.

songy23:
That's right, the open source rpcz page should only contain information on stats, not trace spans. Spans should be displayed on tracez instead.

Issue #590, opened 2018-03-15
all: avoid defer as much as possible unless for correctness


odeke-em:
I noticed that some of our code uses `defer` in short skips of code. For example 
https://github.com/census-instrumentation/opencensus-go/blob/01b39586f932bbb0b4e1cb6499e454c470e9138b/trace/trace.go#L215-L219

Notice that that code could simply be 
```go
exportersMu.Lock() 
for e := range exporters { 
 	e.ExportSpan(sd) 
 }
exportersMu.Unlock() 
```
defers cost way more to the overall runtime and sometime last year I benchmarked small experiments which revealed a 2X cost for a simple lock vs unlock using defer https://gist.github.com/odeke-em/f8fb3a8695b33dd06a700fc3473f3e62

```shell
BenchmarkNoDefer-4   	100000000	        23.3 ns/op	       0 B/op	       0 allocs/op
BenchmarkNoDefer-4   	100000000	        23.3 ns/op	       0 B/op	       0 allocs/op
BenchmarkNoDefer-4   	100000000	        23.3 ns/op	       0 B/op	       0 allocs/op
BenchmarkNoDefer-4   	100000000	        23.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkNoDefer-4   	100000000	        23.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkNoDefer-4   	100000000	        22.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkNoDefer-4   	100000000	        22.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkNoDefer-4   	100000000	        23.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkNoDefer-4   	100000000	        22.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkNoDefer-4   	100000000	        22.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkDefer-4     	30000000	        51.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkDefer-4     	30000000	        51.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkDefer-4     	30000000	        51.1 ns/op	       0 B/op	       0 allocs/op
BenchmarkDefer-4     	30000000	        51.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkDefer-4     	30000000	        51.5 ns/op	       0 B/op	       0 allocs/op
[...]

PR #591, opened 2018-03-15
Remove unnecessary defer in Span.End, shaving significant time

odeke-em:
Updates #590

Defer is quite expensive and in this context, we are using it for
code that's very short. Microbenchmarks of defer show a significant
time improvement such as
https://gist.github.com/odeke-em/f8fb3a8695b33dd06a700fc3473f3e62

With this change, after running the benchmarks 10 times, we
get an ~10% time shave from AlwaysSample and a 4% time shave for
NeverSample i.e. on average 65 nanoseconds for always sampling.

```shell
$ benchstat old.txt new.txt
name                         old time/op    new time/op    delta
StartEndSpan/AlwaysSample-4     679ns ±17%     614ns ± 9%  -9.65%
(p=0.001 n=9+9)
StartEndSpan/NeverSample-4      192ns ± 5%     184ns ± 1%  -4.11%
(p=0.000 n=10+8)

name                         old alloc/op   new alloc/op   delta
StartEndSpan/AlwaysSample-4      576B ± 0%      576B ± 0%    ~     (all
samples are equal)
StartEndSpan/NeverSample-4       128B ± 0%      128B ± 0%    ~     (all
samples are equal)

name                         old allocs/op  new allocs/op  delta
StartEndSpan/AlwaysSample-4      4.00 ± 0%      4.00 ± 0%    ~     (all
samples are equal)
StartEndSpan/NeverSample-4       2.00 ± 0%      2.00 ± 0%    ~     (all
samples are equal)
```

If our sampling probability is 1 in 10000, and used on a system
processing 1BN (1e9 requests) an hour, that's 0.0065 seconds shaved an
hour.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -213,10 +213,10 @@ func (s *Span) End() {
 		if s.spanContext.IsSampled() {
 			// TODO: consider holding exportersMu for less time.
 			exportersMu.Lock()
-			defer exportersMu.Unlock()
 			for e := range exporters {
 				e.ExportSpan(sd)
 			}
+			exportersMu.Unlock()
 		}
 	})
 }


rakyll:
Nice, I almost about to believe defer got so much better in the last release but we may actually go through a few things to avoid defer.

odeke-em:
Offline @ramonza raised a great point that if ExportSpan happens to panic, this lock will be held forever since we don't have the defer. My response was that, yes I acknowledge that, but for the meantime, before we've encountered that problem let's tradeoff waiting for that day of reckoning with speed here.

odeke-em:
@ramonza in the meantime I'll merge this PR, it is all internal stuff so we can revert if we feel strongly about it.

rakyll:
If ExportSpan panics, the process crashes. I don't understand the last comment.

odeke-em:
@rakyll, in the common case sure it will panic but `defer exportsMu.Unlock()` won't be invoked, in the case that Span.End() was invoked by a function that has recover() in it.

rakyll:
In Go, we don't encourage recover at all. Don't optimize behavior for recover.

odeke-em:
for example

```go
func pageContent(w http.ResponseWriter, r *http.Request) {
   defer func() {
       if r := recover(); r != nil {
             log.Printf("Handler crashed: %v\n", r)
       }
   }()

   ctx, span := trace.StartSpan(r.Context(), "/template")
   defer span.End()
   ...
}
```
I've seen a bunch of handlers with recovers in them. In this case exportersMu.Lock() will be held forever yet the program will proceed normally.

rakyll:
This brings the other question of whether it is a good practice to call ExportSpan in the critical path at all.

PR #592, opened 2018-03-16
Benchmarks when not tracing

odeke-em:
For a reality check to see how we are doing, it is essential
that we compare the overhead of work that's done without tracing
entirely. The results are a dunk of one's head in cold water.

1. tracing, sampling: NeverSample vs AlwaysSample
```shell
$ benchstat w_nosample.txt w_sample.txt
name  old time/op    new time/op    delta
It       200ns ± 2%     611ns ± 1%  +205.12%  (p=0.029 n=4+4)

name  old alloc/op   new alloc/op   delta
It        128B ± 0%      576B ± 0%  +350.00%  (p=0.029 n=4+4)

name  old allocs/op  new allocs/op  delta
It        2.00 ± 0%      4.00 ± 0%  +100.00%  (p=0.029 n=4+4)
```

2. no tracing vs NeverSample
```shell
$ benchstat notracing_reuse.txt w_nosample.txt
name  old time/op    new time/op    delta
It      5.68ns ± 2%  200.25ns ± 2%  +3427.08%  (p=0.029 n=4+4)

name  old alloc/op   new alloc/op   delta
It      0.00B ±NaN%   128.00B ± 0%      +Inf%  (p=0.029 n=4+4)

name  old allocs/op  new allocs/op  delta
It       0.00 ±NaN%      2.00 ± 0%      +Inf%  (p=0.029 n=4+4)
```

3. no tracing vs AlwaysSample
```shell
$ benchstat notracing_reuse.txt w_sample.txt
name  old time/op    new time/op    delta
It      5.68ns ± 2%  611.00ns ± 1%  +10661.78%  (p=0.029 n=4+4)

name  old alloc/op   new alloc/op   delta
It      0.00B ±NaN%   576.00B ± 0%       +Inf%  (p=0.029 n=4+4)

name  old allocs/op  new allocs/op  delta
It       0.00 ±NaN%      4.00 ± 0%       +Inf%  (p=0.029 n=4+4)
```

### DIFF
--- a/trace/benchmark_test.go
+++ b/trace/benchmark_test.go
@@ -19,6 +19,65 @@ import (
 	"testing"
 )
 
+func doWork(ctx context.Context) context.Context {
+	vx, _ := ctx.Value("introduction").(string)
+	if len(vx) > 100 {
+		vx = vx[:15]
+		ctx = context.WithValue(ctx, "introduction", vx)
+	}
+	return ctx
+}
+
+func BenchmarkWithSpan_NewContextInLoop_doWork(b *testing.B) {
+	traceBenchmark(b, func(b *testing.B) {
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			ctx := context.Background()
+			_, span := StartSpan(ctx, "/foo")
+			ctx = doWork(ctx)
+			span.End()
+		}
+	})
+}
+
+func BenchmarkWithSpan_ContextReuse_doWork(b *testing.B) {
+	traceBenchmark(b, func(b *testing.B) {
+		ctx := context.Background()
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			_, span := StartSpan(ctx, "/foo")
+			ctx = doWork(ctx)
+			span.End()
+		}
+	})
+}
+
+var sink context.Context
+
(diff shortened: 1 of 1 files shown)


trace/benchmark_test.go
```diff
@@ -94,10 +153,12 @@ func BenchmarkSpanID_DotString(b *testing.B) {
 func traceBenchmark(b *testing.B, fn func(*testing.B)) {
 	b.Run("AlwaysSample", func(b *testing.B) {
 		b.ReportAllocs()
+		b.ResetTimer()
```
ramonza: the order of calling ResetTimer is different here than the below. could we ensure that we call ResetTimer as the very last thing before fn?

ramonza:
You can make the overhead of tracing (percentage-wise) almost arbitrarily large by making "doWork" cheaper. Conversely, you can easily "fix" the high overhead of tracing by replacing "doWork" with something arbitrarily more expensive. 

The point of microbenchmarks is to gradually improve the performance of a small piece of code, not to prove that overhead is acceptable. You can't prove that in a microbenchmark because, as shown above, it all depends on what else your program is doing.

rakyll:
I agree with what @ramonza says. I don't think we should merge this PR.

PR #596, opened 2018-03-16
Add tests for gRPC span duration and Stackdriver gRPC exporting

ramonza:


plugin/ocgrpc/trace_common_test.go
```diff
@@ -110,184 +44,3 @@ func TestClientHandler_traceTagRPC(t *testing.T) {
 		t.Fatal("no metadata")
 	}
 }
-
```
ramonza: moved all this to trace_test.go

PR #598, opened 2018-03-16
Rename NoStats to DisableStats

### LINKED ISSUE #547: Double-negative NoTrace, NoStats options in ocgrpc and ochttp

rakyll:
Fixes #547.


rakyll:
We should instead improve the overhead if there are no view subscriptions.

rakyll:
Closing this give https://github.com/census-instrumentation/opencensus-go/pull/600 looks reasonable and we should instead remove NoStats.

PR #600, opened 2018-03-16
Don't record measurements with no subscription

rakyll:
Suggested by Ramon, immediately dropping the measurements from
the measures with no subscription rather than doing it at a later time
significantly improves the performance on the critical path.

Before:
BenchmarkRecord0-8            	500000000	         3.09 ns/op
BenchmarkRecord1-8            	 5000000	       366 ns/op
BenchmarkRecord8-8            	 3000000	       412 ns/op
BenchmarkRecord8_Parallel-8   	 2000000	       804 ns/op
BenchmarkRecord8_8Tags-8      	 3000000	       415 ns/op

After:
BenchmarkRecord0-8            	500000000	         3.75 ns/op
BenchmarkRecord1-8            	30000000	        39.7 ns/op
BenchmarkRecord8-8            	20000000	       101 ns/op
BenchmarkRecord8_Parallel-8   	30000000	        40.1 ns/op
BenchmarkRecord8_8Tags-8      	20000000	       102 ns/op

### DIFF
--- a/stats/record.go
+++ b/stats/record.go
@@ -22,10 +22,28 @@ import (
 	"go.opencensus.io/tag"
 )
 
+func init() {
+	internal.SubscriptionReporter = func(measure string) {
+		mu.Lock()
+		measures[measure].subscribe()
+		mu.Unlock()
+	}
+}
+
 // Record records one or multiple measurements with the same tags at once.
 // If there are any tags in the context, measurements will be tagged with them.
 func Record(ctx context.Context, ms ...Measurement) {
 	if len(ms) == 0 {
+		return
+	}
+	var record bool
+	for _, m := range ms {
+		if (m != Measurement{}) {
+			record = true
+			break
+		}
+	}
+	if !record {
 		return
 	}
 	if internal.DefaultRecorder != nil {
--- a/stats/view/worker_commands.go
+++ b/stats/view/worker_commands.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/internal"
 	"go.opencensus.io/tag"
 )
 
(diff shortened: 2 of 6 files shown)


rakyll:
We can actually easily use the ref count.

stats/record.go
```diff
@@ -22,10 +22,24 @@ import (
...
+		measures[measure].subscribe()
+		mu.Unlock()
+	}
+}
+
 // Record records one or multiple measurements with the same tags at once.
 // If there are any tags in the context, measurements will be tagged with them.
 func Record(ctx context.Context, ms ...Measurement) {
-	if len(ms) == 0 {
+	var filtered []Measurement
```
ramonza: I don't think it is worth it to construct a filtered slice here. I think we can make this even cheaper by:
1. Instead of having a subscribed call per measure, have `M` construct a `Measure` with nil `m` and define this to mean that the Measurement is ignored
2. Just do a pass over the `ms` and check if all are ignored and drop if that's the case, otherwise pass the original `ms` 
3. In the view code, filter the Measurements by `Measure() != nil` when processing view updates.

This would avoid allocations in `append` and also virtual calls to `subscribed`.

stats/view/worker_commands.go
```diff
@@ -57,6 +58,7 @@ func (cmd *subscribeToViewReq) handleCommand(w *worker) {
 			errstr = append(errstr, fmt.Sprintf("%s: %v", view.Name, err))
 			continue
 		}
+		internal.SubscriptionReporter(view.Measure.Name())
```
ramonza: may as well just pass Measure and call the .subscribe directly

stats/record.go
```diff
@@ -22,10 +22,24 @@ import (
 	"go.opencensus.io/tag"
 )
 
+func init() {
+	internal.SubscriptionReporter = func(measure string) {
```
ramonza: may as well accept Measure and set `internal.SubscriptionReporter = measure.subscribed`

rakyll: I can't, it creates a circular dependency.

stats/view/worker_commands.go
rakyll: I can't, it creates a circular dependency.

stats/record.go
rakyll: Done

Issue #399, opened 2018-02-02
Improve stats.{Count,Sum,Mean,Distribution}Aggregation ergonomics


ramonza:
When creating a view, supplying an aggregation is awkward. It's hard to know what aggregation values are valid. It's also easy to accidentally create pointers to these aggregations, which will type-check but then we have a bunch of things doing type assertions against the base types that will subsequently fail like this: https://play.golang.org/p/BLrZk_I66zV

One suggestion is replace these types with functions returning instances of the type. To solve the discovery issue, we could name them: "AggCount(), AggDist(1,2,3)"

PR #608, opened 2018-03-20
Replace aggregation types with functions

ramonza:
Fixes: #399


ramonza:
discussed with @rakyll yesterday and she preferred to have all of them be functions for consistency.

odeke-em:
Thank you both for the discussion and for the discourse. I see my idea is ugly for consistency as well as for the godoc itself. See
screen shot 2018-03-20 at 3 46 46 pm

ramonza:
verified that this does not break cloud.google.com/go

PR #614, opened 2018-03-20
Set SpanKind for HTTP and gRPC requests

rakyll:


plugin/ocgrpc/client.go
```diff
@@ -25,6 +25,8 @@ import (
 // traces. Use with gRPC clients only.
 type ClientHandler struct {
 	// StartOptions allows configuring the StartOptions used to create new spans.
+	//
+	// SpanKind will always set to SpanKindClient for spans started with this handler.
```
odeke-em: `will always be set to SpanKindClient for spans started by`

plugin/ocgrpc/server.go
```diff
@@ -46,6 +46,8 @@ type ServerHandler struct {
 	// ensures that each service has some opportunity to be traced. If you would
 	// like to not add any additional traces for this gRPC service, set:
 	//   StartOptions.Sampler = trace.ProbabilitySampler(0.0)
+	//
+	// SpanKind will always set to SpanKindServer for spans started from this handler.
```
odeke-em: ditto

plugin/ocgrpc/client.go
rakyll: Done

plugin/ocgrpc/server.go
rakyll: Done

PR #615, opened 2018-03-20
Rename view.Subscribe to view.Register for language consistency

### LINKED ISSUE #612: Rename view.Subscribe to view.Register
To keep consistency with Java and C++, we need to rename view.Subscribe to view.Register and remove the existing deprecated Register functions.

odeke-em:
Fixes #612

* Replace Subscribe with Register in examples
* Deprecate instead of radically removing {"", "Un"}Subscribe

Since Subscribe has been used in many API samples and has
been the function of choice, deprecate it but still underlyingly
use Register. The same for Unsubscribe.

Gradually users should see the deprecation notice and switch over
to using Register and Unregister.

### DIFF
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -73,7 +73,7 @@ func Unsubscribe(v ...*View) error {
 }
 
 // Deprecated: Use the Register function.
-func (v *View) Register() error {
+func (v *View) Subscribe() error {
 	return Register(v)
 }
 
@@ -111,7 +111,7 @@ func Unregister(views ...*View) {
 }
 
 // Deprecated: Use the Unregister function instead.
-func (v *View) Unregister() error {
+func (v *View) Unsubscribe() error {
 	if v == nil {
 		return nil
 	}
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -61,30 +61,31 @@ func Find(name string) (v *View) {
 	return resp.v
 }
 
-// Deprecated: Registering is a no-op. Use the Subscribe function.
-func Register(_ *View) error {
+// Deprecated: Use the Register function
+func Subscribe(v ...*View) error {
+	return Register(v...)
+}
+
+// Deprecated: Use the Unregister function
+func Unsubscribe(v ...*View) error {
+	Unregister(v...)
 	return nil
 }
 
-// Deprecated: Unregistering is a no-op, see: Unsubscribe.
-func Unregister(_ *View) error {
-	return nil
+// Deprecated: Use the Register function.
+func (v *View) Register() error {
+	return Register(v)
 }
 
-// Deprecated: Use the Subscribe function.
-func (v *View) Subscribe() error {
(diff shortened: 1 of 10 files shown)


stats/view/worker.go
```diff
@@ -61,42 +61,43 @@ func Find(name string) (v *View) {
...
+func Unsubscribe(v ...*View) error {
+	Unregister(v...)
 	return nil
 }
 
-// Deprecated: Use the Subscribe function.
-func (v *View) Subscribe() error {
-	return Subscribe(v)
+// Deprecated: Use the Register function.
+func (v *View) Register() error {
```
ramonza: This should not be renamed since the method is only kept for backwards compatibility.

stats/view/worker.go
```diff
@@ -109,12 +110,12 @@ func Unsubscribe(views ...*View) {
 	<-req.done
 }
 
-// Deprecated: Use the Unsubscribe function instead.
-func (v *View) Unsubscribe() error {
+// Deprecated: Use the Unregister function instead.
+func (v *View) Unregister() error {
```
ramonza: This should not be renamed since the method only exists for backwards compatibility

ramonza:
Adding do-not-merge tag so we don't merge until we have checked dependent packages

stats/view/worker.go
odeke-em: Gotcha, thanks. It is a vestige of my earlier commits that I thought I had reverted. I'll fix it right now.

stats/view/worker.go
odeke-em: Roger that.

ramonza:
@odeke-em would you mind looking into the CI failures?

odeke-em:
Thank you for the review @ramonza, I'll take a look at the CI test failure and fix it.

ramonza:
@odeke-em looks like we have some conflicts

odeke-em:
Thanks @ramonza for the ping, I was just syncing up with master and my local branch and going to jump onto this.

odeke-em:
@ramonza, rebased and refreshed, all passes now. PTAL.

odeke-em:
Users|OK|Head
---|---|---
github.com/GoogleCloudPlatform/golang-samples/trace|Yes|https://github.com/GoogleCloudPlatform/golang-samples/blob/098e57c9dbe049b14ee85d950795451cb8eeb22b/trace/trace_quickstart/main.go#L1
cloud.google.com/go/spanner|OK|https://github.com/GoogleCloudPlatform/google-cloud-go/blob/01c1d9b2a0ed098a4e1a7656984d3730e4b56224/spanner/oc_test.go#L1

so cloud/spanner and golang-samples pass

ramonza:
@odeke-em you are not having a good merge day today! looks like what I just merged had some more conflicts, sorry!

odeke-em:
Haha, not a problem at all. I'll fix that.

ramonza:
LGTM

PR #616, opened 2018-03-20
Span names shouldn't be prefixed with Sent and Recv

### LINKED ISSUE #607: HTTP plugin should follow the spec
ochttp package should implement the behavior explained in https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md.

rakyll:
Instead set them for some exporters that cannot
differentiate client spans from server spans in their model.

Fixes #607.

### DIFF
--- a/plugin/ochttp/trace.go
+++ b/plugin/ochttp/trace.go
@@ -50,7 +50,7 @@ type traceTransport struct {
 // The created span can follow a parent span, if a parent is presented in
 // the request's context.
 func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	name := spanNameFromURL("Sent", req.URL)
+	name := spanNameFromURL(req.URL)
 	// TODO(jbd): Discuss whether we want to prefix
 	// outgoing requests with Sent.
 	parent := trace.FromContext(req.Context())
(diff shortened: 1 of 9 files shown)


plugin/ochttp/trace.go
```diff
@@ -126,13 +126,8 @@ func (t *traceTransport) CancelRequest(req *http.Request) {
...
 
-func spanNameFromURL(prefix string, u *url.URL) string {
-	host := u.Hostname()
-	port := ":" + u.Port()
-	if port == ":" || port == ":80" || port == ":443" {
-		port = ""
-	}
-	return prefix + "." + host + port + u.Path
+func spanNameFromURL(u *url.URL) string {
+	return u.Path
```
odeke-em: Will the path be sufficient here?

Requests from "localhost:443/foo", "opencensus.io:8888/foo", "opencensus.io/foo", "ultima.te:9999/foo" will all have the same span name.

rakyll: Users should always breakdown by host. Unlike RPC, there is no concept of a service name in HTTP URLs. The hostname might differ even if the service is the same.

rakyll:
PTAL

PR #620, opened 2018-03-21
Fix plugin/ochttp statusCode mistype and unconditional set

### LINKED ISSUE #619: plugin/ochttp: invalid source code in function status unconditionally sets 200 to codeUnknown
I was boggled while inspecting my Stackdriver traces for a project as to why I had a 200 but had a codeUnknown "2" status
screen shot 2018-03-20 at 11 42 29 pm

But it turns out there is this code
https://github.com/census-instrumentation/opencensus-go/blob/854eae0c1b322c0cda55e7dc63c48233ba12f6fd/plugin/ochttp/trace.go#L154-L157

which looks either like a mistype or a copy and paste error but I believe that we want to set that to be
```go
var code int32 
 if statusCode < 200 || statusCode >= 400 { 
 	code = codeUnknown 
 } 
```

and the regression is traced to https://github.com/census-instrumentation/opencensus-go/pull/567.

Perhaps let's unit test that function.

odeke-em:
Fixes #619

The mistyped source code unconditionally set variable code
to codeUnknown since
```go
var code int32
if code < 200 || code >= 400 {
   code = codeUnknown
}
```
always would set it to codeUnknown as the zero value
of int32 is 0 so always code = codeUnknown.

However if it didn't get matched in the next switch
case, e.g. a 200, we'd get an Unknown status.

Also add unit tests for status so that we never accidentally
regress.

This was discovered by demo app inspections on Stackdriver.

#### Before
screen shot 2018-03-20 at 11 42 29 pm

#### After
screen shot 2018-03-20 at 11 44 26 pm

### DIFF
--- a/plugin/ochttp/trace_test.go
+++ b/plugin/ochttp/trace_test.go
@@ -447,3 +447,27 @@ func TestResponseAttributes(t *testing.T) {
 		})
 	}
 }
+
+func TestStatusUnitTest(t *testing.T) {
+	tests := []struct {
+		in   int
+		want trace.Status
+	}{
+		{200, trace.Status{Code: 0, Message: `"OK"`}},
+		{100, trace.Status{Code: 2, Message: `"UNKNOWN"`}},
+		{500, trace.Status{Code: 2, Message: `"UNKNOWN"`}},
+		{404, trace.Status{Code: 5, Message: `"NOT_FOUND"`}},
+		{600, trace.Status{Code: 2, Message: `"UNKNOWN"`}},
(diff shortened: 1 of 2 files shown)


plugin/ochttp/trace_test.go
```diff
@@ -447,3 +447,27 @@ func TestResponseAttributes(t *testing.T) {
 		})
 	}
 }
+
+func Test_statusUnitTest(t *testing.T) {
```
rakyll: TestStatus

odeke-em:
Done. Thank you for the review @rakyll!

PR #622, opened 2018-03-21
Export trace attributes and annotations properly to Jaeger

### LINKED ISSUE #621: Jaeger exported traces are incomplete
I was experimenting with OpenCensus (thanks for open-sourcing this, it's a great project) and it looks like there are quite a few things I can't get working properly when exporting traces to Jaeger.
I setup a simple project where I create a span and try to add attributes, annotations and a status:
`_, span := trace.StartSpan(ctx, "dummy/api")`
`defer span.End()`

`span.AddAttributes(trace.Int64Attribute("swag", 42))`
`span.Annotate([]trace.Attribute{trace.StringAttribute("attr", "this is an attribute")}, "this is an annotation")`
`span.SetStatus(trace.Status{Code: 2, Message: "failed"})`

The traces are correctly exported to Jaeger and visible in Jaeger UI, but:

- I see the key for the Int64Attribute but the value is empty
- I see a log in the Jaeger UI which contains my 'swag' and 'attr' attributes (but swag is still empty) but it doesn't contain my annotation
- the trace status isn't taken into account

Is there something I'm doing wrong, or is it because there are inconsistencies between Jaeger's format and OpenCensus'?

rakyll:
Fixes #621.

### DIFF
--- a/exporter/jaeger/jaeger.go
+++ b/exporter/jaeger/jaeger.go
@@ -120,26 +120,36 @@ type Exporter struct {
 
 var _ trace.Exporter = (*Exporter)(nil)
 
-// TODO(jbd): Also implement propagation.HTTPFormat.
-
 // ExportSpan exports a SpanData to Jaeger.
 func (e *Exporter) ExportSpan(data *trace.SpanData) {
-	var tags []*gen.Tag
+	e.bundler.Add(spanDataToThrift(data), 1)
+	// TODO(jbd): Handle oversized bundlers.
+}
+
+func spanDataToThrift(data *trace.SpanData) *gen.Span {
+	tags := make([]*gen.Tag, 0, len(data.Attributes))
 	for k, v := range data.Attributes {
 		tag := attributeToTag(k, v)
 		if tag != nil {
 			tags = append(tags, tag)
 		}
 	}
+
+	tags = append(tags,
+		attributeToTag("status.code", data.Status.Code),
+		attributeToTag("status.message", data.Status.Message),
+	)
+
 	var logs []*gen.Log
 	for _, a := range data.Annotations {
-		var fields []*gen.Tag
+		fields := make([]*gen.Tag, 0, len(a.Attributes))
 		for k, v := range a.Attributes {
 			tag := attributeToTag(k, v)
 			if tag != nil {
-				fields = append(tags, tag)
+				fields = append(fields, tag)
 			}
 		}
+		fields = append(fields, attributeToTag("message", a.Message))
 		logs = append(logs, &gen.Log{
 			Timestamp: a.Time.UnixNano() / 1000,
 			Fields:    fields,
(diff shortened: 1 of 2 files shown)


exporter/jaeger/jaeger_test.go
```diff
@@ -49,3 +54,75 @@ func Test_bytesToInt64(t *testing.T) {
...
+	tests := []struct {
+		name string
+		data *trace.SpanData
+		want *gen.Span
+	}{
+		{
+			name: "no parent",
+			data: &trace.SpanData{
+				SpanContext: trace.SpanContext{
+					TraceID: trace.TraceID{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 13, 14, 15, 16},
```
odeke-em: Could you make this a 16 byte TraceID to conform with the usual? This is 15 bytes and it even produces Spans that have 15 byte SpanIDs and TraceIDs.

exporter/jaeger/jaeger.go
```diff
@@ -120,26 +120,36 @@ type Exporter struct {
...
+
 	var logs []*gen.Log
 	for _, a := range data.Annotations {
-		var fields []*gen.Tag
+		fields := make([]*gen.Tag, 0, len(a.Attributes))
 		for k, v := range a.Attributes {
 			tag := attributeToTag(k, v)
 			if tag != nil {
-				fields = append(tags, tag)
+				fields = append(fields, tag)
```
odeke-em: Great catch!

exporter/jaeger/jaeger_test.go
odeke-em: Aha, you missed 12 in there :)

rakyll: Whoops, it was a mistake. Done.

odeke-em: That'll change the value of the TraceIDHigh, *Low, SpanID.
Let's also add tests to ensure that they TraceIDHigh is 8 bytes and same for TraceIDLow otherwise we'd never catch such discrepancies.

rakyll: I don't think we need a test for this. There might be trailing zeros. The trace ID was 16-bytes with a trailing zero which is still valid.

rakyll:
@odeke-em PTAL

exporter/jaeger/jaeger_test.go
```diff
@@ -49,3 +54,75 @@ func Test_bytesToInt64(t *testing.T) {
...
+							"answer": answerValue,
+						},
+					},
+				},
+				Status: trace.Status{Code: 2, Message: "error"},
+			},
+			want: &gen.Span{
+				TraceIdLow:    651345242494996240,
+				TraceIdHigh:   72623859790382856,
+				SpanId:        72623859790382856,
```
odeke-em: How is this possible?

PR #623, opened 2018-03-21
Keep pointer receiver name consistency

rakyll:
The names embedded via measure has m as the
receiver name.


rakyll:
As a side note, I don't know why we embed. Embedding in Go is such an odd concept. I'd prefer to have a named field for measure and copy/paste the accessors.

rakyll:
Closed via b34c6865c1428079adcb73810d52fb1144159cc8.

PR #624, opened 2018-03-21
Instead of having the temporary bridge, fix user's code

rakyll:
Instead of having the temporary bridge, fix the user's code.

Also contains some godoc improvements.


rakyll:
Mailed https://code-review.googlesource.com/c/gocloud/+/25610.

PR #626, opened 2018-03-21
Vendor the Thrift library

rakyll:


rakyll:
I am going to close it for now and we will have a better solution when we are splitting exporters into their own repos.

adriancole:
Fyi was exactly things like this why zipkin changed from default thrift rpc
to default http. I have heard jaeger also has an json http endpoint. If so,
maybe changing to that will buy time on this decision.

For example additional options like thrift rpc for jaeger or kafka or
rabbit could be just that.. options.

Worth reviewing the idea with jaeger people.

PR #627, opened 2018-03-21
Extract stats exporter

ramonza:


exporter/aggregation_data.go
```diff
@@ -0,0 +1,60 @@
...
+	Bounds          []float64 // histogram distribution of the values
+}
+
+// Equal compares two data points and returns true if they are mostly the same,
+// except for small floating point differences. Useful for testing.
+func (a AggregationData) Equal(a2 AggregationData) bool {
+	if len(a.CountPerBucket) != len(a2.CountPerBucket) {
+		return false
+	}
+	for i := range a.CountPerBucket {
```
odeke-em: This is dangerous and can easily lead to bugs later on, could we make it
```go
for i := 0; i < len(a.CountPerBucket); i++ {
```

exporter/aggregation_data.go
```diff
@@ -0,0 +1,60 @@
...
+		if a.CountPerBucket[i] != a2.CountPerBucket[i] {
+			return false
+		}
+	}
+	return a.Count == a2.Count && a.Min == a2.Min && a.Max == a2.Max && math.Pow(a.Mean-a2.Mean, 2) < epsilon && math.Pow(a.variance()-a2.variance(), 2) < epsilon
```
odeke-em: This check is way algorithmically cheaper in the worst case, and should go before the ranging and linear by element comparison

exporter/aggregation_data.go
ramonza: i don't really understand this comment? do you mean to say one should never use this form of range?

exporter/aggregation_data.go
ramonza: this code was just moved around and is also only used in tests.

exporter/export.go
```diff
@@ -0,0 +1,65 @@
...
+// The ExportView method should return quickly; if an
+// Exporter takes a significant amount of time to
+// process a ViewData, that work should be done on another goroutine.
+//
+// The ViewData should not be modified.
+type View interface {
+	ExportView(viewData *ViewData)
+}
+
+// Register registers an exporter. The exporter should implement one of the
```
odeke-em: How about:
`The exporter should implement the View interface`
as there is only one export interface.

exporter/export.go
```diff
@@ -0,0 +1,65 @@
...
+// export interfaces: exporter.View.
+// Collected data will be reported via all the registered exporters.
+// If you no longer want data to be exported, invoke Unregister with the
+// previously registered exporter.
+// Exporters are required to be valid map keys.
```
odeke-em: Nice!

exporter/export.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// View exports the collected view data.
+//
+// The ExportView method should return quickly; if an
+// Exporter takes a significant amount of time to
+// process a ViewData, that work should be done on another goroutine.
+//
+// The ViewData should not be modified.
+type View interface {
```
odeke-em: Hmm, am not sold on making calling this `View`, how about `ViewExporter`?

exporter/export.go
```diff
@@ -0,0 +1,65 @@
...
+	ExportView(viewData *ViewData)
+}
+
+// Register registers an exporter. The exporter should implement one of the
+// export interfaces: exporter.View.
+// Collected data will be reported via all the registered exporters.
+// If you no longer want data to be exported, invoke Unregister with the
+// previously registered exporter.
+// Exporters are required to be valid map keys.
+func Register(exporter interface{}) {
```
odeke-em: Please change this to take a concrete interface to save users from having to deal with runtime failures so that at compile time they'll know if their code compiles or not.

exporter/export.go
```diff
@@ -0,0 +1,65 @@
...
+	exportersMu.Lock()
+	defer exportersMu.Unlock()
+	if ev, ok := exporter.(View); ok {
+		viewExporters[ev] = struct{}{}
+	}
+}
+
+// Unregister unregisters a previously registered exporter.
+func Unregister(exporter interface{}) {
```
odeke-em: This should be a concrete interface e.g
```go
func Unregister(exporter ViewExporter)
```
to avoid runtime surprises.

exporter/export.go
```diff
@@ -0,0 +1,65 @@
...
+	exportersMu.Lock()
+	defer exportersMu.Unlock()
+	if ev, ok := exporter.(View); ok {
```
odeke-em: If this were passed in at compile time as the type, we wouldn't have to do this.

exporter/stackdriver/stats.go
```diff
@@ -264,17 +263,17 @@ func (e *statsExporter) createMeasure(ctx context.Context, vd *view.Data) error
...
 		valueType = metricpb.MetricDescriptor_DOUBLE
-	case view.AggTypeMean:
+	case exporter.AggTypeMean:
 		valueType = metricpb.MetricDescriptor_DISTRIBUTION
-	case view.AggTypeDistribution:
+	case exporter.AggTypeDistribution:
 		valueType = metricpb.MetricDescriptor_DISTRIBUTION
 	default:
-		return fmt.Errorf("unsupported aggregation type: %s", agg.Type.String())
+		return fmt.Errorf("unsupported aggregation type: %s", agg.String())
```
odeke-em: `.String()` automatically be invoked by the "%s" format specifier https://golang.org/pkg/fmt/#hdr-Printing

exporter/stackdriver/stats_test.go
```diff
@@ -99,15 +114,15 @@ func TestExporter_makeReq(t *testing.T) {
...
 	start := time.Now()
 	end := start.Add(time.Minute)
-	count1 := view.CountData(10)
-	count2 := view.CountData(16)
-	sum1 := view.SumData(5.5)
-	sum2 := view.SumData(-11.1)
-	mean1 := view.MeanData{
+	count1 := exporter.AggregationData{Count: 10}
+	count2 := exporter.AggregationData{Count: 16}
+	sum1 := exporter.AggregationData{Mean: 5.5, Count: 1}
```
odeke-em: I personally prefer the old style using the functions, it was consistent and less confusing but it didn't leak
information about the implementation, the user shouldn't have to know the relationship between Mean and Count and Sum. Could we please revert AggregationData as the functions were articulate already?

stats/view/aggregation_data.go
```diff
@@ -17,101 +17,63 @@ package view
...
-// They are reported on the view data during exporting.
-// Mosts users won't directly access aggregration data.
-type AggregationData interface {
-	isAggregationData() bool
+// aggregator receives data points and aggregates them in place.
+type aggregator interface {
 	addSample(v float64)
-	clone() AggregationData
-	equal(other AggregationData) bool
+	writeTo(d *exporter.AggregationData)
```
odeke-em: Aha, reminiscent of Prometheus.

stats/view/aggregation_data.go
```diff
@@ -17,101 +17,63 @@ package view
...
-}
-
-func (a *CountData) equal(other AggregationData) bool {
-	a2, ok := other.(*CountData)
-	if !ok {
-		return false
-	}
-
-	return int64(*a) == int64(*a2)
+func (a *countData) writeTo(dd *exporter.AggregationData) {
```
odeke-em: Perhaps `copyTo`, `write` is a sacred prefix in Go that implies an actual `[]byte` change somewhere.

stats/view/view.go
```diff
@@ -16,18 +16,32 @@
...
-	"time"
 
+	"go.opencensus.io/exporter"
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/internal"
 	"go.opencensus.io/tag"
 )
 
+// Deprecated: use exporter.Register
+func RegisterExporter(e interface{}) {
```
odeke-em: Please use a concrete/compile-time interface.

stats/view/view.go
```diff
@@ -16,18 +16,32 @@
...
+	// TODO(ramonza): move this to the exporter package
+	if e, ok := e.(exporter.View); ok {
+		exporter.Register(e)
+	}
+}
+
+// Deprecated: use exporter.Unregister
+func UnregisterExporter(e interface{}) {
```
odeke-em: Ditto about compile-time vs runtime ready interfaces.

exporter/aggregation_data.go
odeke-em: Nonetheless, if you don't mind, please help fix it.

exporter/aggregation_data.go
odeke-em: I mean that using
```go
for i := range a.CountPerBucket
```
is easier on the eyes.

For an `[]int*` slice could inadvertently lead to the use of i instead of `a.CountPerBucket[i]`. I'd use the former style for maps.

exporter/aggregation_data.go
ramonza: I don't really want to change the body of the function unnecessarily for now, this PR is already very big.

exporter/aggregation_data.go
ramonza: This function was copied out of the stats/view package so I don't really want to change the body for now. I think the variable naming makes it clear that i is an index, not an element of the list.

exporter/stackdriver/stats_test.go
ramonza: The functions actually don't return exporter.AggregationData, they are part of the view package so they can't be used here.

rakyll:
High level comments:

* I'd keep the Aggregation as an interface. We don't know what kind of Aggregation types we will support in the future. Union-like structs are not a very Go friendly approach.

* I'd remove RegisterExporter and UnregisterExporter from the view package. They only encourages users to depend on the view package for exporting. This is a major breaking change, it is ok to break everyone.

Everything else looks good.

ramonza:
@rakyll your comment about Aggregation, I assume you mean exporter.AggregationData (which is union-like) and not view.Aggregation?

ramonza:
Closing this for now since it is just too big to maintain. We need to decide whether we want to go ahead with this or not. If so I can resurrect or (more likely) just re-do the minimal change.

PR #630, opened 2018-03-23
Remove ending dot in documentation URL

bweston92:
The dot is causing SSL common name issues, as GitHub and Godoc linking to the domain name with an extra dot.


bweston92:
I signed it!

PR #639, opened 2018-03-26
Remove charset restrictions on tag values

ramonza:
See: https://github.com/census-instrumentation/opencensus-specs/pull/71

### DIFF
--- a/tag/map_test.go
+++ b/tag/map_test.go
@@ -129,9 +129,7 @@ func TestNewMap(t *testing.T) {
 			name:    "from empty; invalid",
 			initial: nil,
 			mods: []Mutator{
-				Insert(k5, "v\x19"),
-				Upsert(k5, "v\x19"),
-				Update(k5, "v\x19"),
+				Insert(k5, strings.Repeat("x", 256)),
 			},
 			want: nil,
 		},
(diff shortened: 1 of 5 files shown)


tag/validate.go
```diff
@@ -52,5 +52,5 @@ func checkValue(v string) bool {
 	if len(v) > maxKeyLength {
 		return false
 	}
-	return isASCII(v)
+	return true
```
odeke-em: In this case we can fold this entirely into
```go
func checkValue(v string) bool {
        return len(v) <= maxKeyLength
}
```

tag/map_test.go
```diff
@@ -129,9 +129,7 @@ func TestNewMap(t *testing.T) {
 			name:    "from empty; invalid",
 			initial: nil,
 			mods: []Mutator{
-				Insert(k5, "v\x19"),
-				Upsert(k5, "v\x19"),
-				Update(k5, "v\x19"),
+				Insert(k5, strings.Repeat("x", 256)),
```
odeke-em: Perhaps this could be a test of its own since the restriction being newly tested is `len(v) <= 255`

ramonza: added t.Run, so it should appear as it's own test

odeke-em: Cool, thanks. LGTM!

PR #641, opened 2018-03-26
Remove error from stats.{Float,Int}64Measure signatures

ramonza:
Fixes: #544

### DIFF
--- a/exporter/stackdriver/examples/stats/main.go
+++ b/exporter/stackdriver/examples/stats/main.go
@@ -28,6 +28,10 @@ import (
 	"go.opencensus.io/stats/view"
 )
 
+// Create measures. The program will record measures for the size of
+// processed videos and the nubmer of videos marked as spam.
+var videoSize = stats.Int64("my.org/measure/video_size", "size of processed videos", stats.UnitBytes)
+
 func main() {
 	ctx := context.Background()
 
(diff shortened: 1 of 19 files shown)


exporter/stackdriver/examples/stats/main.go
```diff
@@ -49,13 +53,6 @@ func main() {
 	}
 	view.RegisterExporter(exporter)
 
-	// Create measures. The program will record measures for the size of
-	// processed videos and the nubmer of videos marked as spam.
-	videoSize, err := stats.Int64("my.org/measure/video_size", "size of processed videos", "MBy")
```
odeke-em: How about let's bring that back in here instead of creating it as a global variable?

internal/readme/stats.go
```diff
@@ -26,14 +26,11 @@ import (
 // README.md is generated with the examples here by using embedmd.
 // For more details, see https://github.com/rakyll/embedmd.
 
+var videoSize = stats.Int64("my.org/video_size", "processed video size", "MB")
```
odeke-em: Ditto about declaring videoSize where it was originally.

exporter/stackdriver/examples/stats/main.go
ramonza: why? we encourage users to declare these as globals. major justification for this change was to make declaring as globals easier so thought it appropriate to demonstrate this here.

odeke-em: I mean, then sure, let's use it in not only main but say in another helper, otherwise as it is, doesn't look justified to be a global.

ramonza: I will extract the record into a separate function. I think having multiple files for an example would be too much.

odeke-em: In deed. Not multiple files though, just multiple functions.

PR #644, opened 2018-03-27
Export int64 sum as MetricDescriptor_INT64 to Stackdriver

rakyll:
Updates https://github.com/census-instrumentation/opencensus-specs/issues/83.


songy23:
Could you also update `newTypedValue()` (https://github.com/rakyll/opencensus-go/blob/817f33c09ad7f45e284d646a05a0bbe2c5c304d3/exporter/stackdriver/stats.go#L322)?
If the Measure is INT64, sum data should be wrapped to INT64 too.

odeke-em:
@rakyll it might be helpful to update your commit message to mention the issue in question.

Issue #643, opened 2018-03-27
Make it possible to replace the TraceID/SpanID generation algorithms


ramonza:
To support special formats as required by some backends (XRay) we should make the TraceID/SpanID algorithms pluggable. 

See: https://github.com/census-instrumentation/opencensus-specs/issues/64

odeke-em:
Am not sure what @bogdandrutu will think of this change but the last time I spoke with him, there was a reservation towards allowing this.

ramonza:
I think his concern was with replacing the algorithm entirely (which was my original suggestion). This is a more gradual change that users can opt in to.

rakyll:
We briefly chatted about this issue to depend on NewSpanWithRemoteParent with custom ID generation. Btw, Bogdan's concern was not to introduce a new generator.

PR #647, opened 2018-03-28
allow custom for TraceID/SpanID generator

savaki:
See https://github.com/census-instrumentation/opencensus-go/issues/643

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -26,6 +26,42 @@ import (
 	"go.opencensus.io/internal"
 )
 
+// Generator provides a configurable interface to generate a TraceID
+type Generator interface {
+	// NewTraceID generates a new TraceID
+	NewTraceID() TraceID
+}
+
+// GeneratorFunc provides a func interface to Generator
+type GeneratorFunc func() TraceID
+
+// NewTraceID implements IDFactory
+func (fn GeneratorFunc) NewTraceID() TraceID {
+	return fn()
+}
+
+var (
+	// defaultGenerator holds the global id factory
+	defaultGenerator Generator
+)
+
+func init() {
+	SetGenerator(nil)
+}
+
+// SetGenerator sets the generator used to create new traceID.  Useful when a
+// custom TraceID format is required e.g. AWS X-Ray
+func SetGenerator(generator Generator) {
+	mu.Lock()
+	defer mu.Unlock()
+
+	if generator == nil {
+		generator = GeneratorFunc(newTraceIDLocked)
+	}
+
+	defaultGenerator = generator
+}
+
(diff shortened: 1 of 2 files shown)


rakyll:
/cc @bogdandrutu 

I don't think we want to introduce a new generator as a first class citizen to this package at this point while the industry is trying to deprecate all the legacy trace headers. You should be able to generate your own IDs and use NewSpanWithRemoteParent.

rakyll:
I am going to close this issue. Please do not file issues here until there is consensus on the specs repo.  @savaki is spending his time contributing changes that we didn't agree on.

rakyll:
Closed this PR accidentally thinking I was closing the tracking issue oops. @savaki, can generating IDs and using NewSpanWithRemoteParent be good enough?

ramonza:
I should not have suggested this path without getting consensus, sorry for making extra work for everyone.

i had not thought of the solution using NewSpanWithRemoteParent. I think if this works (and doesn't cause XRay to have issues due to the fake parent span) then it is preferable to this for now.

rakyll:
@bogdandrutu pointed out that we are lacking the global TraceConfig explained at https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md.

I think we can put the generator into the config which will be less invasive. See https://github.com/census-instrumentation/opencensus-go/issues/652.

rakyll:
@savaki, can you revisit this PR to add a generator option to trace.Config?

rakyll:
I have one more concern about allowing users to set their custom ID generators. With this approach, we will make them set a global setting to change the whole behavior of the library and breaking it for all the other vendors if user want to register more than one exporter.

I think the right approach to this problem is not to touch the internal representation of IDs at OpenCensus level but to map them to a vendor-specific format when exporting and at propagation.HTTPFormat. Vendor-specific and OpenCensus IDs can be mapped back and forth.

bogdandrutu:
We expect that the application owner will be the one who sets both so it should not be a conflict.

The generator should generate the same sizes 16B for TravceId and 8B for SpanId

rakyll:
If a vendor shows up tomorrow with totally different requirements for span context, are we going to add more abstractions? I think the general solution to this problem is not to add more options but to have the mapping implemented by vendor-specific code.

I can send a PR, it shouldn't really take more than couple of lines to have the mapping done. Changing the core abstractions sounds to invasive to me given this can entirely be done in the vendor-specific package.

bogdandrutu:
@rakyll the mapping has to be global (distributed) so it is not that easy to implement.

rakyll:
Closed via https://github.com/census-instrumentation/opencensus-go/pull/679.

PR #648, opened 2018-03-28
Proxy (http.Hijacker).Hijack() calls to Handler.responseWriter in ochttp

### LINKED ISSUE #642: plugin/ochttp: doesn't implement http.Hijacker
I'm using gorilla websockets on a /websockets route exposed by my http server.
I started using ochttp.Handler to get automagic instrumentation (which is great!) but it apparently doesn't allow hijacking the connection: gorilla errors out saying that the response doesn't implement ([http.Hijacker](https://golang.org/pkg/net/http/#Hijacker).

Do you think it's something that would make sense to implement in ochttp? Or do you recommend exposing my websocket endpoint with a normal http handler separately?

Thanks!

odeke-em:
Fixes #642

Ensure that plugin/ochttp.Handler proxies (http.Hijacker).Hijack()
calls to its responseWriter that we captured. Also add tests to
ensure that the behavior with the Go standard library is preserved
that is:
a) HTTP/1.X connections successfully can use (http.Hijacker).Hijack()
b) HTTP/2.X connections are incompatible with (http.Hijacker).Hijack()
because with HTTP/2.X, multiple requests can be sent on the same
connection thus it doesn't make sense to hijack a connection. The
standard library also panics and enforces this behavior.

Note that both a) and b) are only enforced at runtime, like we do.
We don't need to do any work, except solidify and lock this behavior
in our tests.

As an advantage, we can now successfully use ochttp.Handler with
websockets!

### DIFF
--- a/plugin/ochttp/server.go
+++ b/plugin/ochttp/server.go
@@ -15,7 +15,9 @@
 package ochttp
 
 import (
+	"bufio"
 	"context"
+	"net"
 	"net/http"
 	"strconv"
 	"sync"
@@ -65,7 +67,6 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	defer traceEnd()
 	w, statsEnd = h.startStats(w, r)
 	defer statsEnd()
-
 	handler := h.Handler
 	if handler == nil {
 		handler = http.DefaultServeMux
@@ -140,6 +141,11 @@ type trackingResponseWriter struct {
 }
 
 var _ http.ResponseWriter = (*trackingResponseWriter)(nil)
+var _ http.Hijacker = (*trackingResponseWriter)(nil)
+
+func (t *trackingResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	return t.writer.(http.Hijacker).Hijack()
+}
 
 func (t *trackingResponseWriter) end() {
 	t.endOnce.Do(func() {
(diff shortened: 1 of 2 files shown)


odeke-em:
And now the websocket repro should successfully work as at https://gist.github.com/odeke-em/3c2c664416933f2b349cd662b0b63c68
screen shot 2018-03-27 at 10 29 20 pm

bogdandrutu:
LGTM, I would like one of the Go people to look at this as well. Great job.

plugin/ochttp/server.go
```diff
@@ -140,6 +141,11 @@ type trackingResponseWriter struct {
 }
 
 var _ http.ResponseWriter = (*trackingResponseWriter)(nil)
+var _ http.Hijacker = (*trackingResponseWriter)(nil)
+
+func (t *trackingResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
```
rakyll: Shouldn't we do it only if the underlying writer is a http.Hijacker?

odeke-em: If the question is why we have a `Hijack` method, it is because we can't dynamically create an `http.Hijack` method at runtime for a type. But yes, if the underlying writer is an http.Hijacker, invoking Hijacker() works fine, and if it is on an http2 connection, it panics as the sample code even without OpenCensus shows.

odeke-em: Thanks for the advocacy and discussions offline. We came to the consensus that  if the underlying responseWriter doesn't implement http.Hijacker, return an error, like other implementations do. I've updated that code and tests, PTAL.

PR #653, opened 2018-03-28
Introduce trace.Config

### LINKED ISSUE #652: Introduce trace.Config
The trace package doesn't implement the TraceConfig spec which is the canonical API to set global settings: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md.

We should deprecate the global sampler setter and allow users to set the global config via the new API.

A global configuration setter will be also useful for extending capabilities in the future. And we can allow users to provide custom implementations of generation.

/cc @savaki

rakyll:
trace.Config allows to set global configuration for tracing
and deprecating the global default sampler setter.

Fixes #652.

### DIFF
--- a/trace/config.go
+++ b/trace/config.go
@@ -1,0 +1,36 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trace
+
+func init() {
+	config.DefaultSampler = ProbabilitySampler(defaultSamplingProbability)
+}
+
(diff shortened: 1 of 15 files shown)


trace/config.go
```diff
@@ -0,0 +1,36 @@
+// Copyright 2017, OpenCensus Authors
```
bogdandrutu: 2018

trace/trace.go
```diff
@@ -158,11 +158,12 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
 	span := &Span{}
 	span.spanContext = parent
 	mu.Lock()
+	// TODO(jbd): Reduce the contention.
 	if !hasParent {
 		span.spanContext.TraceID = newTraceIDLocked()
 	}
 	span.spanContext.SpanID = newSpanIDLocked()
-	sampler := defaultSampler
+	sampler := config.DefaultSampler
```
bogdandrutu: Is this thread safe? Shouldn't you use something like this: https://golang.org/pkg/sync/atomic/#Value?

rakyll: This whole block is guarded by a mutex right now.

If you are asking about the reference, in Go, things are pass by value. User cannot mutate the config.DefaultSampler.

rakyll: If you prefer atomic values for all the global state, I added a TODO for that.

trace/config.go
rakyll: Done

PR #655, opened 2018-03-28
Avoid global configuration overriding all

rakyll:
It will be more common for the users to tweak a few
settings rather than all.

Rename SetConfig to ApplyConfig and only
override the non-zero values.

### DIFF
--- a/trace/config_test.go
+++ b/trace/config_test.go
@@ -14,25 +14,15 @@
 
 package trace
 
-func init() {
-	config.DefaultSampler = ProbabilitySampler(defaultSamplingProbability)
-}
+import (
+	"reflect"
+	"testing"
+)
 
-// Config represents the global tracing configuration.
-type Config struct {
-	// DefaultSampler is the default sampler used when creating new spans.
-	DefaultSampler Sampler
-}
-
-// ApplyConfig applies changes to the global tracing configuration.
-//
-// Fields not provided in the given config are going to be preserved.
-func ApplyConfig(cfg Config) {
-	mu.Lock()
-	if cfg.DefaultSampler == nil {
-		cfg.DefaultSampler = config.DefaultSampler
+func TestApplyZeroConfig(t *testing.T) {
+	cfg := config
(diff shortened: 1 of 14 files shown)


trace/config.go
```diff
@@ -24,8 +24,10 @@ type Config struct {
...
 }
 
-// SetConfig sets the global tracing configuration.
-func SetConfig(cfg Config) {
+// ApplyConfig applies changes to the global tracing configuration.
+//
+// Fields not provided in the given config is going to be preserved.
+func ApplyConfig(cfg Config) {
 	if cfg.DefaultSampler == nil {
 		cfg.DefaultSampler = newDefaultSampler()
```
bogdandrutu: Here you don't want to reinstall the default, you want to keep the previous value. Maybe worth having a test.

rakyll: Done.

ramonza:
Nice!

trace/config.go
```diff
@@ -24,12 +24,14 @@ type Config struct {
 	DefaultSampler Sampler
 }
 
-// SetConfig sets the global tracing configuration.
-func SetConfig(cfg Config) {
+// ApplyConfig applies changes to the global tracing configuration.
+//
+// Fields not provided in the given config is going to be preserved.
```
odeke-em: Minor typo:
s/is going to/are going to/

trace/config_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestApplyZeroConfig(t *testing.T) {
+	cfg := config
+	ApplyConfig(Config{})
+	if got, want := reflect.ValueOf(config.DefaultSampler).Pointer(), reflect.ValueOf(cfg.DefaultSampler).Pointer(); got != want {
```
odeke-em: Sorry, what is this testing for?

trace/config.go
rakyll: Done

PR #660, opened 2018-03-30
*ResponseStatusCode* measures and views to use Distribution aggregations

### LINKED ISSUE #659: plugin/ochttp: invalid Measures for *ResponseCountByStatusCode views cause prometheus `inconsistent label cardinality` error
Running the program below
```go
package main

import (
	"context"
	"log"
	"net/http"

	"go.opencensus.io/exporter/prometheus"
	"go.opencensus.io/stats"
	"go.opencensus.io/stats/view"
)

func main() {
	pe, err := prometheus.NewExporter(prometheus.Options{})
	if err != nil {
		log.Fatalf("Creating prometheus Exporter error: %v", err)
	}
	view.RegisterExporter(pe)

	if err := view.Register(ochttp.DefaultServerViews...); err != nil {
		log.Fatalf("Register ochttp.DefaultServerViews error: %v", err)
	}

	points := []float64{0, 1, 10, 20, 40, 80, 160, 180, 250, 500, 800, 1e3, 1.2e3, 2.2e3, 4.3e3,
		9.7e3, 5e2, 17.2e2, 20e1}
	go func() {
		ctx := context.Background()
		for _, point := range points {
			stats.Record(ctx, ochttp.ServerLatency.M(point))
		}
	}()

	mux := http.NewServeMux()
	mux.Handle("/metrics", pe)
	log.Fatal(http.ListenAndServe(":9988", pe))
}
```

I get
```shell
$ go run main.go 
2018/03/29 23:51:15 Failed to export to Prometheus: inconsistent label cardinality
2018/03/29 23:51:15 Failed to export to Prometheus: inconsistent label cardinality
```
and from examining
https://github.com/census-instrumentation/opencensus-go/blob/8069fe56546c2cdcfbeb5290f5c6130590927308/plugin/ochttp/stats.go#L102-L109

and

https://github.com/census-instrumentation/opencensus-go/blob/8069fe56546c2cdcfbeb5290f5c6130590927308/plugin/ochttp/stats.go#L146-L153

those measures look wrong. Did we really mean to set those to *Latency yet the view is `ServerResponseCountByStatusCode`. I believe @songy23 once raised such a question in a conversation somewhere about `ServerResponseCountByStatusCode` but perhaps got lost in a meeting or so.

odeke-em:
Fixes #659

* ClientResponseStatusCode and ServerResponseStatusCode measures can
now record within the respective track* wrappers for client and server
respectively.
* ClientResponseCountByStatusCode and ServerResponseCountByStatusCode
no longer use ClientLatency and ServerLatency measures and are no longer
based on Count aggregations, but now on Distribution aggregations. Also
introduced buckets for status codes:
  [99, 199, 299, 399, 499, 599, 699]
to account for status code buckets:
  [0, 1XX, 2XX, 3XX, 4XX, 5XX, 6XX]

Using *Latency measures doesn't make sense to try to correlate response
count by status code, which arguably are just the view groupings so that
request and response counts can be tagged by StatusCode, but it just
felt so off with that combination.

### DIFF
--- a/plugin/ochttp/client_stats.go
+++ b/plugin/ochttp/client_stats.go
@@ -95,6 +95,7 @@ func (t *tracker) end() {
 		m := []stats.Measurement{
 			ClientLatency.M(float64(time.Since(t.start)) / float64(time.Millisecond)),
 			ClientResponseBytes.M(t.respSize),
+			ClientResponseStatusCode.M(int64(t.statusCode)),
 		}
 		if t.reqSize >= 0 {
 			m = append(m, ClientRequestBytes.M(t.reqSize))
--- a/plugin/ochttp/server.go
+++ b/plugin/ochttp/server.go
@@ -149,6 +149,7 @@ func (t *trackingResponseWriter) end() {
 		m := []stats.Measurement{
 			ServerLatency.M(float64(time.Since(t.start)) / float64(time.Millisecond)),
 			ServerResponseBytes.M(t.respSize),
+			ServerResponseStatusCode.M(int64(t.statusCode)),
 		}
 		if t.reqSize >= 0 {
 			m = append(m, ServerRequestBytes.M(t.reqSize))
--- a/plugin/ochttp/stats.go
+++ b/plugin/ochttp/stats.go
@@ -22,18 +22,20 @@ import (
 
 // The following client HTTP measures are supported for use in custom views.
 var (
-	ClientRequestCount  = stats.Int64("opencensus.io/http/client/request_count", "Number of HTTP requests started", stats.UnitNone)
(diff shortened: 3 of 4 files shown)


ramonza:
These are defined by OpenCensus specs. status_code is meant to be a tag, not the value of the measure. I think any change of this nature would need to go on the specs repo, not here. I also think that using status code as the measure value is just wrong - it is not a measure, it is a tag/label because it has a small fixed cardinality.

rakyll:
Status code is not a measurement but a dimension. You would rather try to understand the count of the responses broken by status code. Please open a spec issue to further discuss this possibility.

PR #664, opened 2018-04-02
Fix equalAggTagKeys() in Stackdriver Exporter.

songy23:
Sum Aggregation should match INT64 or DOUBLE, and LastValue should match
DOUBLE.


songy23:
No problem :)

PR #667, opened 2018-04-03
Remove deprecated APIs

### LINKED ISSUE #494: Clean up deprecated APIs
There are many APIs labelled as deprecated, this is not an ideal situation for the new users. We need to remove them before finalizing R2 and setting a deprecation policy for the beta status.

rakyll:
Fixes #494.

### DIFF
--- a/plugin/ochttp/client_test.go
+++ b/plugin/ochttp/client_test.go
@@ -36,8 +36,8 @@ func TestClient(t *testing.T) {
 	}))
 	defer server.Close()
 
-	for _, v := range ochttp.DefaultClientViews {
-		v.Subscribe()
+	if err := view.Register(ochttp.DefaultClientViews...); err != nil {
+		t.Fatalf("Failed to register ochttp.DefaultClientViews error: %v", err)
 	}
 
 	views := []string{
@@ -54,15 +54,14 @@ func TestClient(t *testing.T) {
 		}
 	}
 
-	var (
-		w    sync.WaitGroup
-		tr   ochttp.Transport
-		errs = make(chan error, reqCount)
-	)
-	w.Add(reqCount)
+	var wg sync.WaitGroup
+	var tr ochttp.Transport
+	errs := make(chan error, reqCount)
+	wg.Add(reqCount)
+
 	for i := 0; i < reqCount; i++ {
 		go func() {
-			defer w.Done()
+			defer wg.Done()
 			req, err := http.NewRequest("POST", server.URL, strings.NewReader("req-body"))
 			if err != nil {
 				errs <- fmt.Errorf("error creating request: %v", err)
@@ -81,7 +80,7 @@ func TestClient(t *testing.T) {
 	}
 
 	go func() {
-		w.Wait()
+		wg.Wait()
 		close(errs)
 	}()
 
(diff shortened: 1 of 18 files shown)


plugin/ochttp/client_test.go
```diff
@@ -36,9 +36,7 @@ func TestClient(t *testing.T) {
 	}))
 	defer server.Close()
 
-	for _, v := range ochttp.DefaultClientViews {
-		v.Subscribe()
-	}
+	view.Register(ochttp.DefaultClientViews...)
```
odeke-em:
```go
if err := view.Register(ochttp.DefaultClientViews...); err != nil {
   t.Fatalf("Failed to register ochttp.DefaultClientViews error: %v", err)
}
```

plugin/ochttp/server_test.go
```diff
@@ -34,9 +34,7 @@ func updateMean(mean float64, sample, count int) float64 {
 }
 
 func TestHandlerStatsCollection(t *testing.T) {
-	for _, v := range DefaultServerViews {
-		v.Subscribe()
-	}
+	view.Register(DefaultServerViews...)
```
odeke-em: Ditto about handling the error on Registration.

stats/view/worker_commands.go
```diff
@@ -41,7 +41,12 @@ type getViewByNameResp struct {
 }
 
 func (cmd *getViewByNameReq) handleCommand(w *worker) {
-	cmd.c <- &getViewByNameResp{w.views[cmd.name].view}
```
odeke-em: I think this new change now hides an error that should never occur i.e. a nil viewInternal.

trace/sampling.go
```diff
@@ -24,12 +24,6 @@ func newDefaultSampler() Sampler {
 	return ProbabilitySampler(defaultSamplingProbability)
 }
 
-// SetDefaultSampler sets the default sampler used when creating new spans.
-//
-// Deprecated: Use ApplyConfig.
-func SetDefaultSampler(sampler Sampler) {
```
odeke-em: This might be too early to remove, there are a bunch of examples and running code in the wild using SetDefaultSampler, perhaps let's leave this one in for a bit, or did we already remove it? I think changing it recently was a mistake as we should give some sunsetting time.

odeke-em: Alright then, but before we merge this PR, we'll need to perhaps fix Google Cloud
https://github.com/GoogleCloudPlatform/google-cloud-go/blob/01dbcbbaf95d5ec2b07fe6dcd7db442be0f5b0c4/internal/testutil/go18.go#L42
and perhaps ping a few other folks such as https://blog.doit-intl.com/measure-once-export-anywhere-opencensus-in-the-wild-61724f44eb00 and also my examples.

odeke-em: I'll mail the one for cloud-go.

odeke-em: I've mailed https://code-review.googlesource.com/#/c/gocloud/+/26130 but there is trouble logging in :)

rakyll:
Also sent https://code-review.googlesource.com/c/gocloud/+/26150.

rakyll:
PTAL

PR #672, opened 2018-04-04
Fix the broken Apache Thrift import path

rakyll:
Thrift's import path is broken. Use the GitHub import
path instead. We don't export any APIs depending on Thrift
symbols, so this change is safe.


rakyll:
Not gonna fix anything because Zipkin is also depending on this import path. Instead, I asked them on the mailing list to fix the import path.

PR #679, opened 2018-04-05
Introduce an internal-only ID generator

rakyll:
We introduce this experimental generator to support AWS X-Ray.
Regular users should not depend on this API, it is preserved
to be used by the exporters.


rakyll:
We can provide an accessor for the default ID generator instance if we want user to be able to revert back to the default.

PR #682, opened 2018-04-06
Span.End should return ASAP if no exporters + if not sampled

### LINKED ISSUE #681: trace: Span.End should return ASAP if there are no exporters and perhaps drop spanData
Just examining this code 
https://github.com/census-instrumentation/opencensus-go/blob/ead56c48fbe540f3601b21ff9eeea1fb60300740/trace/trace.go#L214-L234

If we don't have any exporters and if not `(*Span).IsSampled()`, we'll be doing wasted work here
https://github.com/census-instrumentation/opencensus-go/blob/ead56c48fbe540f3601b21ff9eeea1fb60300740/trace/trace.go#L225-L233

Perhaps let's remove that? In addition perhaps we should drop `(*Span).data` so that any further invocations of `End()` will return asap on this call
https://github.com/census-instrumentation/opencensus-go/blob/ead56c48fbe540f3601b21ff9eeea1fb60300740/trace/trace.go#L215-L217

odeke-em:
Fixes #681

* Span.End should return ASAP if no exporters are provided to
avoid the overhead of preparing data that won't be used
* If the span is not sampled, no need to
invoke makeSpanData(). The current user is the spanStore
that marks it as finished but not sure that it is used.
* Since (*Span).spanContext is a non-pointer value, it is always
zeroed as is the case when it is a root span. Only set
(*Span).spanContext = parent iff hasParent is true thus
bringing the number of sets in the case when it is a root span
always to 1 instead of 2 before.
* Update the benchmarks to include the case with zero and then one
exporter.
* Also unroll the benchmarking code as it was giving spurious/wrong
results when testing no vs with exporters despite us invoking
b.ResetTimer(). Perhaps this is just a flake, but anyways the
benchmarks aren't radically changing so for now it is okay
to have them as they are, unrolled and unrefactored, to ensure we
capture any sensitivity.

This change shows an improvement in the case where there'll be
no exporters nor data. If we want to instrument many libraries,
convincing their owners to use OpenCensus will mean that we have
to show little overhead keeping span.End() in there even when
there are no exporters attached.

Results on my non-CPU scaled Macbook Pro.
```shell
$ benchstat before.txt after.txt
name                                                old time/op    new time/op    delta
StartEndSpan_noExporters_neverSample-4                 201ns ± 3%     197ns ± 0%   -1.99%          (p=0.016 n=5+4)
StartEndSpan_noExporters_alwaysSample-4                602ns ± 1%     307ns ± 1%  -49.00%          (p=0.008 n=5+5)
StartEndSpan_withExporters_neverSample-4               200ns ± 1%     198ns ± 1%     ~             (p=0.087 n=5+5)
StartEndSpan_withExporters_alwaysSample-4              658ns ± 1%     669ns ± 0%   +1.64%          (p=0.016 n=5+5)
SpanWithAnnotations_3_noExporters_neverSample-4        261ns ± 1%     260ns ± 2%     ~             (p=0.159 n=5+5)
SpanWithAnnotations_3_noExporters_alwaysSample-4      1.30µs ± 1%    0.66µs ± 1%  -49.44%          (p=0.008 n=5+5)
SpanWithAnnotations_3_withExporters_neverSample-4      263ns ± 0%     267ns ± 8%     ~             (p=0.714 n=4+5)
SpanWithAnnotations_3_withExporters_alwaysSample-4    1.35µs ± 2%    1.36µs ± 1%     ~             (p=0.151 n=5+5)
SpanWithAnnotations_6_noExporters_neverSample-4        335ns ± 1%     331ns ± 3%     ~             (p=0.159 n=4+5)
SpanWithAnnotations_6_noExporters_alwaysSample-4      1.59µs ± 0%    0.80µs ± 1%  -49.48%          (p=0.008 n=5+5)
SpanWithAnnotations_6_withExporters_neverSample-4      332ns ± 1%     329ns ± 1%   -0.90%          (p=0.040 n=5+5)
SpanWithAnnotations_6_withExporters_alwaysSample-4    1.65µs ± 4%    1.67µs ± 1%     ~             (p=0.190 n=5+4)
SpanID_DotString_noExporters_neverSample-4             229ns ± 4%     229ns ± 2%     ~             (p=0.508 n=5+5)
SpanID_DotString_withExporters_neverSample-4           226ns ± 0%     225ns ± 2%     ~             (p=0.540 n=4+5)
SpanID_DotString_noExporters_alwaysSample-4            231ns ± 1%     226ns ± 1%   -2.23%          (p=0.016 n=4+5)
SpanID_DotString_withExporters_alwaysSample-4          231ns ± 2%     227ns ± 1%     ~             (p=0.151 n=5+5)

name                                                old alloc/op   new alloc/op   delta
StartEndSpan_noExporters_neverSample-4                  128B ± 0%      128B ± 0%     ~     (all samples are equal)
StartEndSpan_noExporters_alwaysSample-4                 576B ± 0%      352B ± 0%  -38.89%          (p=0.008 n=5+5)
StartEndSpan_withExporters_neverSample-4                128B ± 0%      128B ± 0%     ~     (all samples are equal)
StartEndSpan_withExporters_alwaysSample-4               576B ± 0%      576B ± 0%     ~     (all samples are equal)
SpanWithAnnotations_3_noExporters_neverSample-4         152B ± 0%      152B ± 0%     ~     (all samples are equal)
SpanWithAnnotations_3_noExporters_alwaysSample-4      1.27kB ± 0%    0.71kB ± 0%  -44.03%          (p=0.008 n=5+5)
SpanWithAnnotations_3_withExporters_neverSample-4       152B ± 0%      152B ± 0%     ~     (all samples are equal)
SpanWithAnnotations_3_withExporters_alwaysSample-4    1.27kB ± 0%    1.27kB ± 0%     ~     (all samples are equal)
SpanWithAnnotations_6_noExporters_neverSample-4         176B ± 0%      176B ± 0%     ~     (all samples are equal)
SpanWithAnnotations_6_noExporters_alwaysSample-4      1.30kB ± 0%    0.74kB ± 0%  -43.21%          (p=0.008 n=5+5)
SpanWithAnnotations_6_withExporters_neverSample-4       176B ± 0%      176B ± 0%     ~     (all samples are equal)
SpanWithAnnotations_6_withExporters_alwaysSample-4    1.30kB ± 0%    1.30kB ± 0%     ~     (all samples are equal)
SpanID_DotString_noExporters_neverSample-4             56.0B ± 0%     56.0B ± 0%     ~     (all samples are equal)
SpanID_DotString_withExporters_neverSample-4           56.0B ± 0%     56.0B ± 0%     ~     (all samples are equal)
SpanID_DotString_noExporters_alwaysSample-4            56.0B ± 0%     56.0B ± 0%     ~     (all samples are equal)
SpanID_DotString_withExporters_alwaysSample-4          56.0B ± 0%     56.0B ± 0%     ~     (all samples are equal)

name                                                old allocs/op  new allocs/op  delta
StartEndSpan_noExporters_neverSample-4                  2.00 ± 0%      2.00 ± 0%     ~     (all samples are equal)
StartEndSpan_noExporters_alwaysSample-4                 4.00 ± 0%      3.00 ± 0%  -25.00%          (p=0.008 n=5+5)
StartEndSpan_withExporters_neverSample-4                2.00 ± 0%      2.00 ± 0%     ~     (all samples are equal)
StartEndSpan_withExporters_alwaysSample-4               4.00 ± 0%      4.00 ± 0%     ~     (all samples are equal)
SpanWithAnnotations_3_noExporters_neverSample-4         4.00 ± 0%      4.00 ± 0%     ~     (all samples are equal)
SpanWithAnnotations_3_noExporters_alwaysSample-4        10.0 ± 0%       7.0 ± 0%  -30.00%          (p=0.008 n=5+5)
SpanWithAnnotations_3_withExporters_neverSample-4       4.00 ± 0%      4.00 ± 0%     ~     (all samples are equal)
SpanWithAnnotations_3_withExporters_alwaysSample-4      10.0 ± 0%      10.0 ± 0%     ~     (all samples are equal)
SpanWithAnnotations_6_noExporters_neverSample-4         6.00 ± 0%      6.00 ± 0%     ~     (all samples are equal)
SpanWithAnnotations_6_noExporters_alwaysSample-4        12.0 ± 0%       9.0 ± 0%  -25.00%          (p=0.008 n=5+5)
SpanWithAnnotations_6_withExporters_neverSample-4       6.00 ± 0%      6.00 ± 0%     ~     (all samples are equal)
SpanWithAnnotations_6_withExporters_alwaysSample-4      12.0 ± 0%      12.0 ± 0%     ~     (all samples are equal)
SpanID_DotString_noExporters_neverSample-4              3.00 ± 0%      3.00 ± 0%     ~     (all samples are equal)
SpanID_DotString_withExporters_neverSample-4            3.00 ± 0%      3.00 ± 0%     ~     (all samples are equal)
SpanID_DotString_noExporters_alwaysSample-4             3.00 ± 0%      3.00 ± 0%     ~     (all samples are equal)
SpanID_DotString_withExporters_alwaysSample-4           3.00 ± 0%      3.00 ± 0%     ~     (all samples are equal)
```

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -56,6 +56,14 @@ func (s *Span) IsRecordingEvents() bool {
 	return s.data != nil
 }
 
+func (s *Span) ensureNotRecordingEvents() {
+	if s != nil {
+		s.mu.Lock()
+		s.data = nil
+		s.mu.Unlock()
+	}
+}
+
 // TraceOptions contains options associated with a trace span.
 type TraceOptions uint32
 
@@ -157,13 +165,15 @@ func NewSpanWithRemoteParent(name string, parent SpanContext, o StartOptions) *S
 
 func startSpanInternal(name string, hasParent bool, parent SpanContext, remoteParent bool, o StartOptions) *Span {
 	span := &Span{}
-	span.spanContext = parent
 
 	cfg := config.Load().(*Config)
 
 	if !hasParent {
 		span.spanContext.TraceID = cfg.IDGenerator.NewTraceID()
+	} else {
+		span.spanContext = parent
 	}
+
 	span.spanContext.SpanID = cfg.IDGenerator.NewSpanID()
 	sampler := cfg.DefaultSampler
 
@@ -215,21 +225,33 @@ func (s *Span) End() {
 	if !s.IsRecordingEvents() {
 		return
 	}
+
+	// If there are no exporters, no need to create spanData
+	// nor invoke all the code below this check, just return.
+	// See Issue 681.
+	if atomic.LoadInt64(&nExporters) <= 0 {
+		return
+	}
+
 	s.exportOnce.Do(func() {
-		// TODO: optimize to avoid this call if sd won't be used.
+		if !s.spanContext.IsSampled() {
+			return
+		}
 		sd := s.makeSpanData()
 		sd.EndTime = internal.MonotonicEndTime(sd.StartTime)
 		if s.spanStore != nil {
 			s.spanStore.finished(s, sd)
 		}
-		if s.spanContext.IsSampled() {
-			// TODO: consider holding exportersMu for less time.
-			exportersMu.Lock()
-			for e := range exporters {
-				e.ExportSpan(sd)
-			}
-			exportersMu.Unlock()
+
+		// TODO: consider holding exportersMu for less time.
+		exportersMu.Lock()
+		for e := range exporters {
+			e.ExportSpan(sd)
 		}
+		exportersMu.Unlock()
+
+		// Discard any data
+		s.ensureNotRecordingEvents()
 	})
 }
 
--- a/trace/benchmark_test.go
+++ b/trace/benchmark_test.go
@@ -19,87 +19,304 @@ import (
 	"testing"
 )
 
-func BenchmarkStartEndSpan(b *testing.B) {
-	traceBenchmark(b, func(b *testing.B) {
-		ctx := context.Background()
-		b.ResetTimer()
-		for i := 0; i < b.N; i++ {
-			_, span := StartSpan(ctx, "/foo")
-			span.End()
-		}
-	})
+func BenchmarkStartEndSpan_noExporters_neverSample(b *testing.B) {
+	ApplyConfig(Config{DefaultSampler: NeverSample()})
+	ctx := context.Background()
+
+	b.ResetTimer()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		_, span := StartSpan(ctx, "/foo")
+		span.End()
+	}
 }
 
-func BenchmarkSpanWithAnnotations_3(b *testing.B) {
-	traceBenchmark(b, func(b *testing.B) {
-		ctx := context.Background()
-		b.ResetTimer()
+func BenchmarkStartEndSpan_noExporters_alwaysSample(b *testing.B) {
+	ApplyConfig(Config{DefaultSampler: AlwaysSample()})
+	ctx := context.Background()
 
-		for i := 0; i < b.N; i++ {
-			_, span := StartSpan(ctx, "/foo")
-			span.AddAttributes(
-				BoolAttribute("key1", false),
-				StringAttribute("key2", "hello"),
-				Int64Attribute("key3", 123),
-			)
-			span.End()
-		}
-	})
+	b.ResetTimer()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		_, span := StartSpan(ctx, "/foo")
+		span.End()
+	}
 }
 
-func BenchmarkSpanWithAnnotations_6(b *testing.B) {
-	traceBenchmark(b, func(b *testing.B) {
-		ctx := context.Background()
-		b.ResetTimer()
+type anExporter int
 
-		for i := 0; i < b.N; i++ {
-			_, span := StartSpan(ctx, "/foo")
-			span.AddAttributes(
-				BoolAttribute("key1", false),
-				BoolAttribute("key2", true),
-				StringAttribute("key3", "hello"),
-				StringAttribute("key4", "hello"),
-				Int64Attribute("key5", 123),
-				Int64Attribute("key6", 456),
-			)
-			span.End()
-		}
-	})
+func (ae *anExporter) ExportSpan(s *SpanData) {}
+
+var _ Exporter = (*anExporter)(nil)
+
+func BenchmarkStartEndSpan_withExporters_neverSample(b *testing.B) {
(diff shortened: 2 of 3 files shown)


trace/trace.go
```diff
@@ -215,21 +225,33 @@ func (s *Span) End() {
 	if !s.IsRecordingEvents() {
 		return
 	}
+
+	// If there are no exporters, no need to create spanData
+	// nor invoke all the code below this check, just return.
+	// See Issue 681.
+	if atomic.LoadInt64(&nExporters) <= 0 {
```
ramonza: Why not do this up-front when starting the span instead - avoid ever creating the SpanData object in the first place?

trace/trace.go
```diff
@@ -215,21 +225,33 @@ func (s *Span) End() {
...
+		return
+	}
+
 	s.exportOnce.Do(func() {
-		// TODO: optimize to avoid this call if sd won't be used.
+		if !s.spanContext.IsSampled() {
```
ramonza: This breaks zpages (as well as the prior early return outside of the Do). You need to ensure that s.spanStore.finish gets called.

trace/trace.go
```diff
@@ -157,13 +165,15 @@ func NewSpanWithRemoteParent(name string, parent SpanContext, o StartOptions) *S
 
 func startSpanInternal(name string, hasParent bool, parent SpanContext, remoteParent bool, o StartOptions) *Span {
 	span := &Span{}
-	span.spanContext = parent
 
 	cfg := config.Load().(*Config)
 
 	if !hasParent {
```
ramonza: Since you added an else clause I think it would be clearer to invert this condition.

trace/trace.go
```diff
@@ -56,6 +56,14 @@ func (s *Span) IsRecordingEvents() bool {
 	return s.data != nil
 }
 
+func (s *Span) ensureNotRecordingEvents() {
```
ramonza: I wonder if this could be merged with a version of makeSpanData to "steal" the SpanData instead of making a copy. AFAICT makeSpanData is only called in End in order to prevent concurrent modification through the Span.data pointer, but if we nil that pointer we don't need to worry about it any more and maybe there is no longer a need to make a deep copy?

odeke-em:
Sorry for the late reply, I was working on other things urgently.

trace/trace.go
odeke-em: Yeah I was thinking of that but didn't want to make a very colorful diff. I'll fix it. Thanks.

trace/trace.go
odeke-em: Oh, I see, thanks for pointing that out.

ramonza:
@odeke-em any updates on this?

Issue #683, opened 2018-04-06
exporter/stackdriver: regression in Int64 data exporting with ochttp.DefaultServerViews


odeke-em:
I was just running an integration example after pulling the latest commits from master and I've been getting lots of errors on Stackdriver. The small repro below should show the problem

```go
package main

import (
	"io/ioutil"
	"log"
	"net/http"
	"net/http/httptest"
	"os"
	"time"

	"go.opencensus.io/exporter/stackdriver"
	"go.opencensus.io/plugin/ochttp"
	"go.opencensus.io/stats/view"
)

func main() {
	se, err := stackdriver.NewExporter(stackdriver.Options{
		ProjectID: os.Getenv("TEST_PROJECT_ID"),
	})
	if err != nil {
		log.Fatalf("Failed to create Stackdriver exporter: %v", err)
	}
	view.RegisterExporter(se)

	if err := view.Register(ochttp.DefaultServerViews...); err != nil {
		log.Fatalf("Failed to register default server views: %v", err)
	}

	h := &ochttp.Handler{
		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			w.Write([]byte("Pong"))
		}),
	}
	cst := httptest.NewServer(h)
	defer cst.Close()
	view.SetReportingPeriod(10 * time.Millisecond)

	for i := 0; i < 10000; i++ {
		res, _ := http.Get(cst.URL)
		if res != nil && res.Body != nil {
			_, _ = ioutil.ReadAll(res.Body)
			_ = res.Body.Close()
		}
		<-time.After(250 * time.Millisecond)
	}
}
```

```shell
$ GOOGLE_APPLICATION_CREDENTIALS=~/Downloads/census-demos-237a8e1e41df.json TEST_PROJECT_ID=census-demos go run main.go 
2018/04/06 00:41:21 Failed to export to Stackdriver Monitoring: stackdriver metric descriptor was not created as int64
2018/04/06 00:41:21 Failed to export to Stackdriver Monitoring: stackdriver metric descriptor was not created as int64
2018/04/06 00:41:21 Failed to export to Stackdriver Monitoring: stackdriver metric descriptor was not created as int64
2018/04/06 00:41:21 Failed to export to Stackdriver Monitoring: stackdriver metric descriptor was not created as int64
2018/04/06 00:41:21 Failed to export to Stackdriver Monitoring: stackdriver metric descriptor was not created as int64
[...]

PR #685, opened 2018-04-06
Fix equalMeasureAggTagKeys for Stackdriver exporter.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-go/issues/683.

Count aggregation will always be exported as INT64 MetricDescriptor, no
matter whether Measure is INT64 or FLOAT64.


exporter/stackdriver/stats_test.go
```diff
@@ -424,7 +424,7 @@ func TestEqualAggWindowTagKeys(t *testing.T) {
 		wantErr bool
 	}{
 		{
-			name: "count agg",
+			name: "count agg with in64 meausre",
```
odeke-em: s/meausre/measure/g

songy23: Good catch, fixed.

PR #688, opened 2018-04-06
Allow ignoring certain requests.

bweston92:
Adds the ability to ignore certain requests without having to do to much logic on the router/middleware.

Main uses would be the same as set in the example "GET /healthz" or "OPTIONS *" requests.

--
Will add tests if the feature deems it self useful.

### DIFF
--- a/plugin/ochttp/server.go
+++ b/plugin/ochttp/server.go
@@ -30,6 +30,25 @@ import (
 	"go.opencensus.io/trace/propagation"
 )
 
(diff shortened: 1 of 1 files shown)


rakyll:
You don't need to patch OpenCensus to achieve this. You need to implement a Handler that has the same logic to handle /healthz and OPTIONS requests by using a non-OC handler and forward all the other traffic to the OpenCensus handler.

bweston92:
That’s totally fine and is how I am doing it currently just a bit bloated, especially when they might ignore based on headers etc.

Issue #687, opened 2018-04-06
Update gRPC client and server metrics based on Specs.


songy23:
New gRPC metrics specs: https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md.

Similar to C++ (https://github.com/census-instrumentation/opencensus-cpp/pull/138) and Java (https://github.com/census-instrumentation/opencensus-java/pull/1115).

PR #689, opened 2018-04-08
Update gRPC measures and views from spec

ramonza:
Fixes: #687


plugin/ocgrpc/client_metrics.go
```diff
@@ -23,78 +23,82 @@ import (
...
 var (
-	ClientErrorCount       = stats.Int64("grpc.io/client/error_count", "RPC Errors", stats.UnitNone)
-	ClientRequestBytes     = stats.Int64("grpc.io/client/request_bytes", "Request bytes", stats.UnitBytes)
-	ClientResponseBytes    = stats.Int64("grpc.io/client/response_bytes", "Response bytes", stats.UnitBytes)
-	ClientStartedCount     = stats.Int64("grpc.io/client/started_count", "Number of client RPCs (streams) started", stats.UnitNone)
-	ClientFinishedCount    = stats.Int64("grpc.io/client/finished_count", "Number of client RPCs (streams) finished", stats.UnitNone)
-	ClientRequestCount     = stats.Int64("grpc.io/client/request_count", "Number of client RPC request messages", stats.UnitNone)
-	ClientResponseCount    = stats.Int64("grpc.io/client/response_count", "Number of client RPC response messages", stats.UnitNone)
-	ClientRoundTripLatency = stats.Float64("grpc.io/client/roundtrip_latency", "RPC roundtrip latency in msecs", stats.UnitMilliseconds)
+	ClientSentMessagesPerRPC     = stats.Int64("grpc.io/client/sent_messages_per_rpc", "Number of messages sent in the RPC (always 1 for non-streaming RPCs).", stats.UnitNone)
```
odeke-em: This unit is wrong, please update it to "1" as the I believe the spec says in https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md#units
screen shot 2018-04-09 at 10 53 50 am

plugin/ocgrpc/client_metrics.go
```diff
@@ -23,78 +23,82 @@ import (
...
+	ClientSentBytesPerRPC        = stats.Int64("grpc.io/client/sent_bytes_per_rpc", "Total bytes sent across all request messages per RPC.", stats.UnitBytes)
+	ClientReceivedMessagesPerRPC = stats.Int64("grpc.io/client/received_messages_per_rpc", "Number of response messages received per RPC (always 1 for non-streaming RPCs).", stats.UnitNone)
```
odeke-em: Ditto about using "1" as the spec requires.

plugin/ocgrpc/client_metrics.go
```diff
@@ -23,78 +23,82 @@ import (
...
-		Name:        "grpc.io/client/error_count",
-		Description: "RPC Errors",
-		TagKeys:     []tag.Key{KeyStatus, KeyMethod},
-		Measure:     ClientErrorCount,
-		Aggregation: view.Count(),
+	ClientSentBytesPerRPCView = &view.View{
+		Measure:     ClientSentBytesPerRPC,
+		Name:        "grpc.io/client/sent_bytes_per_rpc",
+		Description: "Distribution of bytes sent per RPC, by method.",
+		TagKeys:     []tag.Key{KeyClientMethod},
```
odeke-em: The spec also says that
screen shot 2018-04-09 at 11 09 18 am

Thus we also need "KeyStatus" as a TagKey.

plugin/ocgrpc/client_metrics.go
odeke-em: Oh, I guess I hadn't seen the definition of "stats.UnitNone" which is a dimensionless unit, but it was confusing initially, "UnitNone" to me would be "" instead of "1" whereas perhaps we could use "stats.UnitDimensionless"

plugin/ocgrpc/stats_common.go
```diff
@@ -46,32 +52,148 @@ type rpcData struct {
...
+	}
+}
+
+func statusCodeToString(s *status.Status) string {
+	// see https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+	switch c := s.Code(); c {
+	case codes.OK:
+		return "OK"
+	case codes.Canceled:
+		return "CANCELLED"
```
odeke-em: Aha, it's funny we have "Canceled" as the enumeration but "CANCELLED" as the string as defined in https://godoc.org/google.golang.org/grpc/codes vs https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

stats/record.go
```diff
@@ -17,12 +17,17 @@ package stats
...
+	"os"
 
 	"go.opencensus.io/stats/internal"
 	"go.opencensus.io/tag"
 )
 
+var debug bool
+
 func init() {
+	debug = os.Getenv("OC_STATS_DEBUG") != ""
```
odeke-em: Let's document this perhaps?

zpages/rpcz.go
```diff
@@ -258,39 +260,55 @@ func (s snapExporter) ExportView(vd *view.Data) {
...
-		case ocgrpc.ClientErrorCountView:
-			s.ErrorsTotal = int(count)
+		case ocgrpc.ClientCompletedRPCsView:
+			if _, ok := haveResetErrors[method]; ok {
+				haveResetErrors[method] = struct{}{}
+				s.ErrorsTotal = 0
+			}
+			for _, tag := range row.Tags {
+				if tag.Key == ocgrpc.KeyClientStatus && tag.Value != "OK" {
+					s.ErrorsTotal += int(count)
```
odeke-em: Although we removed the measure and view for viewing the total number of gRPC errors.

odeke-em:
Thank you for this PR @ramonza, it's great work, I like the rigor with the spec adherence. It is quite a large PR though, so I'll need to do a couple of takes on reviewing it.

zpages/rpcz.go
ramonza: Yes, so this is why I had to compute it from the CompletedRPCs measures

stats/record.go
ramonza: Don't think it should be a documented feature of the library so I removed it instead.

examples/grpc/helloworld_client/main.go
```diff
@@ -37,8 +37,8 @@ func main() {
 	// the collected data.
 	view.RegisterExporter(&exporter.PrintExporter{})
 
-	// Register the view to collect client request count.
-	if err := view.Register(ocgrpc.ClientErrorCountView); err != nil {
```
odeke-em: I have a naive question, how are now going to be able to track the number of errors in either the client or server?

zpages/rpcz.go
odeke-em: I have a naive question: how are we then going to view and track the number of errors?

examples/grpc/helloworld_client/main.go
ramonza: We now break down the requests by status, so you can get the total errors by adding up all the non-success statuses like we do in rpcz.

plugin/ocgrpc/stats_common.go
rakyll: The enum type stringifies them differently btw in gRPC Go: https://github.com/grpc/grpc-go/blob/master/codes/code_string.go#L23.

odeke-em: Baby Jesus, perhaps let's follow what grpc-go is doing? Thanks for posting that @rakyll.

ramonza: Go codes look like just the names of the vars, highly unlikely to be consistent across languages. The examples in the OpenCensus spec are consistent with the gRPC doc (all upper case, CANCELLED with two "l", UPPER_SNAKE_CASE) which is the most official source I could find. Also explicitly added to the spec to avoid future confusion: https://github.com/census-instrumentation/opencensus-specs/pull/97

ramonza:
@odeke-em @rakyll PTAL

plugin/ocgrpc/stats_common.go
```diff
@@ -46,32 +52,148 @@ type rpcData struct {
...
+			ServerSentMessagesPerRPC.M(atomic.LoadInt64(&d.sentCount)),
+			ServerReceivedMessagesPerRPC.M(atomic.LoadInt64(&d.recvCount)),
+			ServerReceivedBytesPerRPC.M(atomic.LoadInt64(&d.recvBytes)),
+			ServerLatency.M(latencyMillis))
+	}
+}
+
+func statusCodeToString(s *status.Status) string {
+	// see https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+	switch c := s.Code(); c {
```
bogdandrutu: Usually switch statements are very slow. Probably better to define a static map<Code, String>

ramonza: Curious what the intuition behind that is? Did a benchmark and turns out that the switch is actually significantly faster, at least on my laptop:
```
name                                  time/op
StatusCodeToString_OK-8               2.66ns ± 0%
StatusCodeToString_Unauthenticated-8  2.88ns ± 0%
MapAlternativeImpl_OK-8               7.56ns ± 1%
```

bogdandrutu: Then keep the switch statement. the map search seems to me very slow, maybe is a Go thing with maps. The way how the switch statements are now implemented is using a 2-level jump table (https://www.codeproject.com/Articles/100473/Something-You-May-Not-Know-About-the-Switch-Statem) so it should be one extra jump, but probably the way how the map is implemented in Go is also 2-level jump (does not store the values directly in the map, instead it uses pointers).

plugin/ocgrpc/benchmark_test.go
```diff
@@ -0,0 +1,52 @@
...
+	codes.Internal:           "INTERNAL",
+	codes.Unavailable:        "UNAVAILABLE",
+	codes.DataLoss:           "DATA_LOSS",
+	codes.Unauthenticated:    "UNAUTHENTICATED",
+}
+
+func BenchmarkMapAlternativeImpl_OK(b *testing.B) {
+	st := status.New(codes.OK, "OK")
+	for i := 0; i < b.N; i++ {
+		s := codeToStringMap[st.Code()]
```
rakyll: codeToStringMap[st.Code()]

PR #696, opened 2018-04-11
Proposal: StartSpan with functional options

ramonza:


trace/trace.go
```diff
@@ -123,6 +123,41 @@ type StartOptions struct {
 	// SpanKind represents the kind of a span. If none is set,
 	// SpanKindUnspecified is used.
 	SpanKind int
+
+	parent       *SpanContext
+	parentRemote bool
+}
+
+type StartSpanOption func(*StartOptions)
```
rakyll: I'd s/StartSpanOption/StartOption

plugin/ochttp/server.go
```diff
@@ -76,32 +76,30 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if hasPropagated && !h.IsPublicEndpoint {
+		opts = append(opts, trace.WithRemoteParent(sc))
+	}
+	ctx, span = trace.StartSpan(ctx, name, opts...)
+	if hasPropagated && h.IsPublicEndpoint {
+		span.AddLink(trace.Link{
+			TraceID:    sc.TraceID,
+			SpanID:     sc.SpanID,
+			Type:       trace.LinkTypeChild,
+			Attributes: nil,
```
rakyll: Can you delete this line while you are there?

trace/trace.go
```diff
@@ -123,6 +123,41 @@ type StartOptions struct {
...
+}
+
+func WithRemoteParent(parent SpanContext) StartSpanOption {
+	return func(o *StartOptions) {
+		o.parent = &parent
+		o.parentRemote = true
+	}
+}
+
+func WithParent(span *Span) StartSpanOption {
```
rakyll: I don't think we should ever do this. Parent should always determined from the incoming context if it is not remote.

ramonza: This was mostly added to *avoid* taking the parent from the context e.g. in ocgrpc, ochttp we take a Context but we want to ignore the Span in that context since we are starting a new trace. I will change it to something like "AsRootSpan()" which just sets parent to nil.

rakyll: Ah I see. I actually would prefer us providing an option that makes the span root instead.

ramonza: actually realized I can just explicitly clear it from the context

rakyll:
Can I keep working on this? I need this change earlier this week.

ramonza:
Yeah I will be flying now for the next million years 😞 (until Tuesday
afternoon)

rakyll:
No worries, safe flights :)

PR #699, opened 2018-04-12
Allow Stackdriver users to pass monitoring and tracing options separately

### LINKED ISSUE #691: exporter/stackdriver: separate ClientOptions for monitoring and trace
I would like to pass OpenCensus the gRPC connections directly for Monitoring and Trace, but the current API uses the same ClientOptions list for both of them, which doesn't let me pass in `option.WithGRPCConn`.

rakyll:
Fixes #691.


odeke-em:
Nice, thanks @rakyll!

PR #701, opened 2018-04-12
Make plugin/ochttp/server set its span status on ending

### LINKED ISSUE #700: plugin/ochttp: http.Handler doesn't set its span's Status, derived from the statusCode
Reviewing the code in https://github.com/census-instrumentation/opencensus-go/blob/02b14064ad5f94ef466ddacdc1a90f03ae7e0cfa/plugin/ochttp/server.go I do not see the status code being recorded in the trace, only in stats.

odeke-em:
Fixes #700

The server plugin should set the status on end, after roundtripping.
Also while here ensure that helper function status converts
http.StatusOK i.e. 200 to codeOK i.e. 0

### DIFF
--- a/plugin/ochttp/server_test.go
+++ b/plugin/ochttp/server_test.go
@@ -3,13 +3,16 @@ package ochttp
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"crypto/tls"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync"
 	"testing"
 
 	"golang.org/x/net/http2"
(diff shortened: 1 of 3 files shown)


odeke-em:
Examining the results from a live server for which I requested for a non-existent resource /foo

### Before
screen shot 2018-04-12 at 1 45 01 pm

### After
screen shot 2018-04-12 at 1 45 11 pm

odeke-em:
With this change, now AWS X-Ray also displays the proper error

screen shot 2018-04-12 at 2 53 53 pm
screen shot 2018-04-12 at 2 53 41 pm

plugin/ochttp/server_test.go
```diff
@@ -273,3 +276,88 @@ func TestHandlerProxiesHijack_HTTP2(t *testing.T) {
 		t.Errorf("got = %q; want = %q", g, w)
 	}
 }
+
+func TestEnsureTrackingResponseWriterSetsStatusCode(t *testing.T) {
+	// Ensure that the trackingResponseWriter always sets the spanStatus on ending the span.
+	// Because we can only examine the Status after exporting, this test roundtrips a
+	// couple of requests and then later examines the exported spans.
+	// See Issue #700.
+	spanInspector := new(spanExporter)
```
rakyll: Call this exporter. It is an exporter.

plugin/ochttp/server_test.go
```diff
@@ -273,3 +276,88 @@ func TestHandlerProxiesHijack_HTTP2(t *testing.T) {
...
+
+var _ trace.Exporter = (*spanExporter)(nil)
+
+func (se *spanExporter) ExportSpan(sd *trace.SpanData) {
+	se.Lock()
+	se.store = append(se.store, sd)
+	se.Unlock()
+}
+
+type responseBase struct {
```
rakyll: I'd name these with a test prefix.

type testTransport struct

plugin/ochttp/server_test.go
```diff
@@ -273,3 +276,88 @@ func TestHandlerProxiesHijack_HTTP2(t *testing.T) {
...
+		want trace.Status
+	}{
+		{res: &http.Response{StatusCode: 200}, want: trace.Status{0, `"OK"`}},
+		{res: &http.Response{StatusCode: 500}, want: trace.Status{2, `"UNKNOWN"`}},
+		{res: &http.Response{StatusCode: 403}, want: trace.Status{7, `"PERMISSION_DENIED"`}},
+		{res: &http.Response{StatusCode: 401}, want: trace.Status{16, `"UNAUTHENTICATED"`}},
+		{res: &http.Response{StatusCode: 429}, want: trace.Status{8, `"RESOURCE_EXHAUSTED"`}},
+	}
+
+	for i, tt := range tests {
```
rakyll: Name these tests instead and create subtests.

plugin/ochttp/server_test.go
```diff
@@ -273,3 +276,88 @@ func TestHandlerProxiesHijack_HTTP2(t *testing.T) {
...
+		got, want := spanInspector.store[i].Status, tt.want
+		if got != want {
+			t.Errorf("SpanData: #%d:\ngot =  (%#v)\nwant = (%#v)", i, got, want)
+		}
+	}
+}
+
+type spanExporter struct {
+	sync.Mutex
+	t     *testing.T
```
rakyll: Remove t.

plugin/ochttp/server_test.go
odeke-em: Roger that, thanks.

plugin/ochttp/server_test.go
odeke-em: Roger that, thanks.

plugin/ochttp/server_test.go
odeke-em: Oops, removing, that was a vestige from my previous tests, thanks.

plugin/ochttp/server_test.go
odeke-em: I don't follow, could you please help me with more wording/guidance on what you mean?

rakyll: I was suggesting to create subtests such as https://github.com/rakyll/opencensus-go/blob/master/plugin/ochttp/propagation/tracecontext/propagation_test.go#L74. It makes it possible to run each test individually and generates a pretty output.

rakyll:
go vet your code please.

odeke-em:
Lol, I did with `go test` but is it the change about using unkeyed fields? I just didn't want the code to be extra verbose as that struct is going to have 2 fields but otherwise sure, I'll update that.

odeke-em:
@rakyll PTAL.

plugin/ochttp/server_test.go
```diff
@@ -273,3 +276,78 @@ func TestHandlerProxiesHijack_HTTP2(t *testing.T) {
...
+			go func() {
+				pwc.Write([]byte("Foo"))
+				pwc.Close()
+			}()
+			inRes := tt.res
+			inRes.Body = prc
+			tr := &traceTransport{base: &testResponseTransport{res: inRes}}
+			req, err := http.NewRequest("POST", "https://example.org", bytes.NewReader([]byte("testing")))
+			if err != nil {
+				t.Fatalf("#%d: NewRequest error: %v", i, err)
```
rakyll: t.Fatalf("NewRequest error: %v", err)

Don't print i anymore. The messages will be outputted under the test case name.

odeke-em: For shizzle, thanks.

odeke-em:
I believe all is addressed now.

PR #703, opened 2018-04-13
Add explicit library dependencies

### LINKED ISSUE #684: Publish explicit dependency version requirements
Using OpenCensus for Go with gRPC depends on recent versions of gRPC-Go. However, this dependency is never made explicit anywhere (such as a `Gopkg.toml` or `go.mod` file). I understand this is a bit in flux at the moment with the vgo proposal, but I think we should discuss how we make these version dependencies explicit, so I'm opening this issue for the purposes of that discussion.

Personally, I would welcome simply defining a `Gopkg.toml` and generate a `Gopkg.lock` and check those in right now. It would help users of `golang/dep` to get started with OpenCensus, and we could also include a fix for  https://github.com/census-instrumentation/opencensus-go/issues/675 in the dependency declaration.

Later on, as vgo is merged into the standard tool, we can move to defining a `go.mod` file.

johanbrandhorst:
Fixes #684.

Adds Gopkg.toml and Gopkg.lock with initial dependency versions. Includes workaround for Apache Thrift repo migration.

The versions used were chosen by `dep init`, and I welcome a discussion about exactly which direct dependency versions we want to use. The currently chosen versions allow all packages to build successfully so are a good starting point.


johanbrandhorst:
We'll probably want to discuss some procedure for updating dependencies as well, for example gRPC has a (afaik) 3 monthly release cycle and we'll certainly want to stick to that. This may already be in place.

Gopkg.toml
```diff
@@ -0,0 +1,44 @@
...
+  branch = "master"
+  name = "google.golang.org/api"
+
+[[constraint]]
+  branch = "master"
+  name = "google.golang.org/genproto"
+
+[[constraint]]
+  name = "google.golang.org/grpc"
+  version = "1.11.3"
```
ramonza: I think we should default to specifying major version ranges for everything, so: version = "^1.11.3" here

johanbrandhorst: This is equivalent to `^1.11.3` https://golang.github.io/dep/docs/Gopkg.toml.html#version.

ramonza:
I think we need to be able to test this. What do you think of the following changes?:

- add vendor/ to .gitignore
- modify one of the Travis CI configurations to do "dep ensure" before building

That will make sure we are testing with the versions specified here but avoid vendoring the code.

johanbrandhorst:
Please re-review with the changes suggested.

bweston92:
When will this be in a release?

ramonza:
Let's do a release after #728

PR #704, opened 2018-04-15
Fix example. Do not overwrite new request.

cooldarkdryplace:
Instead of overwriting new request, use only context from original request.

### DIFF
--- a/examples/http/helloworld_server/main.go
+++ b/examples/http/helloworld_server/main.go
@@ -50,7 +50,7 @@ func main() {
 		r, _ := http.NewRequest("GET", "https://example.com", nil)
 
 		// Propagate the trace header info in the outgoing requests.
-		r.WithContext(req.Context())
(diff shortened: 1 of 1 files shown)


examples/http/helloworld_server/main.go
```diff
@@ -50,7 +50,7 @@ func main() {
 		r, _ := http.NewRequest("GET", "https://example.com", nil)
 
 		// Propagate the trace header info in the outgoing requests.
-		r = req.WithContext(req.Context())
```
ramonza: There definitely looks to be a problem here but I don't think your fix is quite right. `WithContext` doesn't mutate the receiver, it returns a copy of it with the supplied context set. I think this should be:

```go
r = r.WithContext(req.Context())
```

cooldarkdryplace: Thanks, I have updated PR.

PR #708, opened 2018-04-16
Deprecate NewSpanXXX in the favor of StartSpan

### LINKED ISSUE #694: Make all span starting APIs take a context
Go runtime/trace package is soon going to support user events. See http://tip.golang.org/pkg/runtime/trace.

In order to provide execution tracer tasks and spans from the OpenCensus library, we should make sure we are not dropping the parent runtime/trace tasks and spans.

Hence I am suggesting us to remove NewSpanXXX functions and replace them with:

```
func StartSpanWithOptions(ctx context.Contex, name string, o StartOptions) (context.Context, *Span)
func StartSpanWithRemoteParent(ctx context.Contex, name string, parent SpanContext o StartOptions) (context.Context, *Span)
```

rakyll:
Required to support the Go execution tracer spans.

Fixes #694.

### DIFF
--- a/exporter/stackdriver/trace.go
+++ b/exporter/stackdriver/trace.go
@@ -117,11 +117,14 @@ func (e *traceExporter) uploadSpans(spans []*trace.SpanData) {
 		req.Spans = append(req.Spans, protoFromSpanData(span, e.projectID))
 	}
 	// Create a never-sampled span to prevent traces associated with exporter.
-	span := trace.NewSpan("go.opencensus.io/exporter/stackdriver.uploadSpans", nil, trace.StartOptions{Sampler: trace.NeverSample()})
+	ctx, span := trace.StartSpan( // TODO: add timeouts
+		context.Background(),
+		"go.opencensus.io/exporter/stackdriver.uploadSpans",
+		trace.WithSampler(trace.NeverSample()),
+	)
 	defer span.End()
 	span.AddAttributes(trace.Int64Attribute("num_spans", int64(len(spans))))
 
-	ctx := trace.WithSpan(context.Background(), span) // TODO: add timeouts
 	err := e.client.BatchWriteSpans(ctx, &req)
 	if err != nil {
 		span.SetStatus(trace.Status{Code: 2, Message: err.Error()})
--- a/exporter/stackdriver/trace_test.go
+++ b/exporter/stackdriver/trace_test.go
@@ -35,9 +35,8 @@ func TestBundling(t *testing.T) {
 	}
 	trace.RegisterExporter(exporter)
 
-	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
 	for i := 0; i < 35; i++ {
-		_, span := trace.StartSpan(context.Background(), "span")
+		_, span := trace.StartSpan(context.Background(), "span", trace.WithSampler(trace.AlwaysSample()))
 		span.End()
 	}
 
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -97,7 +97,14 @@ func FromContext(ctx context.Context) *Span {
 }
 
 // WithSpan returns a new context with the given Span attached.
+//
+// Deprecated: Use NewContext.
 func WithSpan(parent context.Context, s *Span) context.Context {
+	return NewContext(parent, s)
+}
+
+// NewContext returns a new context with the given Span attached.
+func NewContext(parent context.Context, s *Span) context.Context {
 	return context.WithValue(parent, contextKey{}, s)
 }
 
@@ -146,6 +153,12 @@ func WithSampler(sampler Sampler) StartOption {
 }
 
 // WithRemoteParent makes new spans to be the child of the given part.
+//
+// WithRemoteParent is used to set a parent that has been
+// propagated in an incoming request or RPC.
+//
+// If there is already a parent in the incoming context.Context,
+// this option replaces it with the given parent.
 func WithRemoteParent(parent SpanContext) StartOption {
 	return func(o *StartOptions) {
 		o.remoteParent = parent
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -123,34 +123,76 @@ type StartOptions struct {
 	// SpanKind represents the kind of a span. If none is set,
 	// SpanKindUnspecified is used.
 	SpanKind int
+
+	remoteParent SpanContext
+}
+
+// StartOption apply changes to StartOptions.
+type StartOption func(*StartOptions)
+
+// WithSpanKind makes new spans to be created with the given kind.
+func WithSpanKind(spanKind int) StartOption {
+	return func(o *StartOptions) {
+		o.SpanKind = spanKind
+	}
+}
+
+// WithSampler makes new spans to be be created with a custom sampler.
+// Otherwise, the global sampler is used.
+func WithSampler(sampler Sampler) StartOption {
+	return func(o *StartOptions) {
+		o.Sampler = sampler
+	}
+}
+
+// WithRemoteParent makes new spans to be the child of the given part.
+func WithRemoteParent(parent SpanContext) StartOption {
+	return func(o *StartOptions) {
+		o.remoteParent = parent
+	}
 }
 
 // StartSpan starts a new child span of the current span in the context. If
 // there is no span in the context, creates a new trace and span.
 //
-// This is provided as a convenience for WithSpan(ctx, NewSpan(...)). Use it
-// if you require custom spans in addition to the default spans provided by
-// ocgrpc, ochttp or similar framework integration.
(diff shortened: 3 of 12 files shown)


exporter/stackdriver/trace.go
```diff
@@ -117,11 +117,14 @@ func (e *traceExporter) uploadSpans(spans []*trace.SpanData) {
 		req.Spans = append(req.Spans, protoFromSpanData(span, e.projectID))
 	}
 	// Create a never-sampled span to prevent traces associated with exporter.
-	span := trace.NewSpan("go.opencensus.io/exporter/stackdriver.uploadSpans", nil, trace.StartOptions{Sampler: trace.NeverSample()})
+	ctx, span := trace.StartSpan( // TODO: add timeouts
```
odeke-em: @rakyll "timeouts" for what?

exporter/stackdriver/trace_test.go
```diff
@@ -35,9 +35,8 @@ func TestBundling(t *testing.T) {
 	}
 	trace.RegisterExporter(exporter)
 
-	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
 	for i := 0; i < 35; i++ {
-		_, span := trace.StartSpan(context.Background(), "span")
+		_, span := trace.StartSpan(context.Background(), "span", trace.WithSampler(trace.AlwaysSample()))
```
odeke-em: Why not leave it as it was before?

trace/trace.go
```diff
@@ -123,34 +123,76 @@ type StartOptions struct {
...
 //
-// This is provided as a convenience for WithSpan(ctx, NewSpan(...)). Use it
-// if you require custom spans in addition to the default spans provided by
-// ocgrpc, ochttp or similar framework integration.
-func StartSpan(ctx context.Context, name string) (context.Context, *Span) {
-	parentSpan, _ := ctx.Value(contextKey{}).(*Span)
-	span := NewSpan(name, parentSpan, StartOptions{})
+// If WithRemoteParent option is provided, started span will be a child
+// of the given remote parent.
+func StartSpan(ctx context.Context, name string, o ...StartOption) (context.Context, *Span) {
```
odeke-em: Hand clap, this is a great change! It takes away the previous non-ideal usage of:
1. trace.NewSpan
2. trace.WithSpan to get the context

trace/trace.go
```diff
@@ -123,34 +123,76 @@ type StartOptions struct {
...
+	var parent SpanContext
+	if p := FromContext(ctx); p != nil {
+		parent = p.spanContext
+	}
+	for _, op := range o {
+		op(&opts)
+	}
+
+	var span *Span
+	if (opts.remoteParent != SpanContext{}) {
```
odeke-em: Umm, please remove the brackets.

trace/trace.go
```diff
@@ -123,34 +123,76 @@ type StartOptions struct {
...
+		span = startSpanInternal(name, opts.remoteParent != SpanContext{}, opts.remoteParent, true, opts)
```
odeke-em: We've already evaluated this condition `opts.remoteParent != SpanContext{}` so perhaps let's improve it with
```go
if hasRemoteParent := opts.remoteParent != SpanContext{}; hasRemoteParent {
    span = startSpanInternal(name, hasRemoteParent, opts.remoteParent, true, opts)
} else {
    ...
}
```

trace/trace.go
rakyll: Doesn't compile without the brackets.

trace/trace.go
rakyll: hasRemoteParent would always be true. Fixed this.

trace/trace.go
rakyll: Done.

trace/trace.go
```diff
@@ -123,34 +123,76 @@ type StartOptions struct {
...
+// Deprecated: Use StartSpan.
 func NewSpan(name string, parent *Span, o StartOptions) *Span {
-	hasParent := false
 	var parentSpanContext SpanContext
 	if parent != nil {
-		hasParent = true
 		parentSpanContext = parent.SpanContext()
 	}
-	return startSpanInternal(name, hasParent, parentSpanContext, false, o)
+	return startSpanInternal(name, true, parentSpanContext, false, o)
```
odeke-em: Wait, why has this changed? If parent is nil, why should `hasParent` or `remoteParent` be `true`?

trace/trace.go
```diff
@@ -123,34 +123,76 @@ type StartOptions struct {
...
+	if p := FromContext(ctx); p != nil {
+		parent = p.spanContext
+	}
+	for _, op := range o {
+		op(&opts)
+	}
+
+	var span *Span
+	if opts.remoteParent != (SpanContext{}) {
+		span = startSpanInternal(name, true, opts.remoteParent, true, opts)
```
odeke-em: Am not sure if this is the right change, not every span has a remote parent.

odeke-em: Rather, not every span has a parent

rakyll: If opts.remoteParent is non-zero, the span has a parent.

trace/trace.go
rakyll: Oops, fixed.

rakyll:
PTAL

trace/trace.go
```diff
@@ -123,34 +123,76 @@ type StartOptions struct {
...
+// WithSampler makes new spans to be be created with a custom sampler.
+// Otherwise, the global sampler is used.
+func WithSampler(sampler Sampler) StartOption {
+	return func(o *StartOptions) {
+		o.Sampler = sampler
+	}
+}
+
+// WithRemoteParent makes new spans to be the child of the given part.
+func WithRemoteParent(parent SpanContext) StartOption {
```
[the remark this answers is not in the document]
rakyll: Done.

trace/trace.go
```diff
@@ -123,34 +123,76 @@ type StartOptions struct {
...
+// WithSpanKind makes new spans to be created with the given kind.
+func WithSpanKind(spanKind int) StartOption {
+	return func(o *StartOptions) {
+		o.SpanKind = spanKind
+	}
+}
+
+// WithSampler makes new spans to be be created with a custom sampler.
+// Otherwise, the global sampler is used.
+func WithSampler(sampler Sampler) StartOption {
```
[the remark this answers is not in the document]
bogdandrutu: I think this will be such a confusion for users. In all the languages we call this WithSpan, I would rather rename these WithSampler or simply use a builder pattern for the StartOptions.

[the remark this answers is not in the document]
bogdandrutu: Looking more into other libraries like https://godoc.org/google.golang.org/grpc I see that with is a pattern used for optional things. So I am fine with this approach.

trace/trace.go
```diff
@@ -123,34 +130,82 @@ type StartOptions struct {
...
+}
+
+// WithRemoteParent makes new spans to be the child of the given part.
+//
+// WithRemoteParent is used to set a parent that has been
+// propagated in an incoming request or RPC.
+//
+// If there is already a parent in the incoming context.Context,
+// this option replaces it with the given parent.
+func WithRemoteParent(parent SpanContext) StartOption {
```
bogdandrutu: I think this is not actually an "optional" thing, in case of a grpc/http server plugins this is required to be set and in all the other cases this is required to not be set. I will have 2 apis StartSpan and StartSpanWithRemoteParent and make this a required field in the second API.

rakyll: I see your point. I added StartSpanWithRemoteParent and removed the WithRemoteParent.

rakyll:
@odeke-em @ramonza can you take a final look for Go readability before I merge?

trace/trace.go
```diff
@@ -125,32 +132,69 @@ type StartOptions struct {
 	SpanKind int
 }
 
+// StartOption apply changes to StartOptions.
+type StartOption func(*StartOptions)
```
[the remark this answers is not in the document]
ramonza: I guess it does for backwards compatibility but maybe we should mark it as deprecated

trace/trace.go
```diff
@@ -125,32 +132,69 @@ type StartOptions struct {
...
+	return func(o *StartOptions) {
+		o.Sampler = sampler
+	}
+}
+
 // StartSpan starts a new child span of the current span in the context. If
 // there is no span in the context, creates a new trace and span.
+func StartSpan(ctx context.Context, name string, o ...StartOption) (context.Context, *Span) {
+	var opts StartOptions
+
```
ramonza: line?

trace/trace.go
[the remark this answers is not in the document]
rakyll: I think we should keep it as it is. It allows ochttp and ocgrpc to provide it as a configuration field.

trace/trace.go
rakyll: Done.

PR #711, opened 2018-04-16
unregister should properly delete View

### LINKED ISSUE #706: TestMetricsEndpointOutput is flaky
```
=== RUN   TestMetricsEndpointOutput
--- FAIL: TestMetricsEndpointOutput (0.01s)
	prometheus_test.go:263: measurement missing in output: baz
=== RUN   TestCumulativenessFromHistograms
--- PASS: TestCumulativenessFromHistograms (0.04s)
```

TestMetricsEndpointOutput fails once a day. Can someone take a look why this test is flaky?

/cc @odeke-em

### LINKED ISSUE #710: stats/view: Unregister always keeps a reference to its view instead of deleting it
While investigating this flake https://github.com/census-instrumentation/opencensus-go/issues/706
I noticed that Unregister doesn't actually delete the underlying view. The bug manifests when the aggregation is of type distribution

Sample proof of concept
```go
package main

import (
	"testing"

	"go.opencensus.io/stats"
	"go.opencensus.io/stats/view"
)

func TestUnregister(t *testing.T) {
	m := stats.Float64("foo", "Foo", "bar")
	for i := 0; i < 10; i++ {
		v := &view.View{
			Aggregation: view.Count(),
			Name:        "Lookup here", Measure: m,
		}
		if err := view.Register(v); err != nil {
			t.Errorf("Iteration #%d error: %v", i, err)
		}
		view.Unregister(v)
	}

	for i := 0; i < 10; i++ {
		v := &view.View{
			Aggregation: view.Distribution(1, 4, 5),
			Name:        "Testing", Measure: m,
		}
		if err := view.Register(v); err != nil {
			t.Errorf("Iteration #%d error: %v", i, err)
		}
		view.Unregister(v)
	}
}
```

which prints
```shell
$ go test
--- FAIL: TestUnregister (0.00s)
	tf_test.go:29: Iteration #1 error: Testing: cannot subscribe view "Testing"; a different view with the same name is already subscribed x="Testing"
	tf_test.go:29: Iteration #2 error: Testing: cannot subscribe view "Testing"; a different view with the same name is already subscribed x="Testing"
	tf_test.go:29: Iteration #3 error: Testing: cannot subscribe view "Testing"; a different view with the same name is already subscribed x="Testing"
	tf_test.go:29: Iteration #4 error: Testing: cannot subscribe view "Testing"; a different view with the same name is already subscribed x="Testing"
	tf_test.go:29: Iteration #5 error: Testing: cannot subscribe view "Testing"; a different view with the same name is already subscribed x="Testing"
	tf_test.go:29: Iteration #6 error: Testing: cannot subscribe view "Testing"; a different view with the same name is already subscribed x="Testing"
	tf_test.go:29: Iteration #7 error: Testing: cannot subscribe view "Testing"; a different view with the same name is already subscribed x="Testing"
	tf_test.go:29: Iteration #8 error: Testing: cannot subscribe view "Testing"; a different view with the same name is already subscribed x="Testing"
	tf_test.go:29: Iteration #9 error: Testing: cannot subscribe view "Testing"; a different view with the same name is already subscribed x="Testing"
FAIL
exit status 1
FAIL	_/Users/emmanuelodeke/Desktop/openSrc/bugs/opencensus-go/unregister	0.004s
```

and notice that it doesn't error out for the CountData because it performs a lookup that matches with the retrieved view from
https://github.com/census-instrumentation/opencensus-go/blob/36f8bed9c5bb5f506b7cde7e7c8088868d05c9e2/stats/view/worker.go#L182-L190

but really we should invoke delete the view after unsubscribing since Unregister is the only left function
```diff
diff --git a/stats/view/worker_commands.go b/stats/view/worker_commands.go
index 0d244ac..c2a59b0 100644
--- a/stats/view/worker_commands.go
+++ b/stats/view/worker_commands.go
@@ -94,6 +94,7 @@ func (cmd *unsubscribeFromViewReq) handleCommand(w *worker) {
 			// The collected data can be cleared.
 			vi.clearRows()
 		}
+		delete(w.views, name)
 	}
 	cmd.done <- struct{}{}
 }
```

It started in PR https://github.com/census-instrumentation/opencensus-go/pull/492 and more specifically the view deletion part was missed or at least not transferred over too to unsubscribeReqView*, from unregisterReqView* but then that awareness wasn't transferred over in 
https://github.com/census-instrumentation/opencensus-go/pull/615 which made *ubscribe now *egister.

Let's fix this and add tests too.

odeke-em:
Fixes #706
Fixes #710

Confusion that came from having two APIs:
a) Unregister
b) Unsubscribe

that initially had different purposes:
a) To completely remove the view from the underlying worker
b) To keep a reference to it but not completely remove it

However the difference between the two manifested in a bug
because we mistakenly assumed the subtle difference between
them never existed, hence on agreeing to deprecate Unsubscribe
for Unregister, the logic for Unsubscribe was kept but that
for Unregister was already lost in previous API changes in
PR https://github.com/census-instrumentation/opencensus-go/pull/492.

PR https://github.com/census-instrumentation/opencensus-go/pull/615
then naively folded Unsubscribe into Unregister and the bug then
fully manifested and could be tested by creating the same view with
a distribution aggregation in a loop and error-ing if Register fails.

With this change, I've also added a test that registers views of
different types, unregisters them in a loop and checks that we
don't error -- which is the original reproducer.
Also while here, just remove the vestiges of unsubscribeViewReq
to solidify the intent and name to unregisterViewReq.

### DIFF
--- a/stats/view/view_test.go
+++ b/stats/view/view_test.go
@@ -371,3 +371,31 @@ func containsRow(rows []*Row, r *Row) bool {
 	}
 	return false
 }
+
+func TestRegisterUnregisterParity(t *testing.T) {
+	measures := []stats.Measure{
+		stats.Int64("ifoo", "iFOO", "iBar"),
+		stats.Float64("ffoo", "fFOO", "fBar"),
+	}
+	aggregations := []*Aggregation{
+		Count(),
+		Sum(),
+		Distribution(1, 2.0, 4.0, 8.0, 16.0),
+	}
+
+	for i := 0; i < 10; i++ {
+		for _, m := range measures {
+			for _, agg := range aggregations {
+				v := &View{
+					Aggregation: agg,
+					Name:        "Lookup here",
+					Measure:     m,
+				}
+				if err := Register(v); err != nil {
+					t.Errorf("Iteration #%d:\nMeasure: (%#v)\nAggregation (%#v)\nError: %v", i, m, agg, err)
+				}
+				Unregister(v)
+			}
+		}
+	}
+}
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -86,7 +86,7 @@ func Unregister(views ...*View) {
 	for i := range views {
 		names[i] = views[i].Name
 	}
-	req := &unsubscribeFromViewReq{
+	req := &unregisterFromViewReq{
 		views: names,
 		done:  make(chan struct{}),
 	}
--- a/stats/view/worker_commands.go
+++ b/stats/view/worker_commands.go
@@ -73,15 +73,15 @@ func (cmd *registerViewReq) handleCommand(w *worker) {
 	}
 }
 
-// unsubscribeFromViewReq is the command to unsubscribe to a view. Has no
+// unregisterFromViewReq is the command to unsubscribe to a view. Has no
 // impact on the data collection for client that are pulling data from the
 // library.
-type unsubscribeFromViewReq struct {
+type unregisterFromViewReq struct {
 	views []string
 	done  chan struct{}
 }
 
-func (cmd *unsubscribeFromViewReq) handleCommand(w *worker) {
+func (cmd *unregisterFromViewReq) handleCommand(w *worker) {
 	for _, name := range cmd.views {
 		vi, ok := w.views[name]
 		if !ok {
@@ -94,6 +94,7 @@ func (cmd *unsubscribeFromViewReq) handleCommand(w *worker) {
 			// The collected data can be cleared.
 			vi.clearRows()
 		}
+		delete(w.views, name)
 	}
 	cmd.done <- struct{}{}
 }


stats/view/view_test.go
```diff
@@ -371,3 +371,31 @@ func containsRow(rows []*Row, r *Row) bool {
 	}
 	return false
 }
+
+func TestRegisterUnregisterParity(t *testing.T) {
```
rakyll: What does this test case test?

odeke-em: It tests that invoking register then unregister n number of times results in a full deregistration and never errors like it is currently doing before this PR. It is basically the reproducer for bug #710

rakyll: You can just say TestRegisterUnregister.

odeke-em:
Thanks for the review @rakyll!

PR #712, opened 2018-04-17
ocgrpc: Do not use Distribution aggregation for Count stats

atetubou:
I think this is fix for intentional behavior.


atetubou:
Sorry I didn't understand the meaning of this metrics.
I forgot about streaming rpcs...

PR #716, opened 2018-04-18
Allow updating the README in place

ramonza:
Updating README from a separate source file is confusing and has
several times been accidentally missed, leading to lost updates the
next time someone updates it.

Change to using in-place update together with a check (in Travis)
that the file is in sync with the example code.


.travis.yml
```diff
@@ -14,8 +14,10 @@ before_script:
   - GO_FILES=$(find . -iname '*.go' | grep -v /vendor/)  # All the .go files, excluding vendor/ if any
   - PKGS=$(go list ./... | grep -v /vendor/)             # All the import paths, excluding vendor/ if any
   - curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh # Install latest dep release
+  - go get github.com/rakyll/embedmd
 
 script:
+  - embedmd -d README.md
```
rakyll: The output of this command will not be committed back to the repo.

Remove this step from travis.

ramonza: actually the intention here is not to commit it back to the repo but to verify that there are no differences between what would be generated and what is in the repo (the command returns non-zero upon finding a difference). I will add a comment to this effect.

ramonza:
@rakyll tested updating both the README and embedded code and it all seems to work

PR #729, opened 2018-04-25
Fix go get command

cooldarkdryplace:
Right now `go get` returns error:
`can't load package: package go.opencensus.io/examples/grpc: no Go files in /home/heliko/go/src/go.opencensus.io/examples/grpc`


rakyll:
Thanks!

Issue #713, opened 2018-04-17
Problematic inner mapping of HTTP Status code to OC Status


basvanbeek:
ochttp uses an unexported function to map HTTP status codes to unexported trace status codes which locks out usage by other instrumentation that might want to do this. 

Since this contains mappings that we really want to be consistent across instrumentation dealing with HTTP it might be better to refactor this and place the mappings in a separate package of opencensus-go.  Since it is suggested that OC Status codes should follow https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto it might make sense to create a new package holding generated source from this proto together with idiomatic mappings for instrumentations to re-use. This could expand beyond HTTP. 

Currently we lack explicitness and proliferation of instrumentation will probably degrade the quality of Status reporting and usage in the various tracing exporters.

rakyll:
We should export status codes from the trace package ideally, we haven't done it yet because we were debating about the status codes and whether we should follow the rpc/code.proto.

/cc @bogdandrutu

PR #734, opened 2018-04-26
Export constants for trace.Status codes

ramonza:
Fixes: #713

### DIFF
--- a/trace/status_codes.go
+++ b/trace/status_codes.go
@@ -1,0 +1,37 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


trace/status_codes.go
```diff
@@ -0,0 +1,36 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trace
+
+// Status codes for use with Span.SetStatus.
```
odeke-em: It would be nice to keep a reference to where we obtained these status codes i.e. https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

trace/status_codes.go
```diff
@@ -0,0 +1,36 @@
...
+	StatusCodePermissionDenied   = 7
+	StatusCodeResourceExhausted  = 8
+	StatusCodeFailedPrecondition = 9
+	StatusCodeAborted            = 10
+	StatusCodeOutOfRange         = 11
+	StatusCodeUnimplemented      = 12
+	StatusCodeInternal           = 13
+	StatusCodeUnavailable        = 14
+	StatusCodeDataLoss           = 15
+	StatusCodeUnathenticated     = 16
```
atetubou: typo? Unathenticated -> Unauthenticated

ramonza: nice catch thanks!

Issue #441, opened 2018-02-12
Implement hour and minute stats in zPages


ramonza:
After the removal of Window from View, zPages don't support hour and minutely stats. We need to add this back.

PR #736, opened 2018-04-30
WIP: Restore zpages interval stats functionality

ramonza:
Fixes: #441


ramonza:
@odeke-em would you be able to review since @rakyll is away this week?

odeke-em:
@ramonza thank you for the page. For sure, I'll take a look at it over the week.

zpages/internal/templates/rpcz.html
```diff
@@ -28,24 +28,24 @@
     <tr>
         <td><b>{{.Method}}</b></td>
         <td></td>
-        <td align="right">{{.CountMinute|count}}</td>
```
odeke-em: Where are we getting these column changes from?

odeke-em: Oh I see, from the latest spec changes as per https://github.com/census-instrumentation/opencensus-go/commit/85464adb436a3dcfbca051c298ddd2cbc97c5e68

zpages/rpcz.go
```diff
@@ -205,113 +189,112 @@ func (s snapExporter) ExportView(vd *view.Data) {
...
-					s.ErrorsTotal += int(count)
-				}
-			}
-
 		case ocgrpc.ClientRoundtripLatencyView:
-			s.AvgLatencyTotal = convertTime(sum / count)
-
+			hour = &s.LatencyHour
+			minute = &s.LatencyMinute
 		case ocgrpc.ClientSentBytesPerRPCView:
```
odeke-em: It might aid with readability if you could couple the outputs and rates together instead of a new case clause for each view i.e.
```go
case ocgrpc.ClientSentBytesPerRPCView, ocgrpc.ServerCompletedRPCsView:
       hour = &s.OutputHour
       minute = &s.OutputMinute
case ocgrpc.ClientLatencyView, ocgrpc.ServerLatencyView:
       hour = &s.LatencyHour
       minute = &s.LatencyMinute
```

zpages/rpcz.go
```diff
@@ -205,113 +189,112 @@ func (s snapExporter) ExportView(vd *view.Data) {
...
+		hour.update(vd.End, dist)
+		minute.update(vd.End, dist)
+	}
 
-		case ocgrpc.ServerReceivedMessagesPerRPCView:
-			s.CountTotal = int(count)
-			s.RPCRateTotal = computeRate(0, count)
+}
+
+func getStatus(row *view.Row) string {
```
odeke-em:
```go
func clientOrServerStatus
```

zpages/rpcz.go
```diff
@@ -205,113 +189,112 @@ func (s snapExporter) ExportView(vd *view.Data) {
...
+		if tag.Key == ocgrpc.KeyServerStatus || tag.Key == ocgrpc.KeyClientStatus {
+			return tag.Value
+		}
+	}
+	return ""
+}
 
-		case ocgrpc.ServerSentMessagesPerRPCView:
-			// currently unused
+func getMethod(row *view.Row) string {
```
odeke-em:
```go
func clientOrServerMethod
```

zpages/rpcz.go
```diff
@@ -205,113 +189,112 @@ func (s snapExporter) ExportView(vd *view.Data) {
...
+		return s
+	}
+	s := &statSnapshot{Method: method, Received: received}
+	for _, hourWindow := range []*windowStat{&s.OutputHour, &s.LatencyHour, &s.ErrorsHour, &s.InputHour} {
+		hourWindow.init(time.Hour)
+	}
+	for _, minuteWindow := range []*windowStat{&s.OutputMinute, &s.LatencyMinute, &s.ErrorsMinute, &s.InputMinute} {
+		minuteWindow.init(time.Minute)
+	}
+	snaps[key] = s
```
odeke-em: Should we be protecting this with a mutex?

zpages/window_stat.go
```diff
@@ -0,0 +1,124 @@
...
+	startDist, endDist := startInterval.distribution, endInterval.distribution
+	actualStart, actualEnd = startInterval.updateTime, endInterval.updateTime
+
+	buckets := len(endDist.CountPerBucket)
+	if buckets != len(startDist.CountPerBucket) {
+		panic("can't subtract distributions with different number of buckets")
+	}
+	diff = &view.DistributionData{}
+	diff.Count = endDist.Count - startDist.Count
+	diff.Mean = (endDist.Sum() - startDist.Sum()) / float64(diff.Count)
```
odeke-em: Isn't this always going to be NaN? diff.Count will be zero since endDist.Count == startDist.Count

odeke-em: Rather, Infinity

zpages/window_stat.go
```diff
@@ -0,0 +1,124 @@
...
+	if intv.index < ind {
+		*intv = interval{index: ind}
+		return sliceIndex, intv
+	}
+	// we are attempting to get an interval for something that has already
+	// fallen out of the circular buffer
+	return -1, nil
+}
+
+func (ws *windowStat) read() (actualStart, actualEnd time.Time, diff *view.DistributionData) {
```
odeke-em: Is this all newly written code or is it just reverted code? I don't have too much context about it, hence why am asking.

zpages/rpcz.go
ramonza: yes! much better thanks

zpages/templates.go
```diff
@@ -116,3 +114,83 @@ func traceIDFormatter(r traceRow) template.HTML {
...
+	}
+	return countFormatter(ws.intervals[ws.lastUpdate].distribution.Count)
+}
+
+func dataRate(ws *windowStat) string {
+	start, end, diff := ws.read()
+	if diff == nil {
+		return dataRateFormatter(0)
+	}
+	seconds := float64(end.Sub(start) / time.Second)
```
odeke-em: I'd be more comfortable using the method `Seconds` e.g.
```go
seconds := end.Sub(start).Seconds()
```
makes for more clarity

zpages/templates.go
```diff
@@ -116,3 +114,83 @@ func traceIDFormatter(r traceRow) template.HTML {
...
+	rate := float64(diff.Count) / seconds
+	return rateFormatter(rate)
+}
+
+func totalVelocity(ws *windowStat) string {
+	if ws.lastUpdate == -1 {
+		return "-"
+	}
+	last := ws.intervals[ws.lastUpdate]
+	seconds := float64(last.updateTime.Sub(ws.startTime)) / float64(time.Second)
```
odeke-em: Ditto about using `last.updateTime.Sub(ws.startTime).Seconds()`

zpages/templates.go
```diff
@@ -116,3 +114,83 @@ func traceIDFormatter(r traceRow) template.HTML {
...
+	}
+	return dataRateFormatter(float64(diff.Sum()) / seconds)
+}
+
+func totalDataRate(ws *windowStat) string {
+	if ws.lastUpdate == -1 {
+		return dataRateFormatter(0)
+	}
+	last := ws.intervals[ws.lastUpdate]
+	seconds := float64(last.updateTime.Sub(ws.startTime) * time.Second)
```
odeke-em: Ditto

zpages/templates.go
```diff
@@ -116,3 +114,83 @@ func traceIDFormatter(r traceRow) template.HTML {
 func even(x int) bool {
 	return x%2 == 0
```
odeke-em: Nicer to do `x & 1 == 0` although the compiler already converts it with a rewrite to the suggestion I provided.

zpages/templates.go
```diff
@@ -80,22 +81,19 @@ func countFormatter(num int) string {
...
-	return strconv.Itoa(int(d / time.Millisecond))
+	return strconv.Itoa(int(latency))
 }
 
 func rateFormatter(r float64) string {
 	return fmt.Sprintf("%.3f", r)
 }
 
 func dataRateFormatter(b float64) string {
-	return fmt.Sprintf("%.3f", b/1e6)
```
odeke-em: What changed the value here?

zpages/templates.go
```diff
@@ -80,22 +81,19 @@ func countFormatter(num int) string {
...
-		return "0"
-	}
-	if d < 10*time.Millisecond {
-		return fmt.Sprintf("%.3f", float64(d)*1e-6)
+func msFormatter(latency float64) string {
+	if latency < 10.0 {
+		return fmt.Sprintf("%.2f", latency)
 	}
-	return strconv.Itoa(int(d / time.Millisecond))
+	return strconv.Itoa(int(latency))
```
odeke-em: This change doesn't make sense to me, in this case why wouldn't we instead just `fmt.Sprintf` as we have right above

zpages/rpcz.go
ramonza: It already is higher up in the call stack in ExportView

zpages/window_stat.go
ramonza: It can be if there is no difference between start and end (i.e. no change to stats between those time intervals). In this case Sum() will be equal too so the result will be Mean = NaN which is fine.

zpages/window_stat.go
ramonza: It is new

zpages/templates.go
ramonza: I am computing the seconds between start and end. The Seconds() function does not convert to seconds, it just gives you the seconds component of the time.

PR #741, opened 2018-05-05
Add ClientTracer from httptrace for detailed connection info

odeke-em:
Provide detailed connection information to help figure
out where a request is spending most of its time.

## Demo
Given this code below
```go
package main

import (
	"context"
	"io"
	"io/ioutil"
	"log"
	"net/http"

	"go.opencensus.io/exporter/stackdriver"
	"go.opencensus.io/trace"
)

func main() {
	sd, err := stackdriver.NewExporter(stackdriver.Options{ProjectID: "census-demos"})
	if err != nil {
		log.Fatalf("Failed to create Stackdriver exporter: %v", err)
	}
	trace.RegisterExporter(sd)
	defer sd.Flush()

	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})

	ctx, span := trace.StartSpan(context.Background(), "Fetching URL")
	defer span.End()

	tr := &ochttp.Transport{}
	client := &http.Client{Transport: tr}
	req, _ := http.NewRequest("GET", "https://opencensus.io", nil)
	req = req.WithContext(ctx)
	res, err := client.Do(req)
	if err != nil {
		log.Fatalf("Failed to GET: %v", err)
	}
	_, _ = io.Copy(ioutil.Discard, res.Body)
	_ = res.Body.Close()
}
```

and running it 
```shell
GOOGLE_APPLICATION_CREDENTIALS=~/Downloads/census-demos-237a8e1e41df.json go run main.go 
```

### Before
screen shot 2018-05-04 at 10 47 22 pm

### After
screen shot 2018-05-04 at 10 48 02 pm

### DIFF
--- a/plugin/ochttp/trace.go
+++ b/plugin/ochttp/trace.go
@@ -15,8 +15,11 @@
 package ochttp
 
 import (
+	"context"
+	"crypto/tls"
 	"io"
 	"net/http"
+	"net/http/httptrace"
 	"net/url"
 
 	"go.opencensus.io/plugin/ochttp/propagation/b3"
@@ -53,11 +56,80 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	name := spanNameFromURL(req.URL)
 	// TODO(jbd): Discuss whether we want to prefix
 	// outgoing requests with Sent.
-	_, span := trace.StartSpan(req.Context(), name,
+	ctx, span := trace.StartSpan(req.Context(), name,
 		trace.WithSampler(t.startOptions.Sampler),
 		trace.WithSpanKind(trace.SpanKindClient))
 
-	req = req.WithContext(trace.WithSpan(req.Context(), span))
+	tracer := &httptrace.ClientTrace{
+		GetConn: func(hostPort string) {
+			gctx, span := trace.StartSpan(ctx, "GetConn")
+			ctx = context.WithValue(gctx, "getConnSpan", span)
+		},
+		GotConn: func(_ httptrace.GotConnInfo) {
+			span := ctx.Value("getConnSpan").(*trace.Span)
+			span.End()
+		},
+		DNSStart: func(_ httptrace.DNSStartInfo) {
+			dctx, span := trace.StartSpan(ctx, "DNSLookup")
+			ctx = context.WithValue(dctx, "dnsSpan", span)
+		},
+		DNSDone: func(di httptrace.DNSDoneInfo) {
+			span := ctx.Value("dnsSpan").(*trace.Span)
+			if di.Err != nil {
+				span.SetStatus(trace.Status{
+					Code:    trace.StatusCodeInternal,
+					Message: di.Err.Error(),
+				})
+			}
+			span.End()
+		},
+		WroteHeaders: func() {
+			span := trace.FromContext(ctx)
+			span.Annotate(nil, "Wrote request")
+		},
+		ConnectStart: func(net, addr string) {
+			cctx, span := trace.StartSpan(ctx, "Connect")
+			span.Annotate([]trace.Attribute{
+				trace.StringAttribute("network", net),
+				trace.StringAttribute("address", addr),
+			}, "Starting connection")
+			ctx = context.WithValue(cctx, "connectSpan", span)
+		},
+		ConnectDone: func(net, addr string, err error) {
+			span := ctx.Value("connectSpan").(*trace.Span)
+			span.Annotate([]trace.Attribute{
+				trace.StringAttribute("network", net),
+				trace.StringAttribute("address", addr),
+			}, "Established Connection")
+			if err != nil {
+				span.SetStatus(trace.Status{
+					Code:    trace.StatusCodeInternal,
+					Message: err.Error(),
+				})
+			}
+			span.End()
+		},
+		TLSHandshakeStart: func() {
+			tctx, span := trace.StartSpan(ctx, "TLSHandshake")
+			ctx = context.WithValue(tctx, "tlshandshakeSpan", span)
+		},
+		TLSHandshakeDone: func(cs tls.ConnectionState, err error) {
+			span := ctx.Value("tlshandshakeSpan").(*trace.Span)
+			span.Annotate([]trace.Attribute{
+				trace.Int64Attribute("tls_version", int64(cs.Version)),
+			}, "TLS Info")
+			if err != nil {
+				span.SetStatus(trace.Status{
+					Code:    trace.StatusCodeInternal,
+					Message: err.Error(),
+				})
+			}
+			span.End()
+		},
+	}
+	ctx = httptrace.WithClientTrace(ctx, tracer)
+
+	req = req.WithContext(ctx)
 	if t.format != nil {
 		t.format.SpanContextToRequest(span.SpanContext(), req)
 	}


plugin/ochttp/trace.go
```diff
@@ -53,11 +56,80 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
-	_, span := trace.StartSpan(req.Context(), name,
+	ctx, span := trace.StartSpan(req.Context(), name,
 		trace.WithSampler(t.startOptions.Sampler),
 		trace.WithSpanKind(trace.SpanKindClient))
 
-	req = req.WithContext(trace.WithSpan(req.Context(), span))
+	tracer := &httptrace.ClientTrace{
+		GetConn: func(hostPort string) {
+			gctx, span := trace.StartSpan(ctx, "GetConn")
+			ctx = context.WithValue(gctx, "getConnSpan", span)
```
ramonza: let's create a private type to use as the key to avoid potential collisions

plugin/ochttp/trace.go
```diff
@@ -53,11 +56,80 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	name := spanNameFromURL(req.URL)
 	// TODO(jbd): Discuss whether we want to prefix
 	// outgoing requests with Sent.
-	_, span := trace.StartSpan(req.Context(), name,
+	ctx, span := trace.StartSpan(req.Context(), name,
 		trace.WithSampler(t.startOptions.Sampler),
 		trace.WithSpanKind(trace.SpanKindClient))
 
-	req = req.WithContext(trace.WithSpan(req.Context(), span))
+	tracer := &httptrace.ClientTrace{
```
ramonza: Does this need to be allocated afresh each time or can it be statically allocated e.g. in a package global?

plugin/ochttp/trace.go
```diff
@@ -53,11 +56,80 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+			if err != nil {
+				span.SetStatus(trace.Status{
+					Code:    trace.StatusCodeInternal,
+					Message: err.Error(),
+				})
+			}
+			span.End()
+		},
+		TLSHandshakeStart: func() {
+			tctx, span := trace.StartSpan(ctx, "TLSHandshake")
```
ramonza: I would suggest that we start all these internal spans with some prefix, such as "httptrace.TLSHandshake"

plugin/ochttp/trace.go
```diff
@@ -53,11 +56,80 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+				span.SetStatus(trace.Status{
+					Code:    trace.StatusCodeInternal,
+					Message: di.Err.Error(),
+				})
+			}
+			span.End()
+		},
+		WroteHeaders: func() {
+			span := trace.FromContext(ctx)
+			span.Annotate(nil, "Wrote request")
```
ramonza: Annotation should be "wrote request headers" - might be misleading to just say "wrote request"

plugin/ochttp/trace.go
basvanbeek: Agreed... in `zipkin-go` we prefix all with "httptrace."

basvanbeek:
I would not automatically enable this for all http client spans. This is expensive enough to be off by default and enabled by config when needed.

basvanbeek:
@ramonza I disagree... this is a choice we should allow the consumer to make. We can argue about it being important enough with acceptable overhead for it to be on by default but it should be adjustable by the end user.

Personally I think OC overhead should be minimized by default and additional details should be enabled when needed. No need in having customers incur higher cost or needing to add resources because we're being overly verbose by default.

ramonza:
@basvanbeek are you worried about resource consumption (CPU, memory) or trace verbosity? 

If resource consumption, then doesn't sampling handle this already? The user can decrease the sampling if the overhead is too high. It would be good to have some benchmarks here to quantify it. I would prefer not to add configuration complexity if the overhead isn't large compared to the overhead we already incur anyway if a trace is sampled (collecting and exporting).

As for verbosity of traces an alternative would be to log these as events/annotations rather than sub-spans.

basvanbeek:
This will be a very long winded conversation on why we should limit amount of data by default which I do not have time for at this moment to explain.

But just to name a few: 
- lowering sampling rate especially with unweighted endpoints means that valuable information on certain endpoints might be dropped over relatively unexciting high volume endpoints due to chance of being sampled going drastically down.
- lots of times trace investigation is about finding issues in micro service interop, parallelism vs serialism, failures down the tree and information on connection establishment is not really high on the list of things to see if most service to service calls are local.
- storage / retrieval costs on various backends we wish to support might be a bottleneck.
- lots of users start with high sampling rates to simply find out how their system behaves. When they reach high volumes they start to sample, but do this relatively late. forcing this decision to come sooner by over verbosity might not be that great of a story to users.
- OC instrumentation for Go is already on the heavy side with respect to time used on the critical path, no need to make this worse when not explicitly called for by consumer.

Me as a consumer would really dislike for you as an instrumentation engineer to force these decisions on me. I'm very well capable of configuring the items I want / don't want as long as you give me that option. Don't expect you know all the use cases and configuration needs as I'm convinced I don't know either.

As said it is debatable if config for this option should be on or off by default but I'll fight for the option for it to be setable.

ramonza:
@odeke-em what are your thoughts on this? Did you do any benchmarks? Might be good to add benchmarks since the issue has come up now.

I do think that if we add options for performance reasons, we need to support this claim with benchmarks showing the effect of the option. 

Thinking more about it, I would actually double down on the suggestion to use annotations rather than sub-spans. This would reduce the verbosity and probably(?) the resource cost as well. @basvanbeek do you have thoughts on this?

plugin/ochttp/trace.go
odeke-em: Thanks, yeah this slipped my mind and I was thinking of it too.

plugin/ochttp/trace.go
odeke-em: We need to freshly allocate in order to reuse the context passed in between intermediate trace steps.

plugin/ochttp/trace.go
odeke-em: Nice, thanks, I'll do that.

plugin/ochttp/trace.go
odeke-em: Gotcha, thanks!

odeke-em:
Thank you @ramonza and @basvanbeek for the reviews and for the discourse.
In deed, as @basvanbeek advocated against, this change would be very latency heavy as the benchmarks show
```go
var sink *http.Response

func BenchmarkRoundTrip(b *testing.B) {
        b.ReportAllocs()
        // We need a functional server to test out client tracing
        cst := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
                _, _ = ioutil.ReadAll(r.Body)
                _ = r.Body.Close()
                w.Write([]byte("Hello, there"))
        }))
        defer cst.Close()

        client := &http.Client{Transport: new(Transport)}
        for i := 0; i < b.N; i++ {
                res, err := client.Get(cst.URL)
                if err != nil {
                        b.Fatalf("Failed to invoke get: %v", err)
                }
                _ = res.Body.Close()
                sink = res
        }
        if sink == nil {
                b.Fatalf("Did not run benchmark!")
        }
}
```

### After
```shell
BenchmarkRoundTrip-4   	    5000	    380716 ns/op	   21315 B/op	     208 allocs/op
```

### Before
```shell
BenchmarkRoundTrip-4   	    2000	    816372 ns/op	   20333 B/op	     175 allocs/op
```

With annotations added, the number of allocs increases by a bit but not as badly as with subspans
```shell
BenchmarkRoundTrip-4   	    3000	    377179 ns/op	   20836 B/op	     195 allocs/op
```

My use case was investigating which parts of a network bound request were pathologically slow as the rest of the part of my microservices returned in at most 5ms but the network cost me like 400ms or so. 

@basvanbeek and you @ramonza both raise great points; I think the extent of the allocations and latency created don't give this change merit. However, an advanced user that really would like to trace through their request perhaps might be able to attach a clientTracer to their transport.

The annotations work well and are less heavier but perhaps not every user needs as much detail as I did need, in my very niche case?

PR #743, opened 2018-05-07
Introduce tag.NewLocalKey for non-propagating tags

ramonza:
See: census-instrumentation/opencensus-specs#110


ramonza:
We are going with a different approach: https://github.com/census-instrumentation/opencensus-specs/issues/110

PR #745, opened 2018-05-08
Support reading the values from a tag.Map

ramonza:
This is needed so that we can use tag.Decode from Istio to
populate attributes based on the tag header.

### DIFF
--- a/tag/map.go
+++ b/tag/map.go
@@ -33,6 +33,18 @@ type Map struct {
 	m map[Key]string
 }
 
+// GetTags reads the tags contained in the map. It shouldn't usually be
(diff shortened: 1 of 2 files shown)


bogdandrutu:
Why? I don't think this is a good idea. tag.Decode will return a Map that you pass to the stats, why do you need the values?

ramonza:
I need to extract the values from the header as Istio attributes. I am not passing them to OpenCensus stats. The alternative would be to reimplement the tag header parsing logic in Istio which seems unnecessary.

ramonza:
Going to and handle this another way

Issue #744, opened 2018-05-07
plugin/ochttp: implement io.ReadFrom


groob:
https://github.com/census-instrumentation/opencensus-go/issues/642 was opened to implement the hijacker interface for websocket support. 
Beside Hijacker and Flusher the default implementation of `http.ResponseWriter` supports a few other additional "upgradable" interfaces. When these are not implemented in middleware it can lead to subtle bugs and performance regressions in proxies that use the http plugin. 

https://avtok.com/2014/11/05/interface-upgrades.html
https://github.com/felixge/httpsnoop#why-this-package-exists

Popular middleware packages like https://github.com/gorilla/handlers and [chi](https://github.com/go-chi/chi) implement the entire subset of these methods. It looks like the ochttp plugin only supports Hijack and Flush today, leaving `CloseNotify`, `Push`, and `ReadFrom`.

odeke-em:
Thank you for raising this @groob!

I actually also just encountered this while instrumenting a popular webserver
```shell
rec: httpserver.NonCloseNotifierError{Underlying:(*ochttp.trackingResponseWriter)(0xc0001dc070)
```

odeke-em:
@groob I've mailed this PR https://github.com/census-instrumentation/opencensus-go/pull/747 PTAL. I also need most of the interfaces implemented to complete the integration that I am working on.

odeke-em:
Alright, I've merged the PR that implements the pure HTTP interfaces:
* `http.Pusher`
* `http.CloseNotifier`

only `io.ReadFrom` is left

basvanbeek:
I'm not convinced the current approach is a good one. The implementation done in #747 has the negative side effect of implementing all "upgradable" interfaces that might not have been implemented by the underlying wrapped implementation. If any code is making choices based on an upgradable interface being implemented or not it will potentially get a false positive from ochttp.

The `httpsnoop` package as found at https://github.com/felixge/httpsnoop actually correctly deals with this by returning a handler that only returns an implementation which includes the actually wrapped interfaces and not all of them. Since this interface logic from httpsnoop is decoupled from its own metrics middleware handling we could use httpsnoop as a dependency and save ourselves some work.

basvanbeek:
[...]

PR #747, opened 2018-05-10
Make plugin/ochttp/Handler implement common HTTP server interfaces

odeke-em:
For #744

Make handler implement the following interfaces:
* http.CloseNotifier
* http.Pusher
and also adds tests to lock this behavior in and ensure
that we never regress.

In the future, we'll need to figure out why net/http is
still tripping out with http.ErrNotSupported for any
sort of server/connection combination with or without
ochttp.Handler as the handler. For now the current
code I believe satisfactorily implements the http.Pusher
interface.

### DIFF
--- a/plugin/ochttp/server.go
+++ b/plugin/ochttp/server.go
@@ -140,8 +140,12 @@ type trackingResponseWriter struct {
 	writer     http.ResponseWriter
 }
 
-var _ http.ResponseWriter = (*trackingResponseWriter)(nil)
+// Compile time assertions for widely used net/http interfaces
+var _ http.CloseNotifier = (*trackingResponseWriter)(nil)
+var _ http.Flusher = (*trackingResponseWriter)(nil)
 var _ http.Hijacker = (*trackingResponseWriter)(nil)
+var _ http.Pusher = (*trackingResponseWriter)(nil)
+var _ http.ResponseWriter = (*trackingResponseWriter)(nil)
 
 var errHijackerUnimplemented = errors.New("ResponseWriter does not implement http.Hijacker")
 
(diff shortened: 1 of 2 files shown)


plugin/ochttp/server.go
```diff
@@ -153,6 +157,22 @@ func (t *trackingResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 	return hj.Hijack()
 }
 
+func (t *trackingResponseWriter) CloseNotify() <-chan bool {
+	cn, ok := t.writer.(http.CloseNotifier)
```
ramonza: I think it might read better if this was used inline in the `if` like:
``` 
if cn, ok := t.writer.(http.CloseNotifier); ok {
...
```

plugin/ochttp/server.go
```diff
@@ -153,6 +157,22 @@ func (t *trackingResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
...
+	if !ok {
+		return nil
+	}
+	return cn.CloseNotify()
+}
+
+func (t *trackingResponseWriter) Push(target string, opts *http.PushOptions) error {
+	pusher, ok := t.writer.(http.Pusher)
+	if !ok {
```
ramonza: ditto

plugin/ochttp/server.go
```diff
@@ -140,8 +140,12 @@ type trackingResponseWriter struct {
 	writer     http.ResponseWriter
 }
 
-var _ http.ResponseWriter = (*trackingResponseWriter)(nil)
+// Compile time assertions for widely used net/http interfaces
+var _ http.CloseNotifier = (*trackingResponseWriter)(nil)
+var _ http.Flusher = (*trackingResponseWriter)(nil)
 var _ http.Hijacker = (*trackingResponseWriter)(nil)
+var _ http.Pusher = (*trackingResponseWriter)(nil)
+var _ http.ResponseWriter = (*trackingResponseWriter)(nil)
```
ramonza: what about ReaderFrom as mentioned in the linked issue? is that intentionally left to a future PR? Also probably should implement WriterTo. These are more involved (they need to work even if the underlying streams don't support them directly) so maybe we need a separate issue tracking them.

odeke-em:
Thank you @ramonza for the PR review, I am going to merge this, with minor veto on the comments :) but also because I'd like to finish and ship out this integration. What's remaining is `io.WriterTo`

rakyll:
We earlier decided to return an error but I think we should optionally implement these methods if the underlying Handler implements them.

PR #749, opened 2018-05-14
Update cloud.google.com to v0.22.0

aelleth:
`0.x.y` constraints in Gopkg.toml are equivalent to `>= 0.x.y, < 0.(x+1).0`.  To keep in sync with [cloud.google.com/go](https://github.com/GoogleCloudPlatform/google-cloud-go), we have to bump its version in the `dep` manifest.

I'm also hoping to (re)boot a discussion in #703 by @johanbrandhorst regarding dependencies and when they should be updated.

From what I gather, [cloud.google.com/go](https://github.com/GoogleCloudPlatform/google-cloud-go) has a minor version bump roughly once a month. The other two dependencies `< 1.0.0` are [github.com/openzipkin/zipkin-go](https://github.com/openzipkin/zipkin-go) and [github.com/prometheus/client_golang](https://github.com/prometheus/client_golang).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -226,8 +226,6 @@
     "metadata",
     "naming",
     "peer",
-    "reflection",
-    "reflection/grpc_reflection_v1alpha",
     "resolver",
     "resolver/dns",
     "resolver/passthrough",
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -9,8 +9,8 @@
     "monitoring/apiv3",
     "trace/apiv2"
   ]
-  revision = "29f476ffa9c4cd4fd14336b6043090ac1ad76733"
-  version = "v0.21.0"
+  revision = "056a55f54a6cc77b440b31a56a5e7c3982d32811"
+  version = "v0.22.0"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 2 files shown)


johanbrandhorst:
I second @aelleth's suggestion for a formal process for dependency updates - this project uses quite a few high level projects with frequent updates (gRPC, Jaeger, Zipkin, Prometheus). I'm not sure if there is already a planned release cadence for this project, but perhaps 1 week before release dependencies could be bumped (to iron out any problems before releases)?

ramonza:
@aelleth was this change prompted by being unable to upgrade cloud.google.com/go? If so, I think we should forestall this in future by declaring ">=0.21.0" as the dependency version, rather than just bumping one version.

aelleth:
@ramonza this was the case indeed.

Fixing all dependencies < 1.0.0 at `>= 0.x.0` seems reasonable. That way it'll only require review if CI/tests break or any dependency is released at 1.0.0.

ramonza:
@aelleth mind updating the PR along those lines when you get a chance?

ramonza:
I needed this to build another project so implemented in https://github.com/census-instrumentation/opencensus-go/pull/756.

I will try to release on a biweekly schedule every second Mondays and upgrade dependencies right after the release.

Issue #707, opened 2018-04-16
Default Tags


bbassingthwaite-va:
Hi,

It would be very useful to set a default set of attributes that are applied to all traces. For example, we would like to be able to add availability zone, environment, etc... that go out on all traces. I don't believe there is a way to do this? Or I am missing it? 

Thanks!

bogdandrutu:
I think this should be supported in the exporter configuration. When you register the exporter you pass a map of <key, value> for attributes that will be added to all the Spans.

gottwald:
I'd really like to have this feature, so I added a possible implementation.
Not sure if the exporter config is the right place to put this.
One caveat in this implementation is that the exporter attributes override normal span attributes in case of key conflicts. (Not a big issue for my use case but maybe for others)

ramonza:
Once way to do this without adding the functionality to every exporter would be to implement a wrapper exporter. Perhaps we could add this wrapper to an `exporters/exporterutil` package?

So something like:
```go

enhancer := exporterutil.NewSpanEnhancer(myExporter)
enhancer.DefaultAttributes = append(wrapper.DefaultAttributes, trace.StringAttribute("xxx", "xxx"), ...)
enhancer.SpanNameTemplate = "{{.SpanKind}} {{.Attribute "http.route"}}"
enhancer.ComputedAttributes = func(sd *trace.SpanData) []trace.Attribute {
  dur := sd.EndTime.Sub(sd.StartTime)
  if dur > 5*time.Second {
    return []trace.Attribute{ trace.BoolAttribute("slow", true) }
  }
  return nil
}
trace.RegisterExporter(enhancer)
```

(Threw in a bunch of other random ideas that a span "enhancer" might want to do before the Span is exported)

PR #750, opened 2018-05-14
[trace] Add default span attributes

gottwald:
This adds default span attributes that can be submitted when registering
the exporter. All submitted attributes are set on every span when it's
being exported.

refs #707

### DIFF
--- a/trace/export.go
+++ b/trace/export.go
@@ -32,19 +32,20 @@ type Exporter interface {
 
 var (
 	exportersMu sync.Mutex
-	exporters   map[Exporter]struct{}
+	exporters   map[Exporter][]Attribute
 )
 
 // RegisterExporter adds to the list of Exporters that will receive sampled
-// trace spans.
+// trace spans. Passed attributes will be appended to every span that is
+// exported by this exporter.
 //
 // Binaries can register exporters, libraries shouldn't register exporters.
-func RegisterExporter(e Exporter) {
+func RegisterExporter(e Exporter, a ...Attribute) {
 	exportersMu.Lock()
 	if exporters == nil {
-		exporters = make(map[Exporter]struct{})
+		exporters = make(map[Exporter][]Attribute)
 	}
-	exporters[e] = struct{}{}
+	exporters[e] = a
 	exportersMu.Unlock()
 }
 
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -278,7 +278,13 @@ func (s *Span) End() {
 		if s.spanContext.IsSampled() {
 			// TODO: consider holding exportersMu for less time.
 			exportersMu.Lock()
-			for e := range exporters {
+			for e, defaultAttrs := range exporters {
+				if len(defaultAttrs) > 0 && sd.Attributes == nil {
+					sd.Attributes = make(map[string]interface{})
+				}
+				for _, attr := range defaultAttrs {
+					sd.Attributes[attr.key] = attr.value
+				}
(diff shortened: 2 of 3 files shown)


trace/trace.go
```diff
@@ -278,7 +278,13 @@ func (s *Span) End() {
 		if s.spanContext.IsSampled() {
 			// TODO: consider holding exportersMu for less time.
 			exportersMu.Lock()
-			for e := range exporters {
+			for e, defaultAttrs := range exporters {
+				if len(defaultAttrs) > 0 && sd.Attributes == nil {
+					sd.Attributes = make(map[string]interface{})
+				}
+				for _, attr := range defaultAttrs {
+					sd.Attributes[attr.key] = attr.value
```
ramonza: Looks like the same SpanData is re-used when exporting to multiple exporters with the result that the second exporter gets all the attributes from the first, etc. 

I think we should either only support a single set of attributes, or else we need to clone the SpanData for each exporter.

trace/export.go
```diff
@@ -32,19 +32,20 @@ type Exporter interface {
...
 )
 
 // RegisterExporter adds to the list of Exporters that will receive sampled
-// trace spans.
+// trace spans. Passed attributes will be appended to every span that is
+// exported by this exporter.
 //
 // Binaries can register exporters, libraries shouldn't register exporters.
-func RegisterExporter(e Exporter) {
+func RegisterExporter(e Exporter, a ...Attribute) {
```
ramonza: How about using the functional options pattern here instead? So it becomes something like:

`RegisterExporter(e, trace.WithDefaultAttributes(...))`

This keeps our options open w.r.t. adding more arguments in future.

trace/trace.go
gottwald: True, missed that.
Tbh I don't see a real use case in having different attributes on different exporters.
If others don't have one as well, then I'd suggest that we decouple it from the exporter.

We could create a `DefaultAttributes(a ...Attributes)` func that sets the default attrs on the package and then use them in `makeSpanData()` when the `Attributes` map is created.
This would allow for a more expected behavior when attribute keys collide: Default attrs would then be overridden by span attrs instead of the other way round.

WDYT?

gottwald:
Where do you see the benefit in keeping it out of the core API? Besides the already mentioned code duplication in all exporters, it would come with a risk that certain exporters might choose not to implement the feature which would result in different feature sets on different exporters, wouldn't it?

basvanbeek:
I agree with @rakyll we should not pollute the core APIs. 

One reason being that the tracing backends they end up in have different informational requirements and can use different attribute keys for the same information payload.

We're operating in a brown field situation where being idiomatic to the exported ecosystems has value. This also helps an instrumentation consumer switching from one tracing backend to the other see that they have to explicitly take care of migrating the set default attributes to the new exporter (and potentially alter them to be idiomatic to this new platform).

gottwald:
gotcha.
I added another PR census-ecosystem/opencensus-go-exporter-stackdriver#2 which implements what you suggested in the stackdriver exporter.
But it's a bit cumbersome to use, since I can't really work with the `trace.Attribute` type there because in order to add them to the `SpanData.Attributes` I need access to the `key` and `value` which are both private fields of the `trace.Attribute`.
In the end it works out fine but it just looks different from using normal `trace.Attributes`

ramonza:
Replaced by: census-ecosystem/opencensus-go-exporter-stackdriver#2

PR #751, opened 2018-05-16
Update usage of runtime/trace API

hyangah:
runtime/trace.NewContext was replaced by runtime/trace.NewTask
Fixes broken build with tip


rakyll:
Thanks!

PR #755, opened 2018-05-21
Add linkerd trace propagation for ochttp

### LINKED ISSUE #752: Would you like this linkerd-flavoured-Zipkin trace propagator?
https://github.com/negz/linkin

👋 Is there any interest in adding the above linkerd-flavoured-Zipkin-propagation package to Opencensus proper?

As I'm sure you're aware [linkerd](http://linkerd.io/) is a popular service mesh. It offers semi-automatic Zipkin tracing, but propagates trace context via a linkerd-specific HTTP header (`l5d-ctx-trace`) in order to avoid interfering with any existing `X-B3-*` headers.

linkerd requires the services that leverage its proxy mesh to forward the `l5d-ctx-trace` from parent requests to any children they spawn in order to form a full trace. Opencensus's API makes it a great fit for handling that propagation, as well as for creating in-application spans that don't relate to HTTP calls.

I've put this together for use at [Planet](http://planet.com), but I figure it may be useful to others. Note that per https://github.com/linkerd/linkerd/issues/1447 I'm not _really_ sure we're supposed to consume or modify the `l5d-ctx-trace` header, but it's the only option for the time being.

negz:
This propagation plugin allows Opencensus to extend and propagate linkerd traces.

[linkerd](https://linkerd.io/) is a popular service mesh. linkerd [supports Zipkin tracing](https://linkerd.io/config/1.4.0/linkerd/index.html#zipkin-telemeter); it will automatically create spans for all requests forwarded via linkerd. While linkerd uses Zipkin tracing it uses a [linkerd-specific HTTP header](https://linkerd.io/config/1.4.0/linkerd/index.html#http-headers) (`l5d-ctx-trace`) to propagate trace data rather than the regular `X-B3-*` headers. My understanding per https://github.com/linkerd/linkerd/issues/1447 is that linkerd uses this custom header in order to avoid linkerd traces potentially interfering with existing Zipkin traces.

At [Planet](http://planet.com) we're using linkerd because in the simplest case application code can avoid 'real' instrumentation and instead just forward the `l5d-ctx-trace` header from incoming requests to any outgoing requests they spawn. linkerd together with Opencensus's great API allows us to go one step further and emit spans from our application code.

I've tested this pretty thoroughly in our somewhat convoluted trace setup. Specifically:
* linkerd in linker-to-linker mode, i.e. both incoming and outgoing requests being routed via linkerd.
* linkerd running the [`io.zipkin.http`](https://github.com/linkerd/linkerd-zipkin/blob/master/src/main/scala/io/buoyant/linkerd/zipkin/HttpInitializer.scala) plugin, which [sends traces to the Zipkin v1](https://github.com/openzipkin/zipkin-finagle/blob/36ad33b/http/src/main/java/zipkin/finagle/http/HttpSender.java#L63) HTTP API.
* [zipkin-gcp](https://github.com/openzipkin/zipkin-gcp) forwarding traces from both linkerd and Opencensus to Stackdriver Trace.

It's worth noting that per https://github.com/linkerd/linkerd/issues/1447 `l5d-ctx-trace` wasn't _really_ designed for inspection and manipulation; the idea is that applications just forward it blindly. There's talk of adding some user-facing trace headers to linkerd, but no implementation yet. I'd be happy to update this propagation plugin if/when said user-facing headers are added to linkerd.

Fixes #752.

### DIFF
--- a/plugin/ochttp/propagation/linkerd/linkerd.go
+++ b/plugin/ochttp/propagation/linkerd/linkerd.go
@@ -1,23 +1,14 @@
-// Package linkin provides linkerd trace propagation for Opencensus.
-//
-// Opencensus is a single distribution of libraries that automatically collects
-// traces and metrics from your app, displays them locally, and sends them to
-// any analysis tool. Opencensus supports the Zipkin request tracing system.
-//
-// Zipkin is a popular distributed tracing system, allowing requests to be
-// traced through a distributed system. A request is broken into 'spans', each
-// representing one part of the greater request path. Software that wishes to
-// leverage Zipkin must be instrumented to do so (for example via Opencensus).
+// Package linkerd provides linkerd-flavoured Zipkin trace propagation.
 //
 // linkerd is a popular service mesh. One of linkerd's selling points is that it
 // provides Zipkin request tracing 'for free'. Software need not be 'fully'
 // instrumented, and instead need only copy linkerd's l5d-ctx-* HTTP headers
 // from incoming HTTP requests to any outgoing HTTP requests they spawn.
 //
-// Unfortunately while linkerd emits traces to Zipkin, it propagates trace data
-// via a non-standard header. This package may be used as a drop-in replacement
-// for https://godoc.org/go.opencensus.io/plugin/ochttp/propagation/b3 in
-// environments that use linkerd for part or all of their request tracing needs.
+// linkerd propagates trace data via the l5d-ctx-trace HTTP header rather than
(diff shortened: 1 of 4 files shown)


negz:
I signed it!

negz:
@ramonza @rakyll There seems to be a difference of opinion as to whether this package should live. I don't have strong feelings either way; let me know if I should close this and (if so) how to move it to the ecosystem org.

negz:
Closing this per comments in #752.

PR #761, opened 2018-05-24
Fix internal doc comment on View.canonicalize

rogpeppe:
It canonicalizes in place rather than returning a new View.


rakyll:
Thanks much.

PR #769, opened 2018-05-30
Create datadog exporter [WIP]

ChristineTChen:


ramonza:
Hi @ChristineTChen welcome to the project! Really nice to see you're interested in making a DataDog exporter! Just a few thoughts on this:

We are trying to keep the exporters *out* of this base repo and instead have them in separate repos. This allows us to delegate permissions on these repos to maintainers (like yourself!). So I think the best way to do this would be to create a repo under your name or DataDogs's GitHub org and we can fork it into census-ecosystem. Of course, if you would prefer to keep the repo in DataDog's org that is absolutely fine too - we can just link to it from our README.

Second, on the nature of the exporter I believe @savaki tried the statsd path before (https://github.com/census-instrumentation/opencensus-go/pull/517). This isn't really a good fit for the OpenCensus model IMO because OpenCensus wants to do aggregation in-process and statsd also wants to do aggregation internally so you end up with both trying to aggregate and it doesn't really work.

What we thought might work better would be if we could submit *already aggregated* metrics from OpenCensus directly to the DataDog backend API - without going through statsd at all. Have you considered this approach?

ChristineTChen:
Hi @ramonza! I will create a repo under Datadog's Github org instead and make the exporter there. After speaking to my team, we will look into submitting already aggregated metrics from OpenCensus to Datadog via gauge in Dogstatsd. Submitting directly to the backend API will require users to store their API key. We are thinking the best option would be to continue to rely on Dogstatsd and submit these already aggregated metrics as a [gauge type](https://docs.datadoghq.com/developers/dogstatsd/#gauges) so they aren't double aggregated.

rakyll:
@ChristineTChen, thanks for this!

We also have users who need support for tracing. I'd love to contribute a tracing exporter when the repo under Datadog repo comes alive!

ramonza:
Closing this PR since we decided to put this in another repo

PR #770, opened 2018-05-31
Modify some small details in the Example

hbgxsm95:
update the example to make it more clear.  The helloworld example in the go version doesn't visualize the stat result compared to the java version. Moreover, create the zpage server for the client. (be familiar with the workflow

### DIFF
--- a/examples/grpc/helloworld_client/main.go
+++ b/examples/grpc/helloworld_client/main.go
@@ -25,6 +25,8 @@ import (
 	"go.opencensus.io/stats/view"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
+	"net/http"
+	"go.opencensus.io/zpages"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


examples/grpc/helloworld_client/main.go
```diff
@@ -33,6 +35,10 @@ const (
 )
 
 func main() {
+	go func() {
```
songy23: Did you prevent client from exiting after response is received? Same for HTTP client.

examples/helloworld/main.go
```diff
@@ -75,6 +76,8 @@ func main() {
 	// library reports the collected data.
 	fmt.Println("Wait longer than the reporting duration...")
 	time.Sleep(2 * time.Second)
+	retrievedData, _ := view.RetrieveData(newView.Name)
+	fmt.Printf("Collected Data: %v\n", retrievedData[0].Data)
```
songy23: I think `PrintExporter` already printed the output? (line 47)

examples/helloworld/main.go
```diff
@@ -58,13 +58,14 @@ func main() {
 	// Create view to see the processed video size
 	// distribution broken down by frontend.
 	// Register will allow view data to be exported.
-	if err := view.Register(&view.View{
+	newView := &view.View{
```
songy23: Please revert the change here, since `newView` is not used elsewhere.

Issue #772, opened 2018-06-01
Option for enabling / disabling zpages at startup time


toffaletti:
It is a bit surprising that just importing zpages package (even if you don't register a handler) causes the infinite memory growth described in #597 even if DefaultSampler is NeverSample because of 

```
func init() {
	internal.LocalSpanStoreEnabled = true
}
```
and
```
	if !internal.LocalSpanStoreEnabled && !span.spanContext.IsSampled() {
		return span
	}
```

This means enabling / disabling has to be done with a build tag right now, but a runtime command line flag would be best.

ramonza:
Agree this would be better enabled at runtime.

odeke-em:
Thank you for filing this @toffaletti and thanks @ramonza for working on it!

@ramonza do you think we should remove that init invocation that enables zpages on the zpages.Handler variable? What else could we do here? Thank you.

songy23:
FYI in Java ZPages needs to be explicitly initialized in application code:
https://github.com/census-instrumentation/opencensus-java/blob/9d533927c764069641ed56dcdcd0a83d05a5a14a/examples/src/main/java/io/opencensus/examples/zpages/ZPagesTester.java#L105

ramonza:
I think we can go ahead now and remove the Handler var that causes eager init.

PR #774, opened 2018-06-04
Lazily enable zpages at runtime

ramonza:
See: #772

### DIFF
--- a/examples/grpc/helloworld_server/main.go
+++ b/examples/grpc/helloworld_server/main.go
@@ -49,7 +49,8 @@ func (s *server) SayHello(ctx context.Context, in *pb.HelloRequest) (*pb.HelloRe
 func main() {
 	// Start z-Pages server.
 	go func() {
-		mux := zpages.AppendHandlers("/debug", http.NewServeMux())
+		mux := http.NewServeMux()
+		zpages.Handle(mux, "/debug")
 		log.Fatal(http.ListenAndServe("127.0.0.1:8081", mux))
 	}()
 
--- a/examples/grpc/helloworld_server/main.go
+++ b/examples/grpc/helloworld_server/main.go
@@ -49,7 +49,8 @@ func (s *server) SayHello(ctx context.Context, in *pb.HelloRequest) (*pb.HelloRe
 func main() {
 	// Start z-Pages server.
 	go func() {
-		log.Fatal(http.ListenAndServe("127.0.0.1:8081", zpages.NewHandler("/debug")))
+		mux := zpages.AppendHandlers("/debug", http.NewServeMux())
+		log.Fatal(http.ListenAndServe("127.0.0.1:8081", mux))
 	}()
 
 	// Register stats and trace exporters to export
--- a/examples/grpc/helloworld_server/main.go
+++ b/examples/grpc/helloworld_server/main.go
@@ -47,10 +47,11 @@ func (s *server) SayHello(ctx context.Context, in *pb.HelloRequest) (*pb.HelloRe
 }
 
 func main() {
+	// Start z-Pages server.
 	go func() {
-		http.Handle("/debug/", http.StripPrefix("/debug", zpages.Handler))
-		log.Fatal(http.ListenAndServe(":8081", nil))
+		log.Fatal(http.ListenAndServe("127.0.0.1:8081", zpages.NewHandler("/debug")))
 	}()
+
 	// Register stats and trace exporters to export
 	// the collected data.
 	view.RegisterExporter(&exporter.PrintExporter{})
--- a/zpages/zpages.go
+++ b/zpages/zpages.go
@@ -31,32 +31,41 @@
 package zpages // import "go.opencensus.io/zpages"
 
 import (
-	"fmt"
 	"net/http"
 	"sync"
+
+	"path"
 
 	"go.opencensus.io/internal"
 )
 
-// Handler is deprecated: Use AppendHandlers.
-var Handler http.Handler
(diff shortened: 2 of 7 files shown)


zpages/zpages.go
```diff
@@ -31,16 +31,40 @@
...
 
 import (
+	"fmt"
+	"log"
+	"net"
 	"net/http"
 )
 
-// Handler is an http.Handler that serves the zpages.
+// Handler is deprecated: Use NewHandler.
```
ramonza: Once this is removed according to the deprecation policy, we can move the init logic into NewHandler.

examples/grpc/helloworld_server/main.go
```diff
@@ -31,6 +30,7 @@ import (
 	"go.opencensus.io/zpages"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
+	"net/http"
```
rakyll: Revert this change.

zpages/zpages.go
```diff
@@ -31,16 +31,31 @@
...
 var Handler http.Handler
 
+var enableOnce sync.Once
+
 func init() {
+	Handler = NewHandler("")
+}
+
+// GetHandler returns a handler that serves the z-pages.
```
rakyll: NewHandler

zpages/zpages.go
```diff
@@ -31,16 +31,31 @@
...
+func NewHandler(prefix string) http.Handler {
```
rakyll: We shouldn't have the prefix here.

It is coupling the responsibilities of a mux and a handler into a handler.

ramonza: I this this is preferable to using StripPrefix. It is much more readable, does not allocate a new request object, and does not try to fake the URL (like StripPrefix does).

ramonza:
Thanks @rakyll updated. I prefer the approach with taking a prefix as explained above and would like to keep it unless you feel strongly that it's wrong. After looking at what StripPrefix does, I don't think it's a good idea to recommend its use here if there's an straightforward way to avoid it.

ramonza:
So one way to fix this would be to not care what the prefix is and just match based on the last component of the URL. What do you think about that? Something like:

```go
parts := path.Split(req.URL)
page := parts[len(parts-1)]
switch page {
case "rpcz": 
...
default:
return http.NotFoundHandler().Handle(...)
}
```

Then we wouldn't need to take a prefix parameter at all and it would work regardless of the path it is installed at?

zpages/zpages.go
```diff
@@ -31,16 +31,31 @@
...
+	"fmt"
 	"net/http"
+	"sync"
+
+	"go.opencensus.io/internal"
 )
 
-// Handler is an http.Handler that serves the zpages.
+// Handler is deprecated: Use NewHandler.
 var Handler http.Handler
```
nfisher: If this were changed to http.ServerMux I think that would work for what I referenced in #783. The http.Handler strips the `.Handle*` functions and while I could simply cast this to a ServerMux it feels dirty as I'm relying on knowledge of the implementation detail rather than a public contract.

zpages/zpages.go
```diff
@@ -31,16 +31,32 @@
...
-	zpagesMux.HandleFunc("/tracez", tracezHandler)
-	zpagesMux.Handle("/public/", http.FileServer(fs))
-	Handler = zpagesMux
+	mux := http.NewServeMux()
+	AppendHandlers("", mux)
+	Handler = mux
+}
+
+// AppendHandlers adds the z-pages to the given ServeMux rooted at pathPrefix.
+func AppendHandlers(pathPrefix string, mux *http.ServeMux) *http.ServeMux {
```
nfisher: Since it's returning a `*ServeMux` and there's no '/' specified on the mux it's probably ok to drop the mux parameter and create it in the method. `/public` is a fs folder so additional assets could be added at run-time and (famous last words) hopefully no one would want sub-paths under `/tracez` and `/rpcz`.

What are peoples thoughts on two functions?

1. `func zpages.NewPrefixed(pathPrefix string) *ServeMux`
2. `func zpages.New() *ServeMux`

The second basically calls the first with "" and would behaviour similar to the original `zpages.Handler`.

examples/grpc/helloworld_server/main.go
```diff
@@ -47,10 +47,12 @@ func (s *server) SayHello(ctx context.Context, in *pb.HelloRequest) (*pb.HelloRe
 }
 
 func main() {
+	// Start z-Pages server.
 	go func() {
-		http.Handle("/debug/", http.StripPrefix("/debug", zpages.Handler))
-		log.Fatal(http.ListenAndServe(":8081", nil))
+		mux := zpages.AppendHandlers("/debug", http.NewServeMux())
+		log.Fatal(http.ListenAndServe("127.0.0.1:8081", mux))
```
nfisher: Like the change to 127.0.0.1 as it doesn't trigger FW notifications on OS X. 👍

zpages/zpages.go
ramonza: I think constructing our own ServeMux reduces the flexibility, in particular to add to the DefaultServeMux. Renamed the method to Handle and no longer return the ServeMux to avoid ambiguity. 

I don't think we need a separate non-prefixed call. The user can just pass an empty string explicitly, I don't think it's worth adding a convenience method because I would think that the prefixed form would actually be more common in practice.

ramonza:
@rakyll are you ok with the new interface?

rakyll:
LGTM

Issue #773, opened 2018-06-03
Mechanism to explicitly flush stats to exporters


arjantop:
Currently I see no way to push collected metrics to prometheus before program exit (especially in short batch programs), since views are reported periodically based on a timer.

What I have currently done is set `view.SetReportingPeriod` to something low and wait some duration and then perform the push.

ramonza:
Prometheus natively doesn't really support this using the pull model. What seems like it would work would be to provide OpenCensus support for the Prometheus Push Gateway: https://github.com/prometheus/pushgateway.

Mind if we repurpose this as a feature request on the Prometheus exporter?

arjantop:
You can currently push using the prometheus api (combined with opencensus). But if we get "native" push support in opencensus even better.

What I am doing currently:
* set up exporter with provided registry
* flush views lowering the `view.SetReportingPeriod` (a hack)
* user prometheus api to push

ramonza:
Curious how you are pushing using the prometheus API. As I understand it, you register the exporter as an http.Handler and Prometheus scrapes it periodically. There isn't really a way to "force" Prometheus to scrape your app in a timely fashion. Sure, you can use SetReportingPeriod but all this does is change how often the page that Prometheus scrapes is updated. It doesn't guarantee that Prometheus will actually get the updated data.

arjantop:
No, I do not register the endpoint, opencensus Prometheus exporter is just an interface between opencensus views and native prometheus registry. I am using the `push` api that is using the gateway.

So I pass in explicit registry in `NewExporter`: `prometheus.Options{Registry: registry}`

And then use the `push` api: `push.FromGatherer(job, nil, promGateway, registry)`

ramonza:
Ah I see. I didn't realize that was possible. I will keep this open to add an example and document that usage.

I will also retitle to be closer to your original issue, namely that we need a way to flush exporters at process termination.

songy23:
Duplicated of https://github.com/census-instrumentation/opencensus-go/issues/862.

PR #778, opened 2018-06-05
Make the duration longer in order to export the stats

hbgxsm95:
Make the duration longer in order to export the stats by printing it out.  The previous branch is contaminated that I delete it directly.

### DIFF
--- a/examples/helloworld/main.go
+++ b/examples/helloworld/main.go
@@ -74,7 +74,7 @@ func main() {
 	// Wait for a duration longer than reporting duration to ensure the stats
 	// library reports the collected data.
 	fmt.Println("Wait longer than the reporting duration...")
-	time.Sleep(2 * time.Second)
+	time.Sleep(10 * time.Second)
 }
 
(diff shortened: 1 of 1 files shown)


examples/helloworld/main.go
```diff
@@ -74,7 +74,7 @@ func main() {
 	// Wait for a duration longer than reporting duration to ensure the stats
 	// library reports the collected data.
 	fmt.Println("Wait longer than the reporting duration...")
-	time.Sleep(2 * time.Second)
+	time.Sleep(10 * time.Second)
```
rakyll: Can we instead make the reporting period shorter? It is not nice for a user to wait for 10 seconds on an example.

hbgxsm95: It sounds to be a better way.  I would have a try.

songy23: SGTM. We can re-store to the previous reporting period once https://github.com/census-instrumentation/opencensus-go/issues/773 is resolved.

PR #781, opened 2018-06-06
Move version accessor to the opencensus package

### LINKED ISSUE #779: Move go.opencensus.io/exportutil to go.opencensus.io/export
There is no export package. We can instead put the Version under export package.

OTHO, we should make Version an accessor, it is immutable.

```
func Version() string
```

/cc @ramonza

rakyll:
Version returns the OpenCensus library version. It is not an exporter utility.
If it is an exporter utility, it shouldnt be in exportutil. In Go, utility packages
are organized under an existing package.

For example, io/iotuil or net/http/httputil.

This API should not yet be picked by anyone else other than the exporters.

Fixes #779.


exporterutil/version.go
```diff
@@ -1,20 +0,0 @@
...
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-// Package exporterutil contains common utilities for exporter implementations.
-package exporterutil
-
-// Version is the current release version of OpenCensus in use. It is made
-// available for exporters to include in User-Agent-like metadata.
-var Version = "0.13.0"
```
ramonza: This is already in a released version unfortunately, we need to deprecate and remove according to the policy.

ramonza:
Even though it is only used by the Stackdriver exporter, both are at released versions so it requires a coordinated removal. I recommend we just follow the deprecation policy and remove in a month.

rakyll:
I reverted the breaking change and deprecated the package. PTAL.

PR #787, opened 2018-06-13
Documentation cleanup

ramonza:


plugin/ochttp/stats.go
```diff
@@ -60,8 +68,16 @@ var (
 	DefaultLatencyDistribution = view.Distribution(0, 1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1000, 2000, 5000, 10000, 20000, 50000, 100000)
 )
 
-// Package ochttp provides some convenience views.
-// You need to subscribe to the views for data to actually be collected.
+// Convenience HTTP client views.
```
rakyll: Predefined HTTP client views.

There are not just convenience, we think these views are essential.

plugin/ochttp/stats.go
```diff
@@ -60,8 +68,16 @@ var (
...
+//
+// In many cases, you will want to create your own views over the measures
+// provided by this package, but for simple programs these convenience views
```
rakyll: "but for simple programs these convenience views"

This is misleading. We think these views are essentially important and they are not just for simple programs.

plugin/ochttp/stats.go
```diff
@@ -106,7 +122,19 @@ var (
 		Measure:     ClientLatency,
 		Aggregation: view.Count(),
 	}
+)
 
+// Convenience HTTP server views.
```
rakyll: Ditto.

plugin/ochttp/stats.go
```diff
@@ -152,7 +180,8 @@ var (
 	}
 )
 
-// DefaultClientViews are the default client views provided by this package.
+// DefaultClientViews are the default client views provided by this package,
+// provided for easy mass-registration.
```
rakyll: This is odd wording "provided for easy mass-registration".

I'd rather say "DefaultClientViews are the default client views to register for HTTP clients."

plugin/ochttp/stats.go
```diff
@@ -162,7 +191,8 @@ var DefaultClientViews = []*view.View{
 	ClientResponseCountByStatusCode,
 }
 
-// DefaultServerViews are the default server views provided by this package.
+// DefaultServerViews are the default server views provided by this package,
```
rakyll: ditto

plugin/ochttp/trace.go
```diff
@@ -146,7 +146,7 @@ func responseAttrs(resp *http.Response) []trace.Attribute {
 	}
 }
 
-// TraceStatus converts the HTTP status code to a trace.Status that
+// TraceStatus is a utility to convert the HTTP status code to a trace.Status that
```
rakyll: TraceStatus converts the HTTP status code.

Whether it is a utility or not is irrelevant.

stats/measure.go
```diff
@@ -20,19 +20,33 @@ import (
...
+// the measure needs to be used in a View. If no Views are defined over a
+// measure, there is very little cost in recording it.
 type Measure interface {
+	// Name returns the name of this measure.
+	//
+	// Measure names are globally unique (among all libraries linked into your program).
+	// We recommend prefixing the measure name with a domain name relevant to your
+	// project or application.
+	//
+	// Measure names can be as long as you like, they are
```
rakyll: Don't document if there is no restriction. You'd only document where user needs to know about a restriction.

Remove "Measure names can be as long as you like, they are..." sentence.

stats/measure.go
```diff
@@ -20,19 +20,33 @@ import (
...
+	// never sent over the wire or exported to backends in any way. They are only
+	// used to create Views.
 	Name() string
+
+	// Description returns the human-readable description of this measure.
+	// For documentation only.
```
rakyll: For documentation of what? This is very confusing, please remove.

stats/measure_float64.go
```diff
@@ -45,7 +45,9 @@ func (m *Float64Measure) M(v float64) Measurement {
 }
 
 // Float64 creates a new measure of type Float64Measure.
-// It never returns an error.
+//
+// See the documentation for interface Measure for more guidance on the
```
rakyll: We should rather say "Float64 returns a Measure of type float64".

stats/measure_int64.go
```diff
@@ -45,7 +45,9 @@ func (m *Int64Measure) M(v int64) Measurement {
 }
 
 // Int64 creates a new measure of type Int64Measure.
-// It never returns an error.
+//
+// See the documentation for interface Measure for more guidance on the
```
rakyll: Ditto

opencensus.go
```diff
@@ -17,5 +17,5 @@ package opencensus // import "go.opencensus.io"
 
 // Version is the current release version of OpenCensus in use.
 func Version() string {
-	return "0.13.0"
+	return "0.14.0"
```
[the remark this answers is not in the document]
ramonza: The test fails if I don't have this in this PR. I prefer to just merge this without squashing so that they still appear as two commits. Is that okay?

plugin/ochttp/stats.go
ramonza: There seems to be disagreement about this. I am going to remove the controversial parts to a separate PR.

plugin/ochttp/trace.go
ramonza: Looking through docs, it's unclear why this is there. My saying it's a utility, I intend to deflect attention off of them function as something that most users won't be interested in. Do you want to suggest another way of saying this?

stats/measure.go
ramonza: I'll remove the part about being as long as you like. I think it's worth making it clear that they aren't sent to backends.

stats/measure.go
ramonza: What would you suggest replacing it with? If it doesn't exist for documentation purposes, what does is it for?

PR #792, opened 2018-06-15
Deprecate public symbols individually in Stackdriver exporter

ramonza:
Some tools may not prominently display the package-level deprecation.
To improve the changes that users will notice the deprecation,
deprecate the public types and functions as well.

### DIFF
--- a/exporter/stackdriver/stackdriver.go
+++ b/exporter/stackdriver/stackdriver.go
@@ -33,6 +33,8 @@ import (
 )
 
 // Options contains options for configuring the exporter.
+//
+// Deprecated: This package has been moved to: contrib.go.opencensus.io/exporter/stackdriver.
 type Options struct {
 	// ProjectID is the identifier of the Stackdriver
 	// project the user is uploading the stats data to.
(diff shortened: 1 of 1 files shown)


ramonza:
That would violate the style guideline to start the doc with the name of the function.

odeke-em:
Roger that, thanks!

exporter/stackdriver/stackdriver.go
```diff
@@ -33,6 +33,7 @@ import (
 )
 
 // Options contains options for configuring the exporter.
+// Deprecated: This package has been moved to: contrib.go.opencensus.io/exporter/stackdriver.
```
rakyll: Add a line in between so the deprecation notice starts at its own paragraph. 

// Options contains options for configuring the exporter.
//
// Deprecated: This package has been moved to: contrib.go.opencensus.io/exporter/stackdriver.

exporter/stackdriver/stackdriver.go
```diff
@@ -81,13 +82,15 @@ type Options struct {
 
 // Exporter is a stats.Exporter and trace.Exporter
 // implementation that uploads data to Stackdriver.
+// Deprecated: This package has been moved to: contrib.go.opencensus.io/exporter/stackdriver.
```
rakyll: ditto

exporter/stackdriver/stackdriver.go
```diff
@@ -81,13 +82,15 @@ type Options struct {
...
 type Exporter struct {
 	traceExporter *traceExporter
 	statsExporter *statsExporter
 }
 
 // NewExporter creates a new Exporter that implements both stats.Exporter and
 // trace.Exporter.
+// Deprecated: This package has been moved to: contrib.go.opencensus.io/exporter/stackdriver.
```
rakyll: ditto

PR #793, opened 2018-06-18
Do not exit if request fails

rakyll:
This allows us to report the unsuccessful outgoing calls
rather than killing the process.

Also, logs should be proper sentences, they are
capitalized.


rakyll:
Thanks for the blazing fast review :)

Issue #663, opened 2018-04-02
ochttp should provide a way to pass route info


rozaliev:
According to the [spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md) the path should be used as a span name only if route can't be determined.

I haven't found any way pass route name,  which is unfortunate for REST-like apis. 

I imagine something like this for api: 
```go
// ochttp.Handler
type Handler struct {
   ...
   RouteName string
   ...
}
```

rakyll:
I wished that mux had a way to report the matched route. Putting this on the Handler as a field might not be ideal because it will require to initiate a new Handler for each route even though our primary recommendation is to wrap the entire mux with it.

ramonza:
Maybe what we need is a way to override the Span name of an active Span:

    trace.FromContext(req.Context()).SetName("...")

basvanbeek:
Gorilla's Mux allows for pulling route information from Request Context. Here's an example of a Go kit serverBefore handler using this i.c.m. zipkin-go and overriding the default span name.

```go
func ZipkinHTTPRoute(ctx context.Context, r *http.Request) context.Context {
	if span := zipkin.SpanFromContext(ctx); span != nil {
		if route := mux.CurrentRoute(r); route != nil {
			if tpl, err := route.GetPathTemplate(); err == nil {
				zipkin.TagHTTPRoute.Set(span, tpl)
				span.SetName(r.Method + " " + tpl)
			}
		}
	}
	return ctx
}
```

ramonza:
+1 for FormatSpanName callback in Handler

ramonza:
Fixed by #796

PR #796, opened 2018-06-19
Span naming flexibility

basvanbeek:
This PR enables more flexibility in span naming for various use cases.

span.SetName() allows one to update the span name after creation time. This allows routing middlewares to update the span name to the matched route providing lower cardinality then path names (think REST services).

ochttp client and server handlers now allow for the span name to be generated from pluggable functions defaulting to the already existing spanNameFromURL.  This can be used if span naming conventions in a brown field deployment are different from OpenCensus defaults or if using URL constructors for REST services that can also report route templates instead of expanded URL path.

### DIFF
--- a/plugin/ochttp/client.go
+++ b/plugin/ochttp/client.go
@@ -43,6 +43,11 @@ type Transport struct {
 	// for spans started by this transport.
 	StartOptions trace.StartOptions
 
+	// NameFromRequest holds the function to use for generating the span name
+	// from the information found in the outgoing HTTP Request. By default the
+	// name equals the URL Path.
+	NameFromRequest func(*http.Request) string
+
 	// TODO: Implement tag propagation for HTTP.
 }
 
@@ -54,6 +59,10 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	if format == nil {
 		format = defaultFormat
 	}
+	nameFromRequest := t.NameFromRequest
+	if nameFromRequest == nil {
+		nameFromRequest = spanNameFromURL
+	}
 	rt = &traceTransport{
 		base:   rt,
 		format: format,
@@ -61,6 +70,7 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 			Sampler:  t.StartOptions.Sampler,
 			SpanKind: trace.SpanKindClient,
 		},
+		nameFromRequest: nameFromRequest,
 	}
 	rt = statsTransport{base: rt}
 	return rt.RoundTrip(req)
--- a/plugin/ochttp/client.go
+++ b/plugin/ochttp/client.go
@@ -46,7 +46,7 @@ type Transport struct {
 	// NameFromRequest holds the function to use for generating the span name
 	// from the information found in the outgoing HTTP Request. By default the
 	// name equals the URL Path.
-	NameFromRequest func(*http.Request) string
+	FormatSpanName func(*http.Request) string
 
 	// TODO: Implement tag propagation for HTTP.
 }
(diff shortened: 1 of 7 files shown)


plugin/ochttp/server.go
```diff
@@ -60,6 +60,11 @@ type Handler struct {
 	// be added as a linked trace instead of being added as a parent of the
 	// current trace.
 	IsPublicEndpoint bool
+
+	// NameFromRequest holds the function to use for generating the span name
+	// from the information found in the incoming HTTP Request. By default the
+	// name equals the URL Path.
+	NameFromRequest func(*http.Request) string
```
ramonza: I think we should remove this. Seems like anything you can accomplish with NameFromRequest you can always accomplish by installing a middleware that calls SetName. 

And I think encouraging the use of SetName is actually preferable, let me explain. We want to encourage users to install ochttp.Handler as the outer-most handler (so we measure as much as possible). However, the proper span name is only available after the routing middleware has executed, hence the need for changing it later. If we provide NameFromRequest, it may encourage users to install ochttp.Handler *after* the routing middleware in order to capture the routing key. This would be bad because we want to measure as much of the application processing time as possible (including the routing middleware execution).

plugin/ochttp/client.go
```diff
@@ -43,6 +43,11 @@ type Transport struct {
 	// for spans started by this transport.
 	StartOptions trace.StartOptions
 
+	// NameFromRequest holds the function to use for generating the span name
+	// from the information found in the outgoing HTTP Request. By default the
+	// name equals the URL Path.
+	NameFromRequest func(*http.Request) string
```
ramonza: Could we have a unit test for this new functionality?

trace/trace_test.go
```diff
@@ -446,6 +446,30 @@ func TestMessageEvents(t *testing.T) {
...
+			TraceID:      tid,
+			SpanID:       SpanID{},
+			TraceOptions: 0x1,
+		},
+		ParentSpanID:    sid,
+		Name:            "NewName",
+		Status:          Status{},
+		HasRemoteParent: true,
+	}
+	if !reflect.DeepEqual(got, want) {
```
ramonza: Just check the name here to make it clear what you are testing.

ramonza: Also would be good to test what happens if you call SetName when not collecting events and also after the Span has ended.

plugin/ochttp/server.go
basvanbeek: I appreciate the concern with respect to potential misuse in order to capture routing details and yes SetName can overwrite the span name to the preferred name for all use cases.

Unfortunately that comes at a cost for people that want to use a different naming scheme which can be identified from the upfront raw http request details. As an example, people in polyglot brownfield environments where the name of a call is preferred to be low cardinality like`http:post` because the used backend due to indexing really dislikes high cardinality path based naming.

By removing this feature from the PR we now force them to add another middleware layer in their service just to update the span name and incur the cost of the lock when updating Span.data with SetName.
 
I suggest keeping this functionality but being very clear about how ochttp should be used when composing the multi handler flow. I think this mitigates your concerns enough to warrant keeping this functionality.

plugin/ochttp/client.go
basvanbeek: sure

trace/trace_test.go
basvanbeek: sure

plugin/ochttp/server.go
ramonza: Alright. If it were me, I would probably err on the side of having only one way to do things but I can see your points too so I happy to merge.

ramonza:
I just found this related issue, not sure if you were aware of it: https://github.com/census-instrumentation/opencensus-go/issues/663

In that issue @rakyll suggests calling the method `FormatSpanName`, which I think is a nicer name than `NameFromRequest`. @basvanbeek would you consider changing this field name?

PR #800, opened 2018-06-21
Fix stats/view.View comment

mvndaai:
The comment said to use the Subscribe function when it should said Register


mvndaai:
I signed it!

mvndaai:
@songy23, @ramonza thank you for the super rapid response!

PR #801, opened 2018-06-21
Do not enforce singleton property of Prometheus exporter

ramonza:
It is unnecessarily defensive to enforce that only a single
instance of the Prometheus exporter is created.

Doing this gets in the way of using the exporter in systems where
exporters are created dynamically at runtime based on configuration,
such as Istio.

### DIFF
--- a/exporter/prometheus/prometheus.go
+++ b/exporter/prometheus/prometheus.go
@@ -12,14 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package prometheus contains a Prometheus exporter.
-//
-// Please note that this exporter is currently work in progress and not complete.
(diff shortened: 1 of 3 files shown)


exporter/prometheus/prometheus.go
```diff
@@ -51,23 +50,9 @@ type Options struct {
...
 }
 
-var (
-	newExporterOnce      sync.Once
-	errSingletonExporter = errors.New("expecting only one exporter per instance")
-)
-
 // NewExporter returns an exporter that exports stats to Prometheus.
-// Only one exporter should exist per instance
+// Most users should only register a single Prometheus exporter per process.
```
rakyll: I'd not add this to the docs.

This is what the Prometheus community is already aware about, don't document it. Document things if there is something not expected.

PR #804, opened 2018-06-27
Allow setting the Zipkin LocalEndpoint per-span

ramonza:
This is needed for rare cases when the logical endpoint differs from
the process submitting the span, for example Istio.

### DIFF
--- a/exporter/zipkin/zipkin.go
+++ b/exporter/zipkin/zipkin.go
@@ -68,13 +68,13 @@ func (e *Exporter) ExportSpan(s *trace.SpanData) {
 }
 
 func (e *Exporter) spanLocalEndpoint(s *trace.SpanData) *model.Endpoint {
-	hostPoint, ok := s.Attributes[serviceEndpointKey].(string)
-	if !ok {
-		return e.localEndpoint
-	}
 	serviceName, ok := s.Attributes[serviceNameKey].(string)
 	if !ok {
 		return e.localEndpoint
+	}
+	hostPoint, ok := s.Attributes[serviceEndpointKey].(string)
+	if !ok {
+		hostPoint = defaultHostPort
 	}
 	ep, err := zipkin.NewEndpoint(serviceName, hostPoint)
 	if err != nil {
@@ -89,6 +89,8 @@ const (
 	statusDescriptionTagKey = "opencensus.status_description"
 	serviceNameKey          = "opencensus.service_name"
 	serviceEndpointKey      = "opencensus.service_endpoint"
+
+	defaultHostPort = "0.0.0.0:0"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


exporter/zipkin/zipkin.go
```diff
@@ -29,6 +33,11 @@ import (
 type Exporter struct {
 	reporter      reporter.Reporter
 	localEndpoint *model.Endpoint
+
+	// Logger will be used to report errors from this exporter.
+	Logger interface {
```
rakyll: Can we instead have hook like onError at https://godoc.org/contrib.go.opencensus.io/exporter/stackdriver#Options?

It is useful for those who wants to report the errors to a service.

exporter/zipkin/zipkin_integration_test.go
```diff
@@ -0,0 +1,31 @@
+package zipkin_test
```
rakyll: Missing license header.

exporter/zipkin/zipkin_test.go
```diff
@@ -18,11 +18,15 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
-	"reflect"
 	"strings"
 	"testing"
 	"time"
```
rakyll: Remove the line and gofmt.

trace/basetypes.go
```diff
@@ -50,16 +50,19 @@ type Attribute struct {
 }
 
 // BoolAttribute returns a bool-valued attribute.
+// Attribute keys starting with "opencensus" are reserved for internal use.
```
rakyll: Is this in the trace specs?

exporter/zipkin/zipkin.go
```diff
@@ -38,25 +47,50 @@ type Exporter struct {
 // can be created with the openzipkin library, using one of the packages under
 // github.com/openzipkin/zipkin-go/reporter.
 //
-// localEndpoint sets the local endpoint of exported spans.  It can be
+// localEndpoint sets the default local endpoint of exported spans.  It can be
 // constructed with github.com/openzipkin/zipkin-go.NewEndpoint, e.g.:
 // 	localEndpoint, err := NewEndpoint("my server", listener.Addr().String())
 // localEndpoint can be nil.
+//
+// localEndpoint can be overridden on a per-span basis by setting Span attributes
```
rakyll: This doesn't look like a Zipkin-only case. Can't we implement it more generally?

We can export a canonical attribute key from the trace package for the service name.

Most importantly, it needs to be reflected in the specs especially if we will preserve the opencensus prefix to OpenCensus libraries.

ramonza:
Going to take a different approach with Istio

PR #806, opened 2018-06-28
Add tag.DecodeEach to decode a serialized tag map

ramonza:
This can be used to parse the tags in header into a datastructure
other than tag.Map: it's useful for applications that want to read
the tags header but don't use the stats package.

### DIFF
--- a/tag/map_codec.go
+++ b/tag/map_codec.go
@@ -177,45 +177,58 @@ func Encode(m *Map) []byte {
 // Decode decodes the given []byte into a tag map.
 func Decode(bytes []byte) (*Map, error) {
 	ts := newMap()
+	err := DecodeEach(bytes, ts.upsert)
+	if err != nil {
+		// no partial failures
+		return nil, err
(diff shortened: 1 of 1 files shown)


rakyll:
Why don't we just add a Map iterator?

ramonza:
You mean to the tag.Map? IIRC there was some pushback against making tags
readable in general (from Bogdan I believe). There was a concern I think
that it would encourage storing arbitrary request-scoped values in tags.

rakyll:
What else we can do is to create a propagation package where we will have the encoders/decoders similar to the trace/propagation package.

Then, tag package can use the encoders/decoders there.

```
package propagation

func Encode(map[string]string) []byte
func Decode(buf []bytes) (map[string]string, error)
```

ramonza:
I will try this

rakyll:
Awesome. It would be consistent with the trace package and we can add the HTTP propagation interfaces over there too.

PR #808, opened 2018-06-28
Fix links to zpages in http example.

rghetia:
Added prefix 'debug' to fix the zpages link.


ramonza:
Looks like a problem with the Appveyor environment

Issue #401, opened 2018-02-02
Limit the number of annotations we buffer in SpanData


ramonza:
To avoid leaking memory on long-running spans, we should make sure SpanData has a fixed size regardless of how many annotations have been added. We should add a count of dropped annotations. This should apply to any of the variable sized fields in SpanData (message events, links, etc.).

rghetia:
fixed by #1020 and  #1022

PR #824, opened 2018-07-12
trace: add buffer limit

jgracenin:
This applies to all span data entities that are slices: Annotations,
Links, and Message Events.

At the limit, it pops the oldest sample off prior to appending the new
sample. This should not cause an additional allocation.

Each entity shares the same WithBufferLimit option following the YAGNI
principle.

Each test ensures 4 entities are added in total -- the first 2 are
dropped and the last 2 remain. This should prevent off by 1 bugs.

### DIFF
--- a/trace/export.go
+++ b/trace/export.go
@@ -73,4 +73,11 @@ type SpanData struct {
 	Status
 	Links           []Link
 	HasRemoteParent bool
+
+	// The count of all span.bufferLimit overflows.
+	// See trace.WithBufferLimit for more info.
+	// TODO: This is not currently exported anywhere (e.g. tracez page).
+	DroppedAnnotations   int
+	DroppedMessageEvents int
+	DroppedLinks         int
 }
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -46,6 +46,7 @@ type Span struct {
 	endOnce sync.Once
 
 	executionTracerTaskEnd func() // ends the execution tracer span
+	bufferLimit            int    // limits all variable span data (slices)
 }
 
 // IsRecordingEvents returns true if events are being recorded for this span.
@@ -110,6 +111,9 @@ const (
 	SpanKindClient
 )
 
+// DefaultBufferLimit is the default value for trace.StartOptions.BufferLimit.
+const DefaultBufferLimit = 1000
+
 // StartOptions contains options concerning how a span is started.
 type StartOptions struct {
 	// Sampler to consult for this Span. If provided, it is always consulted.
@@ -125,6 +129,11 @@ type StartOptions struct {
 	// SpanKind represents the kind of a span. If none is set,
 	// SpanKindUnspecified is used.
 	SpanKind int
+
+	// BufferLimit makes new spans with a custom buffer limit.
+	// This limits variable span data: Message Events, Links, and Annotations.
+	// The default is 1000 (trace.DefaultBufferLimit) and the minimum is 1.
+	BufferLimit int
 }
 
 // StartOption apply changes to StartOptions.
@@ -145,13 +154,25 @@ func WithSampler(sampler Sampler) StartOption {
 	}
 }
 
+// WithBufferLimit makes new spans with a custom buffer limit.
+// This limits variable trace span data: Message Events, Links, and Annotations.
+// The default is 1000 (trace.DefaultBufferLimit) and the minimum is 1.
+func WithBufferLimit(limit int) StartOption {
+	return func(o *StartOptions) {
+		o.BufferLimit = limit
+		if o.BufferLimit <= 0 {
+			o.BufferLimit = 1
+		}
+	}
+}
+
 // StartSpan starts a new child span of the current span in the context. If
 // there is no span in the context, creates a new trace and span.
 //
 // Returned context contains the newly created span. You can use it to
 // propagate the returned span in process.
 func StartSpan(ctx context.Context, name string, o ...StartOption) (context.Context, *Span) {
-	var opts StartOptions
+	opts := StartOptions{BufferLimit: DefaultBufferLimit}
 	var parent SpanContext
 	if p := FromContext(ctx); p != nil {
 		parent = p.spanContext
@@ -174,7 +195,7 @@ func StartSpan(ctx context.Context, name string, o ...StartOption) (context.Cont
 // Returned context contains the newly created span. You can use it to
 // propagate the returned span in process.
 func StartSpanWithRemoteParent(ctx context.Context, name string, parent SpanContext, o ...StartOption) (context.Context, *Span) {
-	var opts StartOptions
+	opts := StartOptions{BufferLimit: DefaultBufferLimit}
 	for _, op := range o {
 		op(&opts)
 	}
@@ -185,7 +206,7 @@ func StartSpanWithRemoteParent(ctx context.Context, name string, parent SpanCont
 }
 
 func startSpanInternal(name string, hasParent bool, parent SpanContext, remoteParent bool, o StartOptions) *Span {
-	span := &Span{}
+	span := &Span{bufferLimit: o.BufferLimit}
 	span.spanContext = parent
 
 	cfg := config.Load().(*Config)
@@ -340,6 +361,10 @@ func (s *Span) lazyPrintfInternal(attributes []Attribute, format string, a ...in
 		m = make(map[string]interface{})
 		copyAttributes(m, attributes)
 	}
+	if l := len(s.data.Annotations); l > 0 && l >= s.bufferLimit {
+		s.data.Annotations = s.data.Annotations[1:]
+		s.data.DroppedAnnotations++
+	}
 	s.data.Annotations = append(s.data.Annotations, Annotation{
 		Time:       now,
 		Message:    msg,
@@ -355,6 +380,10 @@ func (s *Span) printStringInternal(attributes []Attribute, str string) {
 	if len(attributes) != 0 {
 		a = make(map[string]interface{})
 		copyAttributes(a, attributes)
+	}
+	if l := len(s.data.Annotations); l > 0 && l >= s.bufferLimit {
+		s.data.Annotations = s.data.Annotations[1:]
+		s.data.DroppedAnnotations++
 	}
 	s.data.Annotations = append(s.data.Annotations, Annotation{
 		Time:       now,
@@ -393,6 +422,10 @@ func (s *Span) AddMessageSendEvent(messageID, uncompressedByteSize, compressedBy
 	}
 	now := time.Now()
 	s.mu.Lock()
+	if l := len(s.data.MessageEvents); l > 0 && l >= s.bufferLimit {
+		s.data.MessageEvents = s.data.MessageEvents[1:]
+		s.data.DroppedMessageEvents++
+	}
 	s.data.MessageEvents = append(s.data.MessageEvents, MessageEvent{
 		Time:                 now,
 		EventType:            MessageEventTypeSent,
@@ -415,6 +448,10 @@ func (s *Span) AddMessageReceiveEvent(messageID, uncompressedByteSize, compresse
 	}
 	now := time.Now()
 	s.mu.Lock()
+	if l := len(s.data.MessageEvents); l > 0 && l >= s.bufferLimit {
+		s.data.MessageEvents = s.data.MessageEvents[1:]
+		s.data.DroppedMessageEvents++
+	}
 	s.data.MessageEvents = append(s.data.MessageEvents, MessageEvent{
 		Time:                 now,
 		EventType:            MessageEventTypeRecv,
@@ -431,6 +468,10 @@ func (s *Span) AddLink(l Link) {
 		return
 	}
 	s.mu.Lock()
+	if l := len(s.data.Links); l > 0 && l >= s.bufferLimit {
+		s.data.Links = s.data.Links[1:]
+		s.data.DroppedLinks++
+	}
 	s.data.Links = append(s.data.Links, l)
 	s.mu.Unlock()
 }
--- a/trace/trace_test.go
+++ b/trace/trace_test.go
@@ -224,6 +224,10 @@ func TestStartSpanWithRemoteParent(t *testing.T) {
 
 // startSpan returns a context with a new Span that is recording events and will be exported.
 func startSpan(o StartOptions) *Span {
+	bufferLimit := DefaultBufferLimit
+	if o.BufferLimit != 0 {
+		bufferLimit = o.BufferLimit
+	}
 	_, span := StartSpanWithRemoteParent(context.Background(), "span0",
 		SpanContext{
 			TraceID:      tid,
@@ -232,6 +236,7 @@ func startSpan(o StartOptions) *Span {
 		},
 		WithSampler(o.Sampler),
 		WithSpanKind(o.SpanKind),
+		WithBufferLimit(bufferLimit),
 	)
 	return span
 }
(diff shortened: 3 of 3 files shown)


jgracenin:
Hey,

I had one service with a very long running stream and saw 15GB+ of message events in /debug/pprof/heap :) . I suppose I should be using gRPC's MAX_CONNECTION_AGE at some point: https://github.com/grpc/proposal/blob/master/A9-server-side-conn-mgt.md

I see that this already be in progress (#401) so feel free to close if that's the case.

trace/trace.go
```diff
@@ -125,6 +129,11 @@ type StartOptions struct {
 	// SpanKind represents the kind of a span. If none is set,
 	// SpanKindUnspecified is used.
 	SpanKind int
+
+	// BufferLimit makes new spans with a custom buffer limit.
+	// This limits variable span data: Message Events, Links, and Annotations.
+	// The default is 1000 (trace.DefaultBufferLimit) and the minimum is 1.
+	BufferLimit int
```
basvanbeek: As with the above mentioned upper limits I also think we should be careful with setting the bottom limit. Where it could be sensible for links to have a lower limit of 1 (or even 0) it probably is a bad idea to have one as low as this for annotations or message events.

trace/trace.go
```diff
@@ -340,6 +361,10 @@ func (s *Span) lazyPrintfInternal(attributes []Attribute, format string, a ...in
 		m = make(map[string]interface{})
 		copyAttributes(m, attributes)
 	}
+	if l := len(s.data.Annotations); l > 0 && l >= s.bufferLimit {
```
basvanbeek: not all annotations are equal. especially once we start to have common annotations which have global meaning or even have annotations which have consumer platform meaning we need to be able to whitelist these so they don't get lost by tags of lesser importance.

I think we need some sort of "must keep" list that is configurable and probably have a list of system annotations that should never be removed.

trace/trace.go
```diff
@@ -393,6 +422,10 @@ func (s *Span) AddMessageSendEvent(messageID, uncompressedByteSize, compressedBy
 	}
 	now := time.Now()
 	s.mu.Lock()
+	if l := len(s.data.MessageEvents); l > 0 && l >= s.bufferLimit {
```
basvanbeek: same as my comment on annotations

trace/trace.go
```diff
@@ -110,6 +111,9 @@ const (
 	SpanKindClient
 )
 
+// DefaultBufferLimit is the default value for trace.StartOptions.BufferLimit.
+const DefaultBufferLimit = 1000
```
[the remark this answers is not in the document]
jgracenin: Yeah, I suppose I was just trying to put a cap on the current long-lived stream memory leak, but we should think it through a bit more.

Your suggestion is pretty easy to implement, but I'll just leave it as is until others chime in (or have a different suggestion altogether).

I think as long as we have sensible defaults for these 3 options (so the client doesn't need to always specify 3 options), then I wouldn't mind having all 3.

I'd also be in favor of whatever the other language implementations are doing (or the spec) to keep the behavior consistent, but I haven't looked that up yet.

trace/trace.go
jgracenin: My expectation was: only the most recent message events would be recorded / exported, even if it was just 1. But it sounds like I might be mistaken here (I'm not very familiar with the internals).

If we end up doing different floors, then we may need to validate the options (instead of just silently fixing it within the option). I was debating that already with the shared buffer limit, but it's a breaking change and makes it less convenient to use. Something like this:

ocsh, err := ocgrpc.NewServerHandler(...trace.StartOptions)  

(and a drawback: the ServerHandler.IsPublicEndPoint option isn't a trace.StartOption)

trace/trace.go
jgracenin: Interesting, I didn't know this was in the works.

Sounds like the "must keep" annotations should be stored in a separate slice, but that would require other changes. SpanData could have an Annotations() method that merges the separate annotation slices together (and maybe a Kind enum field -- system/user/bulk should be in each one). I suppose the individual slices could still be exported alongside the merge helper method, though.

Regardless, maybe we should consider doing two commits for this (global / simple limits first, classified ones later)? The message events leak in particular is fairly serious for anyone that has a long lived stream.

rakyll:
Can we discuss changed like this on the specs first? I'd expect all language implementations have support for this, especially if we are going to introduce new APIs. https://github.com/census-instrumentation/opencensus-specs

jgracenin:
Good idea.  census-instrumentation/opencensus-specs/issues/138

trace/export.go
```diff
@@ -73,4 +73,11 @@ type SpanData struct {
 	Status
 	Links           []Link
 	HasRemoteParent bool
+
+	// The count of all span.bufferLimit overflows.
+	// See trace.WithBufferLimit for more info.
+	// TODO: This is not currently exported anywhere (e.g. tracez page).
+	DroppedAnnotations   int
```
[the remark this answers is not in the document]
ramonza: Added something to specs to this effect: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md

(Totally optional IMO, we can do it in a follow up if you like)

trace/trace.go
jgracenin: Hey, I have an updated CL ready. I'd like to clarify one thing before sending it out.

The spec refers to a global trace config (allowing global overrides). It looks like this is what Java/C++ does today as well.

Are the per-span overrides in addition to the global default overrides or do they replace the global default overrides altogether?

I assumed it was in addition and ended up adding a GlobalOption variadic option type for trace.ApplyConfig. I don't mind removing this if you don't want global overrides. The signature looks like: type GlobalOption func(*Config).

Here's how it works:
1) global var config is set with defaults per the spec
2) trace.ApplyConfig rolls variadic GlobalOptions on top of that (if any)
3) trace.StartSpan rolls variadic StartOptions on top of that (if any)
4) unit test added to cover each scenario

This doesn't break trace.ApplyConfig's API, but I couldn't use the required config struct arg (since it's ambiguous if the integers were set or not (due to the zero value for that type). If we move forward with this change, we may want to deprecate the required config struct arg at a later time (I know it's used in a lot of places though).

trace/export.go
```diff
@@ -73,4 +73,11 @@ type SpanData struct {
 	Status
 	Links           []Link
 	HasRemoteParent bool
+
+	// The count of all span.bufferLimit overflows.
```
rakyll: golint your code.

// DroppedAnnotations represents all the annotations dropped
// due to the annotations limit.
DroppedAnnotations   int

trace/export.go
```diff
@@ -73,4 +73,11 @@ type SpanData struct {
...
+	// See trace.WithBufferLimit for more info.
+	// TODO: This is not currently exported anywhere (e.g. tracez page).
+	DroppedAnnotations   int
+	DroppedMessageEvents int
```
rakyll: Add godoc.

trace/export.go
```diff
@@ -73,4 +73,11 @@ type SpanData struct {
...
+	DroppedLinks         int
```
rakyll: Add godoc.

trace/trace.go
rakyll: I think we need to be consistent and allow user to override this default via trace.ApplyConfig. It is not represented in the specs and I filed https://github.com/census-instrumentation/opencensus-specs/issues/147 to discuss.

trace/trace.go
```diff
@@ -145,13 +154,25 @@ func WithSampler(sampler Sampler) StartOption {
...
 }
 
+// WithBufferLimit makes new spans with a custom buffer limit.
+// This limits variable trace span data: Message Events, Links, and Annotations.
+// The default is 1000 (trace.DefaultBufferLimit) and the minimum is 1.
+func WithBufferLimit(limit int) StartOption {
+	return func(o *StartOptions) {
+		o.BufferLimit = limit
+		if o.BufferLimit <= 0 {
+			o.BufferLimit = 1
```
rakyll: it should be the default not 1.

trace/trace.go
```diff
@@ -145,13 +154,25 @@ func WithSampler(sampler Sampler) StartOption {
...
+}
+
 // StartSpan starts a new child span of the current span in the context. If
 // there is no span in the context, creates a new trace and span.
 //
 // Returned context contains the newly created span. You can use it to
 // propagate the returned span in process.
 func StartSpan(ctx context.Context, name string, o ...StartOption) (context.Context, *Span) {
-	var opts StartOptions
+	opts := StartOptions{BufferLimit: DefaultBufferLimit}
```
rakyll: Don't set the default here.

In startSpanInternal, set the default if opts.BufferLimit is zero.

trace/trace.go
```diff
@@ -174,7 +195,7 @@ func StartSpan(ctx context.Context, name string, o ...StartOption) (context.Cont
 // Returned context contains the newly created span. You can use it to
 // propagate the returned span in process.
 func StartSpanWithRemoteParent(ctx context.Context, name string, parent SpanContext, o ...StartOption) (context.Context, *Span) {
-	var opts StartOptions
+	opts := StartOptions{BufferLimit: DefaultBufferLimit}
```
rakyll: Ditto.

trace/trace.go
rakyll: We are not sure if user should change the defaults right now. We shouldn't export this constant.

rakyll: The spec have different limits for each entities. Attributes: 32, Message events: 128, ...

https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md

jgracenin:
Hi @rakyll,

Sorry for the confusion, I didn't push anything new since the new spec was published / @ramonza's initial review on Monday. Most of these changes are already implemented, but I'll double check anything you pointed out as well.

I added a question in a comment above (which caused the notification), I'll just paste it again here. 

Though, I think you just answered no to this with your "We are not sure if user should change the defaults right now" comment.

---

Hey, I have an updated CL ready. I'd like to clarify one thing before sending it out.

The spec refers to a global trace config (allowing global overrides). It looks like this is what Java/C++ does today as well.

Are the per-span overrides in addition to the global default overrides or do they replace the global default overrides altogether?

I assumed it was in addition and ended up adding a GlobalOption variadic option type for trace.ApplyConfig. I don't mind removing this if you don't want global overrides. The signature looks like: type GlobalOption func(*Config).

Here's how it works:

- global var config is set with defaults per the spec
- trace.ApplyConfig rolls variadic GlobalOptions on top of that (if any)
- trace.StartSpan rolls variadic StartOptions on top of that (if any)
- unit test added to cover each scenario

This doesn't break trace.ApplyConfig's API, but I couldn't use the required config struct arg (since it's ambiguous if the integers were set or not (due to the zero value for that type). If we move forward with this change, we may want to deprecate the required config struct arg at a later time (I know it's used in a lot of places though).

trace/trace.go
ramonza: I don't think we need global overrides. Having three levels of overrides sounds like going too far. I think its enough to have global defaults that can be changed by trace.ApplyConfig and per-Span values.

bogdandrutu: We can always add the "panic level" later if needed.

trace/export.go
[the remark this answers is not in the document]
jgracenin: Sure, I'll do this in a follow-up right after this (updated the TODO).

trace/trace.go
jgracenin: We've decided on a different strategy (per @ramonza). The client can pass anything in. If it's <= 0, then that trace type is disabled. I think it simplifies things.

trace/trace.go
jgracenin: See previous comment on line 175.

trace/trace.go
jgracenin: Done. Annotation limit removed for now (TODO put in place).

trace/trace.go
```diff
@@ -356,6 +381,10 @@ func (s *Span) printStringInternal(attributes []Attribute, str string) {
 		a = make(map[string]interface{})
 		copyAttributes(a, attributes)
 	}
+	if l := len(s.data.Annotations); l > 0 && l >= s.bufferLimit {
+		s.data.Annotations = s.data.Annotations[1:]
```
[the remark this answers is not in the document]
jgracenin: Ah, good catch. The underlying array in the slice is still subject to re-allocation with this. Ring buffer sounds like a good idea. I'm not sure if we should have an interface{} based ring or separate ones for each type. If Go had generics, this would be easier to do.

Another trick we could do: when we hit max capacity, we keep overwriting the last event (we'd also get the first and last events this way). I think this may confuse people though.

I'd say we leave it as is and open up a performance issue to implement a ring buffer.

trace/trace.go
jgracenin: I'm not sure if you mean a separate limit for message events (now implemented) or separate limits within message events like annotation priorities (is that also happening?)

trace/export.go
jgracenin: Done. I'm a Xoogler that misses the Critique golint badges.

trace/export.go
jgracenin: Done.

trace/export.go
jgracenin: Done.

trace/trace.go
jgracenin: Interesting, I've been referring to a "global override" as something that overrides/changes global defaults with trace.ApplyConfig. Naming things is hard :)

So, we're in agreement then and I've posted my updated commit.

However, trace.ApplyConfig's existing implementation (coupled with limit values have meaning when they're <= 0) still makes this trickier than it seems due to Go's handling of zero values. I'll start a comment thread over there.

trace/config.go
```diff
@@ -16,25 +16,72 @@ package trace
...
+	return func(c *Config) {
+		c.maxLinks = limit
+	}
 }
 
 // ApplyConfig applies changes to the global tracing configuration.
 //
 // Fields not provided in the given config are going to be preserved.
-func ApplyConfig(cfg Config) {
+func ApplyConfig(cfg Config, o ...GlobalOption) {
```
jgracenin: This is where it gets tricky:

With the current trace.ApplyConfig implementation:
1) cfg is a required parameter and is expected to be a partial update (preserving what was already set).
2) We want limits set to <= 0 to mean "disabled".
3) Go's zero value for an int is 0, so if someone changes one limit (or even just the DefaultSampler), the others would incorrectly become disabled.

To handle this, I added in a new variadic GlobalOption (which doesn't break the API), but I think the cfg arg should be deprecated in favor of all variadic options. I'm sure a lot of clients currently use this, though.

Another option: we could keep the cfg struct arg, add *int fields in for the limits, and add a trace.Limit function that takes an int and returns an *int, but I don't like that API as much.

bogdandrutu:
@jgracenin look what we did in java:
https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/config/TraceParams.java

You start from the current config -> get a builder -> apply new mutations (everything else stays the same) -> create the new config -> apply the new config.

ramonza:
@bogdandrutu agree that would be a nicer approach but currently users are not expected to call any function to construct a trace.Config so there is no opportunity to set defaults.

trace/config.go
```diff
@@ -16,25 +16,72 @@ package trace
...
 type Config struct {
 	// DefaultSampler is the default sampler used when creating new spans.
 	DefaultSampler Sampler
 
 	// IDGenerator is for internal use only.
 	IDGenerator internal.IDGenerator
+
+	// The below config options must be set with a GlobalOption.
+	// maxAttributes sets a global limit on the number of attributes.
+	maxAttributes int
```
ramonza: I'm not sure these really belong in `trace.Config`. The simpler thing to do would be to just have them as package-vars. The reason to have them in `trace.Config` would be to be able to change them at arbitrary times at runtime in a goroutine-safe manner. But I can't really think of a compelling reason you'd want to do that. @rakyll interested to know what do you think of this?

ramonza: Discussed with @rakyll offline. We think the best approach is to remove the `GlobalOption` type and instead rely on `0` to mean "don't change" (just like it does for the other fields of `Config`. If the user wants to not buffer anything (alway drop) they can always set the field to -1 explicitly.

trace/trace.go
```diff
@@ -427,10 +491,14 @@ func (s *Span) AddMessageReceiveEvent(messageID, uncompressedByteSize, compresse
 
 // AddLink adds a link to the span.
 func (s *Span) AddLink(l Link) {
-	if !s.IsRecordingEvents() {
+	if !s.IsRecordingEvents() || s.maxLinks <= 0 {
 		return
 	}
 	s.mu.Lock()
+	if l := len(s.data.Links); l > 0 && l >= s.maxLinks {
+		s.data.Links = s.data.Links[1:]
```
ramonza: Ditto

Issue #762, opened 2018-05-24
examples/exporter: PrintExporter output is not clear 


rogpeppe:
When I run the helloworld_server example, it prints lines looking like this:

    2018/05/24 14:55:49 &{0xa14a80 2018-05-24 14:55:28.201596042 +0100 BST m=+1.004805399 2018-05-24 14:55:49.202266421 +0100 BST m=+22.005475512 []}

This is because PrintExporter just uses log.Println(SpanData), which doesn't look great.
It might be nice if the example used an output format that made it clearer what information is available to the exporter.

ramonza:
Important to remember that this is just for illustrative examples. Maybe we should have a true log exporter that's designed for production use with features like templating of the log message?

vaijab:
> Maybe we should have a true log exporter that's designed for production

This could be an interesting feature to spec out. I am actually looking into some logger that logs at request (gRPC) level. It would be pretty cool to have a logger that could actually log selected fields from a trace, i.e. traceId, spanId, attributes, etc.

I would be happy to contribute to a spec and/or code. :-)

ramonza:
Repurposing this for a production-quality structured logging exporter.

savaki:
I'm interested in picking this one up.  We ended already doing something like this internally.  Here's an example of what it looks like (census is our wrapper for opencensus)

```go
census.Info(ctx, "message",
   census.String("key", "value"),
   census.Int("id", "123"),
);
```

ramonza:
Retitled to make it clear that this issue is around supporting the existing {view,trace}.Exporter interfaces rather than creating a new signal type.

ramonza:
Yup, happy to restore the original purpose. If we want to do a zap exporter, it would probably get it's own repo.

rghetia:
fixed by #911

PR #827, opened 2018-07-15
proposed commit for log package

savaki:
I saw the request posted for a log package posted here, https://github.com/census-instrumentation/opencensus-go/issues/762, and wanted to propose this package.  

* Exporters using same pattern as stats and trace
* capturing TraceID and SpanID from current span (if present)
* capturing specified tags from context
* customizable time function

Per Dave Cheney's post, https://dave.cheney.net/2015/11/05/lets-talk-about-logging, the package only supports Debug and Info level logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 OpenCensus Go is a Go implementation of OpenCensus, a toolkit for
 collecting application performance and behavior monitoring data.
-Currently it consists of three major components: tags, stats, and tracing.
+Currently it consists of three major components: tags, stats, tracing, and logging.
 
 ## Installation
 
@@ -190,6 +190,21 @@ methods for different protocols.
 * HTTP integrations uses Zipkin's [B3](https://github.com/openzipkin/b3-propagation)
   by default but can be configured to use a custom propagation method by setting another
   [propagation.HTTPFormat](https://godoc.org/go.opencensus.io/trace/propagation#HTTPFormat).
+  
+## Logs
+
+OpenCensus provides a structured logger that is both tag aware and trace aware.  `log` can be configured to 
+look for tags within the current `context.Context` and capture them as part of the log record.  In addition,
+if the log is associated with a `trace.Span`, the `TraceID` and `SpanID` will be captured as well.
+
+Structured fields can be associated with the message. For example:
+
+```go
+log.Info(ctx, "message",
+	log.String("userID", "abc"),
+	log.Int("count", 123),
+)
+```
 
 ## Execution Tracer
 
--- a/log/README.md
+++ b/log/README.md
@@ -1,0 +1,36 @@
+log
+--------
+
+`log` package supports structured logging that is both tag and trace aware.
+
+
+#### QuickStart
+
+The simplest way to get started is to register a logger and get started.
+
+```go
+var exporter log.Exporter = ...
+log.RegisterExporter(exporter)
+log.Info(ctx, "hello world", log.String("userID", "abc"), log.Int("count", 123))
+```
+
+#### Customizing
+
+The behavior of the `log` package can be customized via `ApplyConfig`.
+
+```go
+var exporter log.Exporter = ...
+userID, _ := tag.NewKey("userID")
+log.RegisterExporter(exporter)
+
+// customize behavior of log package here
+log.ApplyConfig(log.Config{
+	LogLevel: log.DebugLevel,                          // change log level to debug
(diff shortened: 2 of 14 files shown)


vaijab:
I wonder, wouldn't be better to use existing logging libraries as plugins, such as logrus, zap, etc?

savaki:
@vaijab That's not incompatible with what's being put together here.  One could certainly write a logrus, zap, etc log exporter.  In fact, if this PR is accepted, a zap exporter will probably be the first one that I write.  

The intent of this log package is to provide a trace AND tag aware log framework with many exporters logrus, zap, cloudwatch, stackdriver, etc.

Internally, we already have something like this and have found it incredibly useful to share tags across traces, logs, and metrics.  Also, being able to correlate logs to traces is very valuable as well.

ramonza:
@savaki this is really awesome work. I really wish we could accept this right now but one of the core principles of OpenCensus is to have features like this designed in abstract form in the opencensus-specs repo first and agreed to by all the language implementations before they are implemented. This process ensures that users who have software in multiple languages still have a consistent experience.

The original issue you mention (#762) is for implementing a logging exporter for the existing trace and stats APIs. Exporters do not generally require spec changes since they are just implementing the existing interfaces. I really am sorry if the original issue wasn't clear enough about this. I will comment there to make it clearer.

So what I propose we do now is that you put this code in a separate GitHub repo. Alternatively, if you don't want to create a separate repo you are welcome to add it to census-ecosystem/opencensus-experiments under the go/ subdirectory.  I would still be very happy to review and give feedback. Hopefully having it in a separate repo doesn't require any internal access to the OpenCensus core code but even it it does we can negotiate providing an API for you.

Once you have it in a separate repo, we can link to it from the main README for discoverability. After that, if you'd like to pursue adding this as an official logging API for OpenCensus for all languages, we will have to go through the specs process. I would be happy to help with that, although you should be aware that up 'till now, our logging strategy that we've discussed and partially implemented in Java has revolved around using existing logging packages rather than adding a new logging API.

ramonza:
> Internally, we already have something like this and have found it incredibly useful to share tags across traces, logs, and metrics. Also, being able to correlate logs to traces is very valuable as well.

Out of interest, what logging API do you use for this?

savaki:
Thanks for the feedback.  The intent was to provide a trace and tag aware logging interface that could be implemented by any existing logging library.  Despite the tests and coverage, there is no default implementation for log.Exporter provided by the library.  It's draws heavily from Uber's zap logging framework as well as the structured logging done in opentracing.

One of the main reasons I thought you were looking for this is that none of the existing logging frameworks (that I'm aware of) accept `context.Context` as a parameter.   Consequently, integration with trace and tags would be problematic at best, downright annoying at worst.

> Maybe we should have a true log exporter that's designed for production use with features like templating of the log message?

I saw this comment as part of https://github.com/census-instrumentation/opencensus-go/issues/762 and took this as a request for a package.

> Out of interest, what logging API do you use for this?

We wrote something ourselves that wraps `go.uber.org/zap` in much the same way that this proposal would wrap a log.Exporter.

ramonza:
Just to be clear, I would personally really really like to see some kind of general structured logging support go in eventually. But I do think we need agreement from all stakeholders first. I think prototyping this externally in a different repo is the right approach to start out with. 

> One of the main reasons I thought you were looking for this is that none of the existing logging frameworks (that I'm aware of) accept context.Context as a parameter. 

I know, this struck me as an omission as well initially. But it's quite easy to build your own wrapper (as you have done). We need to think very carefully before introducing yet another logging API to the ecosystem.

### CONTROVERSIAL OPINION BELOW :)

I have been thinking about this for a while...

There is a very important pattern that I have seen over and over in my career that is not well captured by existing libraries and framework: the "one very wide event/log line per top-level request with lots of fields and counters" pattern (better name suggestions welcome). I think it's best captured in [this blog post](https://www.honeycomb.io/blog/2018/06/how-are-structured-logs-different-from-events/). 

If you have an incredible tracing system integrated everywhere (like we are lucky enough to have with Dapper at Google) then you don't need this kind of thing as much. But in many environments this is an incredibly useful source of debugging information.

Furthermore, that seems to be what we *really* want when we say we want to include context values/tags in the logs. But they are not enough. Specifically: tags (and context values) are immutable. This makes them badly suited to this pattern because what you want is to *accumulate* key/value pairs and counters as you process the request describing how the request was processed - really a lot like a Span.

ramonza:
Discussion needs to move to specs repo, closing this PR for now.

PR #829, opened 2018-07-16
DO NOT MERGE - Intended for discussion of local aggregations

savaki:
Thought I would stir up discussion on support for local aggregation agents with a proof of concept, https://github.com/census-instrumentation/opencensus-specs/issues/63

The challenge, as I see it, is handling where the aggregation is to be done.  In the current opencensus-go implementation, the opencensus client handles aggregation and sends the server pre-aggregated results.  Clearly, there are many benefits to this.

The difficulty arises when one wants to be compatible with myriad of existing hosted stats services such as AWS CloudWatch, statsd, DataDog, etc.  Each of these solutions expects more raw data so they can do the aggregation themselves (either on an agent or the server).

So how to resolve this?  

* One solution would be what the DataDog team took.  Rather than support Sum and Counter types which would require de-aggregating the data, they opted to treat everything as a gauge, https://github.com/DataDog/opencensus-go-exporter-datadog/blob/c11c70d4821e451dfb12cca1e38f338d7cb26374/stats.go#L66  While a simpler solution, I think losing support for Sum/Counter is a hefty price to pay.
* Another alternative is to have each client de-aggregate the opencensus view.Aggregation  While this is possible, it seems like the additional round trip would be something to be avoided.
* Lastly is a solution proposed by this code.  Add an exporter to the stats package directly and allow exporters access to the raw Measurements.

The conflict I can see arising with this last approach is what ends up happening to the view package.  If I use opencensus with say DataDog or AWS CloudWatch, I wouldn't need to declare a view, that's what AWS and DataDog do.  I can simply declare a Measure and be done with it.

This seems less than ideal as over time, if many people use opencensus with cloud providers that do their own aggregation, the use of view may become inconsistent.

This leads me to believe that Measures and Views have a closer relationship than the current opencensus implementation would imply.

Would love to hear people's thoughts on this.

### DIFF
--- a/stats/export.go
+++ b/stats/export.go
@@ -1,0 +1,64 @@
+package stats
+
+import (
+	"sync/atomic"
+
+	"go.opencensus.io/tag"
+)
+
+var (
+	registry     = map[Exporter]struct{}{}
+	registrySize int32
+)
+
+// Data encapsulates the contents of the stats to be exported
+type Data struct {
+	Tags         *tag.Map
+	Measurements []Measurement
+}
+
+// Exporter exports the collected records as view data.
+//
+// The ExportView method should return quickly; if an
+// Exporter takes a significant amount of time to
+// process a Data, that work should be done on another goroutine.
+//
+// The Data should not be modified.
+type Exporter interface {
+	ExportStats(data Data)
+}
+
+// RegisterExporter registers an exporter.
+// Collected data will be reported via all the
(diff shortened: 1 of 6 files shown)


ramonza:
@savaki can we move this discussion to opencensus-specs repo? I am closing this PR for now, we can re-open once we have agreement on the specs repo on the way to proceed.

PR #830, opened 2018-07-16
made trackingResponseWriter a properly transparent wrapper

### LINKED ISSUE #744: plugin/ochttp: implement io.ReadFrom

basvanbeek:
The trackingResponseWriter used by ochttp implemented most of the additional interfaces that can be found in http ResponseWriter implementations by wrapping the involved methods. It lacked the io.ReaderFrom additional interface.

The downside of wrapping all potential methods per these interfaces is that code testing for the availability of these interfaces can signal false positives when the underlying implementation does not support them allowing for changed behavior.

This PR solves the problem of false positives by only embedding the interfaces found in the underlying implementation, making trackingResponseWriter completely transparent to its consumers. It also includes handling of io.ReaderFrom which fixes #744.

### DIFF
--- a/plugin/ochttp/server.go
+++ b/plugin/ochttp/server.go
@@ -15,10 +15,8 @@
 package ochttp
 
 import (
-	"bufio"
 	"context"
-	"errors"
-	"net"
+	"io"
 	"net/http"
 	"strconv"
 	"sync"
@@ -139,7 +137,7 @@ func (h *Handler) startStats(w http.ResponseWriter, r *http.Request) (http.Respo
 		track.reqSize = r.ContentLength
 	}
 	stats.Record(ctx, ServerRequestCount.M(1))
-	return track, track.end
+	return track.wrappedResponseWriter(), track.end
 }
 
 type trackingResponseWriter struct {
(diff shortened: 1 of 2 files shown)


plugin/ochttp/server.go
```diff
@@ -223,8 +191,231 @@ func (t *trackingResponseWriter) WriteHeader(statusCode int) {
...
 }
 
-func (t *trackingResponseWriter) Flush() {
-	if flusher, ok := t.writer.(http.Flusher); ok {
-		flusher.Flush()
+// wrappedResponseWriter returns a wrapped version of the original
+//  ResponseWriter and only implements the same combination of additional
+// interfaces as the original.
+// This implementation is based on https://github.com/felixge/httpsnoop.
+func (t *trackingResponseWriter) wrappedResponseWriter() http.ResponseWriter {
```
ramonza: Is this code generated? It would be great to get the script used committed as well. I can imagine this being useful for databases/sql stuff as well.

plugin/ochttp/server.go
```diff
@@ -153,39 +151,9 @@ type trackingResponseWriter struct {
 	writer     http.ResponseWriter
 }
 
-// Compile time assertions for widely used net/http interfaces
-var _ http.CloseNotifier = (*trackingResponseWriter)(nil)
-var _ http.Flusher = (*trackingResponseWriter)(nil)
-var _ http.Hijacker = (*trackingResponseWriter)(nil)
-var _ http.Pusher = (*trackingResponseWriter)(nil)
+// Compile time assertion for ResponseWriter interface
```
ramonza: I get that it's not possible to test this with a static assertion any longer but would it be possible to test with reflection?

plugin/ochttp/server.go
basvanbeek: It wasn't generated, but it could be. I might build a generic generator for optional interfaces as I have some ideas about it. database/sql doesn't have many optional interfaces that need to be wrapped like this. Most of the optional interfaces allow the method implementations to return `driver.ErrSkip` to signal the fast path / implementation is unavailable so I only had to use this logic for `driver.StmtExecContext` and `driver.StmtQueryContext`

plugin/ochttp/server.go
basvanbeek: Feels like a who tests the tester kind of scenario. I think it makes more sense to test for actual scenarios (which we already have a few in) then some synthetic catch all test.

plugin/ochttp/server.go
ramonza: ack

ramonza:
LGTM

PR #832, opened 2018-07-17
optimizes Span.End() to return faster if possible

basvanbeek:
Currently Span.End() will create SpanData and place a lock around exporter iteration regardless of the spandata being consumed or actual exporters being registered.

This PR solves that issue by keeping an atomic exporter counter and testing if span data will be consumed by either SpanStore or any Exporter. If no exporters are registered the lock will not be taken and if SpanData is not needed it will not be created.

This solves #681.

### DIFF
--- a/trace/export.go
+++ b/trace/export.go
@@ -15,7 +15,6 @@
 package trace
 
 import (
-	"sync"
 	"sync/atomic"
 	"time"
 )
@@ -31,10 +30,10 @@ type Exporter interface {
 	ExportSpan(s *SpanData)
 }
 
+type exportersMap map[Exporter]struct{}
+
 var (
-	exportersMu  sync.Mutex
-	exportersCnt int32
-	exporters    map[Exporter]struct{}
+	exporters atomic.Value
 )
 
 // RegisterExporter adds to the list of Exporters that will receive sampled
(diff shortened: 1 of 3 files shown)


trace/trace.go
```diff
@@ -248,19 +248,20 @@ func (s *Span) End() {
...
-				e.ExportSpan(sd)
+		mustExport := s.spanContext.IsSampled() && atomic.LoadInt32(&exportersCnt) > 0
+		if s.spanStore != nil || mustExport {
+			sd := s.makeSpanData()
+			sd.EndTime = internal.MonotonicEndTime(sd.StartTime)
+			if s.spanStore != nil {
+				s.spanStore.finished(s, sd)
+			}
+			if mustExport {
+				exportersMu.Lock()
```
ramonza: I think the major problem here is really the global lock itself. 
How about replacing it with an atomic.Value that holds the list of exporters and use copy-on-write to update the exporter list?

basvanbeek:
@ramonza, I removed the global lock as requested.

trace/export.go
```diff
@@ -30,30 +30,38 @@ type Exporter interface {
...
+	new := make(exportersMap)
+	if old, ok := exporters.Load().(exportersMap); ok {
+		for k, v := range old {
+			new[k] = v
+		}
 	}
-	exporters[e] = struct{}{}
-	exportersMu.Unlock()
+	new[e] = struct{}{}
+	exporters.Store(new)
```
ramonza: I think we still need a mutex in RegisterExporter (but not when reading) to synchronize writers. See the "ReadMostly" example in https://golang.org/pkg/sync/atomic/#Value.

trace/export.go
```diff
@@ -30,30 +30,38 @@ type Exporter interface {
...
 }
 
 // UnregisterExporter removes from the list of Exporters the Exporter that was
 // registered with the given name.
 func UnregisterExporter(e Exporter) {
-	exportersMu.Lock()
-	delete(exporters, e)
-	exportersMu.Unlock()
+	new := make(exportersMap)
```
ramonza: Ditto

trace/export.go
basvanbeek: I purposely didn't do that because I didn't see a major use case for concurrent registration / deregistration of exporters. 

We have 2 choices:
1. add locks
2. document register and deregister not being concurrency safe.

trace/export.go
basvanbeek: as above... I think 99% of use cases for register / deregister is from start up or config change... most of which will always be serial.

basvanbeek:
I guess since these won't be called often it is no issue to just add the lock for safety... will add...

trace/export.go
ramonza: Yeah I would normally agree but since concurrent registration/deregistration won't be picked up by the race detector I would say its worth it in this case.

PR #840, opened 2018-07-20
Fix race to update config in trace.ApplyConfig

ramonza:


trace/config.go
```diff
@@ -25,10 +28,14 @@ type Config struct {
...
 }
 
+var configWriteMu sync.Mutex
+
 // ApplyConfig applies changes to the global tracing configuration.
 //
 // Fields not provided in the given config are going to be preserved.
 func ApplyConfig(cfg Config) {
+	configWriteMu.Lock()
+	defer configWriteMu.Unlock()
```
basvanbeek: LGTM.

Maybe add a comment on `config.Load()` always returning a valid Config object, as config is populated by an `init()`

ramonza: @basvanbeek you need to actually Approve the review for it to be mergeable - LGTM is not enough.

basvanbeek: We didn't discuss me officially reviewing stuff... but I guess I can ;)

odeke-em:
@ramonza @basvanbeek, sorry that I am late to the party, I've been swamped here but how did we discover this race? Perhaps could we write a test to ensure that we never regress? I kindly ask because this is the second PR that I've seen fixing a race in ApplyConfig.

PR #841, opened 2018-07-22
Add Jaeger Trace header propagation.

tcolgate:


tcolgate:
I still need to test this with an actual jaeger instance, but pushing for general yes/no

tcolgate:
@basvanbeek similar discussions in https://github.com/census-instrumentation/opencensus-node/pull/74 , and  https://github.com/census-instrumentation/opencensus-python/pull/222 . The specs state that other propagation mechanisms can be supported. Obviously I can keep this out of tree if this is the decision.

ramonza:
@basvanbeek I disagree that this should be in a contrib repo - the Jaeger exporter is in this repo so I think it only makes sense for the propagation format to be here too. It would be strange to have the exporter here but the propagation format elsewhere.

basvanbeek:
@ramonza, I disagree strongly... you're conflating Jaeger the backend and Jaeger the instrumentation libraries.

See: https://github.com/census-instrumentation/opencensus-specs/issues/150

ramonza:
@bogdandrutu who can make a call here on a policy that we can follow across all libraries?

ramonza:
ping @bogdandrutu

tcolgate:
I'm not sure if this is still relevant. I'm going to close to get this off of my open PRs list.

PR #842, opened 2018-07-23
Fix Jaeger exporter duration computation

zhuyie:


basvanbeek:
The existing code `duration = endTime.Sub(startTime)` is correct as `Sub` uses a monotonic clock. The solution you provide does not and will provide incorrect duration timings.

Too see an example of how it went wrong before Go included monotonic clock logic to the time package see: https://blog.cloudflare.com/how-and-why-the-leap-second-affected-cloudflare-dns/

zhuyie:
Original data(in nanoseconds):
ParentSpan: StartTime=1532163129411176800, EndTime=1532163129455178700, (Duration=44001900)
ChildSpan : StartTime=1532163129426177700, EndTime=1532163129455178700, (Duration=29001000)

Old algorithm(in microseconds):
ParentSpan : StartTime=1532163129411176, Duration=44001, (EndTime=1532163129455177)
ChildSpan  : StartTime=1532163129426177, Duration=29001, (EndTime=1532163129455178)

New algorithm(in microseconds):
ParentSpan : StartTime=1532163129411176, EndTime=1532163129455178, (Duration=44002)
ChildSpan  : StartTime=1532163129426177, EndTime=1532163129455178, (Duration=29001)

zhuyie:
In old algorithm, ChildSpan.EndTime > ParentSpan.EndTime, new algorithm fix this.

basvanbeek:
That's because of a rounding error going from nanos to micros. Unfortunately your solution has critical flaws as described above and therefore can't be accepted.

basvanbeek:
Can you try this:
```
StartTime:     data.StartTime.Round(time.Microsecond).UnixNano() / 1000,
Duration:      data.EndTime.Sub(data.StartTime).Round(time.Microsecond).Nanoseconds() / 1000,
```

exporter/jaeger/jaeger.go
```diff
@@ -163,15 +163,20 @@ func spanDataToThrift(data *trace.SpanData) *gen.Span {
 			SpanId:      bytesToInt64(link.SpanID[:]),
 		})
 	}
+
+	startTimeInMS := data.StartTime.UnixNano() / 1000
```
bogdandrutu: This should be us not ms based on the calculation here.

exporter/jaeger/jaeger.go
```diff
@@ -226,6 +223,14 @@ func attributeToTag(key string, a interface{}) *gen.Tag {
 	return tag
 }
 
+// convertTime returns startTime/duration in microseconds.
+func convertTime(st, et time.Time) (startTime, duration int64) {
```
bogdandrutu: calculateDurantionUs?

ramonza: I would inline this

exporter/jaeger/jaeger.go
```diff
@@ -221,6 +223,14 @@ func attributeToTag(key string, a interface{}) *gen.Tag {
...
+	startTime = st.UnixNano() / 1000
+	endTime := et.UnixNano() / 1000
+	duration = endTime - startTime
```
ramonza: Why not compute the duration explicitly: 
```
duration = endTime.Sub(startTime) / time.Microsecond
```

This will use the monotonic clock. I understand that the EndTime should already be adjusted to be in the reference frame of the StartTime so this should always be the same but it is easier if the reader of this code doesn't need to know that and can locally determine that its correct.

zhuyie: That is just exact the same as the existing code:
```
Duration:      data.EndTime.Sub(data.StartTime).Nanoseconds() / 1000,
```
because
```
// Nanoseconds returns the duration as an integer nanosecond count.
func (d Duration) Nanoseconds() int64 { return int64(d) }
```
and
```
const (
        Nanosecond  Duration = 1
        Microsecond          = 1000 * Nanosecond
)
```

zhuyie: As I explained before, the existing code will output wrong data (ChildSpan.End>ParentSpan.End) to Jaeger under some circumstances, and Jaeger will adjusted these data due to "**child not fits within the parent span**", the result UI looks very bad.

A good trace:


and a **BAD** one:

zhuyie: Look at the existing code:
```
StartTime:     data.StartTime.UnixNano() / 1000,
Duration:      data.EndTime.Sub(data.StartTime).Nanoseconds() / 1000,
```

the problem here is:
1. StartTime will discards hundreds of nanoseconds.
2. data.EndTime.Sub(data.StartTime) will use those hundreds of nanoseconds in calculation.

basvanbeek: Yes this is a tricky one and only happens if the end time between child and parent span is in the sub microsecond range so rounding differences can make this occur.

Your fix solves the issue of a child span ending after a parent span due to these rounding differences but it comes at the expense of the durations being less accurate as they could be.

I personally favor duration accuracy over span end time ordering especially since a backend could remedy this while a duration inaccuracy can't be resolved. But then again how much info does a rounding error of 1 microsecond really mean.

That said, I'm ok with the proposed change.

ramonza: Oh I think I misunderstood the problem. Looking at what the Jaeger Go client does, it seems to do the same thing as our previous code:

In BuildJaegerThrift:
```go
duration := span.duration.Nanoseconds() / int64(time.Microsecond)
```

In Span.FinishWithOptions:
```go
s.duration = options.FinishTime.Sub(s.startTime)
```

So I actually think we should leave it as it was before since that's the behavior of the official Jaeger client.

zhuyie:
in my opinion, two spans end at the same time but have different end after export, it's a bug obviously. but if your guys think keep the same behavior with the official Jaeger client is more important, fine.

basvanbeek:
@zhuyie maybe it is a good idea to ask about this situation at the Jaeger project. As I mentioned it is something that could potentially be resolved at the backend when the trace data is aggregated.

If not, than at least census reports in the same way as the official Jaeger libraries as @ramonza has shown and duration accuracy is simply favored over potential rounding inaccuracies in span end times.

ramonza:
Closing this for now since the behavior matches the official Jaeger client.

PR #844, opened 2018-07-24
Don't trace /healthz

rakyll:
Kubernetes users end up having tons of /healthz traces
which is costly and no-value.

Disable /healthz tracing until we have a better solution.

### DIFF
--- a/plugin/ochttp/client.go
+++ b/plugin/ochttp/client.go
@@ -57,6 +57,9 @@ type Transport struct {
 // RoundTrip implements http.RoundTripper, delegating to Base and recording stats and traces for the request.
 func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	rt := t.base()
+	if isHealthEndpoint(req.URL.Path) {
+		return rt.RoundTrip(req)
+	}
 	// TODO: remove excessive nesting of http.RoundTrippers here.
 	format := t.Propagation
 	if format == nil {
--- a/plugin/ochttp/server_test.go
+++ b/plugin/ochttp/server_test.go
@@ -555,3 +555,44 @@ func TestHandlerImplementsHTTPCloseNotify(t *testing.T) {
 		t.Errorf("HTTP2Log got\n\t%q\nwant\n\t%q", g, w)
 	}
 }
+
+func TestIgnoreHealthz(t *testing.T) {
+	var spans int
+
+	ts := httptest.NewServer(&Handler{
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			span := trace.FromContext(r.Context())
+			if span != nil {
+				spans++
+			}
+			fmt.Fprint(w, "ok")
+		}),
+		StartOptions: trace.StartOptions{
+			Sampler: trace.AlwaysSample(),
+		},
+	})
+	defer ts.Close()
+
+	client := &http.Client{}
+
+	for _, path := range []string{"/healthz", "/_ah/health"} {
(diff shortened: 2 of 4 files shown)


rakyll:
I will update the PR to also add `_ah/health`.

plugin/ochttp/client.go
```diff
@@ -57,6 +57,9 @@ type Transport struct {
 // RoundTrip implements http.RoundTripper, delegating to Base and recording stats and traces for the request.
 func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	rt := t.base()
+	if req.URL.Path == "/healthz" {
```
odeke-em: Thanks for the quick fix @rakyll!
 
Could we please add a comment for the rationale of this change?
```go
// Disable tracing /healthz since tracing it is costly and
// it bloats Kubernetes traces without adding much value
```

rakyll:
Also needs support for the server.

plugin/ochttp/client.go
rakyll: Done.

rakyll:
Needs tests for the client spans.

plugin/ochttp/server_test.go
```diff
@@ -555,3 +555,33 @@ func TestHandlerImplementsHTTPCloseNotify(t *testing.T) {
 		t.Errorf("HTTP2Log got\n\t%q\nwant\n\t%q", g, w)
 	}
 }
+
+func TestIgnoreHealthz(t *testing.T) {
```
odeke-em: Aha you beat me to this I had just spun up this suggestion for a test
```go
func TestHealthEndpointsUntraced_handler(t *testing.T) {
        logBuf := new(bytes.Buffer)
        cst := httptest.NewServer(&Handler{
                Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
                        fmt.Fprintln(logBuf, r.URL.Path)
                }),
        })
        defer cst.Close()

        endpoints := []string{"/foo", "/", "", "/health", "/healthz", "/_ah/health"}
        client := &http.Client{Transport: new(Transport)}
        for i, endpoint := range endpoints {
                _, err := client.Get(cst.URL + endpoint)
                if err != nil {
                        t.Errorf("#%d: got error: %v", i, err)
                }
        }

        wantLog := "/foo\n/\n/\n/health\n"
        if g, w := logBuf.String(), wantLog; g != w {
                t.Fatalf("Got:\n%s\nWant:\n%s", g, w)
        }
}
```

basvanbeek:
I feel this is rushed in, which might have been necessary but I'd like to revisit with a better solution for the future... 

this just hard coded a couple of paths to disable tracing. I can see where we'd want it configurable depending on the consumer's ecosystem and maybe make it possible to decide path based sampling strategy instead of only a hard drop all for certain paths. that would allow for more functionality while also providing a good solution for this situation (with these paths using the neverSample strategy)

ramonza:
agree with @basvanbeek, this should have been discussed more broadly probably on the specs repo. what was the justification for rushing it in like this?

rakyll:
Sorry, this was rushed in. I created https://github.com/census-instrumentation/opencensus-specs/issues/151 but haven't proposed anything new such as a blacklisting mechanism. Let's discuss on the specs what we want to do and fix the implementation accordingly.

mtwo:
Agree with Bas, we should have a simple blacklist like OC Ruby (I think) does. For example, it'd be great to not trace requests for Stackdriver Profiler.

rakyll:
@mtwo, can you update the spec issue I have field? Let's keep the conversation over there.

PR #846, opened 2018-07-25
Make it easier to manually report accumulated data

### LINKED ISSUE #773: Mechanism to explicitly flush stats to exporters

knyar:
Report data for a given view when it is unregistered.

This might help with #773.

### DIFF
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -194,26 +194,30 @@ func (w *worker) tryRegisterView(v *View) (*viewInternal, error) {
 	return vi, nil
 }
 
+func (w *worker) reportView(v *viewInternal, now time.Time) {
+	if !v.isSubscribed() {
+		return
+	}
+	rows := v.collectedRows()
+	_, ok := w.startTimes[v]
+	if !ok {
+		w.startTimes[v] = now
+	}
+	viewData := &Data{
+		View:  v.view,
+		Start: w.startTimes[v],
+		End:   time.Now(),
+		Rows:  rows,
+	}
+	exportersMu.Lock()
+	for e := range exporters {
+		e.ExportView(viewData)
+	}
+	exportersMu.Unlock()
+}
+
 func (w *worker) reportUsage(now time.Time) {
 	for _, v := range w.views {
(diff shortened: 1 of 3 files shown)


stats/view/worker.go
```diff
@@ -129,6 +129,16 @@ func SetReportingPeriod(d time.Duration) {
 	<-req.c // don't return until the timer is set to the new duration.
 }
 
+// ReportUsage immediately reports collected data to all registered exporters
+// outside of the regular reporting cycle.
+func ReportUsage() {
```
ramonza: This is new API surface area so probably needs to be discussed on opencensus-specs first so that other language implementations can also be aware of it. I would recommend that for now we split this PR into two: one that just adds the flush-on-unregister behavior and another with the new API.

stats/view/worker_test.go
```diff
@@ -362,9 +362,80 @@ func TestWorkerStarttime(t *testing.T) {
...
+func TestUnregisterReportsUsage(t *testing.T) {
+	restart()
+	ctx := context.Background()
+
+	m1 := stats.Int64("measure", "desc", "unit")
+	view1 := &View{Name: "count", Measure: m1, Aggregation: Count()}
+	m2 := stats.Int64("measure2", "desc", "unit")
+	view2 := &View{Name: "count2", Measure: m2, Aggregation: Count()}
+
+	SetReportingPeriod(5 * time.Second)
```
ramonza: For this test, I would recommend setting the reporting period to something high to emphasize that we don't want the usual reporting cycle to occur.

knyar:
Thanks for the quick review! I've adjusted this PR to only flush when a view is unregistered.

I will also create a discussion issue similar to https://github.com/census-instrumentation/opencensus-specs/issues/126 but focused on stats specifically.

ramonza:
Thanks @knyar!

PR #847, opened 2018-07-29
Follow modification to TraceContext HTTP propagation specs

munisystem:
Hi,

According to [the TraceContext specification](https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#header-name), it should use `traceparent` in HTTP request header when propagating context.

This PR is that follow modification to specs.
I changed to use`traceparent` instead of `Trace-Parent`.

### DIFF
--- a/plugin/ochttp/propagation/tracecontext/propagation.go
+++ b/plugin/ochttp/propagation/tracecontext/propagation.go
@@ -29,7 +29,7 @@ import (
 const (
 	supportedVersion = 0
 	maxVersion       = 254
-	header           = "Trace-Parent"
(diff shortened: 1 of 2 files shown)


munisystem:
> This update is correct. It was decided at the distributed tracing workshop in Solingen to be explicit about lowercase and without special characters to have maximum interoperability if also used in structured logging, other transport headers, etc.

I see!
By the way, can I browse the document or the note about this decision?
I was interested it but I couldn't find documents referring to change in the repository.

rakyll:
Thanks!

PR #850, opened 2018-07-31
Root span performance improvement: re-use  lower 64 bits of TraceID for SpanID

basvanbeek:
To avoid using an extra lock and needing to generate additional 64 random bits for the SpanID for Root Spans this PR will re-use the lower 64 bits of the newly generated TraceID as SpanID.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -192,8 +192,10 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
 
 	if !hasParent {
 		span.spanContext.TraceID = cfg.IDGenerator.NewTraceID()
+		copy(span.spanContext.SpanID[:], span.spanContext.TraceID[8:])
+	} else {
+		span.spanContext.SpanID = cfg.IDGenerator.NewSpanID()
 	}
-	span.spanContext.SpanID = cfg.IDGenerator.NewSpanID()
 	sampler := cfg.DefaultSampler
 
 	if !hasParent || remoteParent || o.Sampler != nil {
--- a/trace/trace_test.go
+++ b/trace/trace_test.go
@@ -80,6 +80,21 @@ func TestStartSpan(t *testing.T) {
 	}
 }
 
+func TestRootSpanIdentifiers(t *testing.T) {
+	_, span := StartSpan(context.Background(), "StartSpan")
+	if span == nil {
+		t.Error("StartSpan: expected valid Span")
+	}
+	var (
+		spanContext = span.SpanContext()
+		got         = spanContext.SpanID.String()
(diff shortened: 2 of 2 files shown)


trace/trace.go
```diff
@@ -192,8 +192,10 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
 
 	if !hasParent {
 		span.spanContext.TraceID = cfg.IDGenerator.NewTraceID()
+		copy(span.spanContext.SpanID[:], span.spanContext.TraceID[8:])
```
ramonza: This is too surprising a behavior to implementors of IDGenerator. The reason IDGenerator was added, was to allow custom SpanID formats (as used by X-Ray). They need to set the root span ID as well.

basvanbeek:
So the majority of tracing backends we're potentially somewhat compatible with (including XRay) have no problem with this logic as most regard SpanID's purely random.

TraceID's are often purely random or in the case they are not (XRay) we see custom encoding of some form at the highest bits and randomness at the lower ones. XRay uses 32 bit epoch in the high bits and 96 randomness in the lower ones. This PR does not brake XRay or any others I know that would previously have worked with our current logic.

We can completely ignore performance benefits and random generator contention here as you propose or we can look at how we can improve this PR.

Note that Zipkin (optional but default) and Jaeger (always) use lower bits of TraceID for root Span ID's so it's a well known optimization in the space.

Example, we could update the IDGenerator interface to something like this:
```go
type IDGenerator interface {
	NewTraceID() [16]byte
	NewSpanID(traceID *[16]byte) [8]byte
}
```

Where we'd need to document that NewSpanID would receive a nil param in case the span is not a root span and would receive a pointer to the array in case it is. This would allow IDGenerators to reuse a part of the traceID to create a root span id. If not desired or possible the param could be ignored.

Another option would be an interface like this:
```go
type IDGenerator interface {
	NewTraceID() [16]byte
	NewSpanID(traceID [16]byte, isRoot bool) [8]byte
}
```

This would allow to always use the TraceID to help generate a SpanID and/or do something special if it concerns a root span. All would be optional.

Currently I don't think we'd break anything by either accepting the PR as is or adjusting in some form as presented above. `opencensus-go-exporter-aws` currently does not implement a IDGenerator to swap out the existing internal one. Also the existing one is buried in the trace/internal package without any documentation.

adriancole:
very similar to an older design in brave
https://github.com/openzipkin/brave/pull/110/files#diff-81d78836b6ff0d4b4196a39d62d30ca5R13

another option is to present the parent span id which would be 0 if root.
That said I think your option of bool for root is better.

basvanbeek:
I'm closing this PR... just noticed the actual span ID creation logic (must have been the 100F in my office yesterday)

PR #851, opened 2018-08-01
lockless version of defaultIDGenerator.NewSpanID

basvanbeek:
Without resorting to common patterns of reusing lower bits of TraceID for a root SpanID as I proposed in #850 we can actually get a lock free version of the nextSpanID logic as currently implemented for all calls to NewSpanID.


trace/trace.go
```diff
@@ -479,13 +479,10 @@ type defaultIDGenerator struct {
 // NewSpanID returns a non-zero span ID from a randomly-chosen sequence.
 // mu should be held while this function is called.
```
ramonza: Is this statement still true?

basvanbeek: Thanks for noticing, removing it.

PR #790, opened 2018-06-15
Remove Stackdriver exporter

ramonza:
The Stackdriver exporter has been moved to a different repo:
https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver

New import path: contrib.go.opencensus.io/exporter/stackdriver

Fixes: #789

PR #854, opened 2018-08-02
Revert "Remove Stackdriver exporter (#790)"

ramonza:
This reverts commit df6bd8075012c9e7088c96bc588eac92ab36d7d1.

This broke GCP client libraries - we need to revert for now and fix.


songy23:
How do we plan to move forward with this? Do we need to update the GCP client libraries first?

ramonza:
This may fix it: https://code-review.googlesource.com/c/google-api-go-client/+/31330

PR #864, opened 2018-08-13
Document that ExportView is only called from one Goroutine

ramonza:
This allows Exporter implementations to avoid unnecessary internal
synchronization.

### DIFF
--- a/stats/view/export.go
+++ b/stats/view/export.go
@@ -27,6 +27,9 @@ var (
 // Exporter takes a significant amount of time to
 // process a Data, that work should be done on another goroutine.
 //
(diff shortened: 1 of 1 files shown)


stats/view/export.go
```diff
@@ -27,6 +27,9 @@ var (
 // Exporter takes a significant amount of time to
 // process a Data, that work should be done on another goroutine.
 //
+// It is safe to assume that ExportView will not be called concurrently from
+// multiple Goroutines.
```
rakyll: "multiple goroutines"

Goroutine is not capitalized.

PR #871, opened 2018-08-18
Add possibility to Jaeger to add process tags

thebod:
Jaeger allows to attach tags to it's process, which this patch
allows to specify.
Currently this is limited to bool, string, int64 and int32, as
attributeToTag is used for converting it to *jaeger.Tag.

For reference see the official Jaeger client go:
https://github.com/jaegertracing/jaeger-client-go/blob/252d853b2a4f3cfc98bb89c3a8bb3d3aa50ed4d6/tracer.go#L64

This change helps to specify tags, such as 'ip', which has a special
handling in Jaeger (for example jaeger-query skips clock skew
adjustments for processes from the same IP).

### DIFF
--- a/exporter/jaeger/example_test.go
+++ b/exporter/jaeger/example_test.go
@@ -25,8 +25,10 @@ func ExampleNewExporter_collector() {
 	// Register the Jaeger exporter to be able to retrieve
 	// the collected spans.
 	exporter, err := jaeger.NewExporter(jaeger.Options{
-		Endpoint:    "http://localhost:14268",
-		ServiceName: "trace-demo",
+		Endpoint: "http://localhost:14268",
+		Process: jaeger.Process{
+			ServiceName: "trace-demo",
+		},
 	})
 	if err != nil {
 		log.Fatal(err)
@@ -39,7 +41,31 @@ func ExampleNewExporter_agent() {
 	// the collected spans.
 	exporter, err := jaeger.NewExporter(jaeger.Options{
 		AgentEndpoint: "localhost:6831",
-		ServiceName:   "trace-demo",
+		Process: jaeger.Process{
+			ServiceName: "trace-demo",
+		},
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	trace.RegisterExporter(exporter)
+}
+
+// ExampleNewExporter_processTags shows how to set ProcessTags
+// on a Jaeger exporter. These tags will be added to the exported
+// Jaeger process.
+func ExampleNewExporter_processTags() {
+	// Register the Jaeger exporter to be able to retrieve
+	// the collected spans.
+	exporter, err := jaeger.NewExporter(jaeger.Options{
+		AgentEndpoint: "localhost:6831",
+		Process: jaeger.Process{
+			ServiceName: "trace-demo",
+			ProcessTags: []jaeger.Tag{
+				jaeger.StringTag("ip", "127.0.0.1"),
+				jaeger.BoolTag("demo", true),
+			},
+		},
 	})
 	if err != nil {
 		log.Fatal(err)
--- a/exporter/jaeger/example_test.go
+++ b/exporter/jaeger/example_test.go
@@ -61,7 +61,7 @@ func ExampleNewExporter_processTags() {
 		AgentEndpoint: "localhost:6831",
 		Process: jaeger.Process{
 			ServiceName: "trace-demo",
-			ProcessTags: []jaeger.Tag{
+			Tags: []jaeger.Tag{
 				jaeger.StringTag("ip", "127.0.0.1"),
 				jaeger.BoolTag("demo", true),
 			},
--- a/exporter/jaeger/jaeger.go
+++ b/exporter/jaeger/jaeger.go
@@ -58,10 +58,10 @@ type Options struct {
 	Password string
 
 	// ServiceName is the Jaeger service name.
-	// deprecated: specify Process instead
+	// Deprecated: Specify Process instead.
 	ServiceName string
 
-	// Process contains the information about the exporting process
+	// Process contains the information about the exporting process.
 	Process Process
 }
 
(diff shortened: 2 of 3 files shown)


exporter/jaeger/jaeger.go
```diff
@@ -59,6 +59,9 @@ type Options struct {
 
 	// ServiceName is the Jaeger service name.
 	ServiceName string
+
+	// Tags are added to Jaeger Process exports
+	Tags []Tag
```
ramonza: How about calling this `ProcessTags` to distinguish them from tags associated with attributes on Spans?

exporter/jaeger/jaeger.go
```diff
@@ -107,13 +115,41 @@ func NewExporter(o Options) (*Exporter, error) {
...
+	return Tag{key, value}
+}
+
+// Int64Tag creates a new tag of type int64, exported as jaeger.TagType_LONG
+func Int64Tag(key string, value int64) Tag {
+	return Tag{key, value}
+}
+
+// Int32Tag creates a new tag of type int32, exported as jaeger.TagType_LONG
+func Int32Tag(key string, value int32) Tag {
```
ramonza: What's the reason for adding this? Seems like it is exported as TagType_LONG as well so why not just rely on the previous function?

rakyll:
I don't think we should merge this before we finalize what we are planning to do with tags on a span.

You can implement a wrapper Exporter that applies tags and calls the Jaeger exporter for now.

exporter/jaeger/jaeger.go
thebod: Thought about that too, but then I sticked with the way it is defined in Jaeger: https://github.com/jaegertracing/jaeger-idl/blob/b59e84ebb35cd858614f836e47540d6f95e0cf88/thrift/jaeger.thrift#L68
But no big deal to change that!

exporter/jaeger/jaeger.go
thebod: True, I added both to have a similar interface to other Tags, and support all 4 cases of `attributeToTag` (which handles both int32 and int64). I can remove it if you think it is not necessary?

thebod:
@rakyll my thoughts on a wrapper: If in the future there is a generic concept of span tags, the exporter would need to make sure to not overwrite those tags, so that wouldn't be much different from what this change would do. So I am unsure if this would be a better solution?

Jaeger handles these tags as information on the source of the spans, and not on a span level, I don't know if this concept can be (or is already in planning) generalized for more exporters?

thebod:
@ramonza I'll add an example (and possible tests) when we decided if it stays like this or we go with the wrapper, ok?

ramonza:
@rakyll I don't think a wrapper is possible: the tags mentioned here are on the Process, not the Span (as attributes would be). Without this PR, we provide no mechanism to set the tags on Process at all. An alternative would be to accept the Process itself on the Options struct instead of the name and tags separately.

exporter/jaeger/jaeger.go
ramonza: I think ProcessTags is clearer since Tags might be confused with the tags package the context of OpenCensus.

exporter/jaeger/jaeger.go
ramonza: Yes, I think we should remove it if it's not required.

rakyll:
> I don't think a wrapper is possible: the tags mentioned here are on the Process

The current change applies a set of tags to every span as attributes. I didn't truly understand this sentence, can you reword it maybe?

Btw, tag is an overloaded term. I didn't realize these are Jaeger tags but assuming we are applying OpenCensus tags on the exporter. This realization changes everything.

We should have this feature, this is not different than DefaultTraceAttributes at https://godoc.org/contrib.go.opencensus.io/exporter/stackdriver#Options. My only comment would be to improve the confusion between OC and Jaeger tags.

ramonza:
> The current change applies a set of tags to every span as attributes.

@rakyll no, this is not the case. Here is how they are used (added to Process, not Span):
```
Process: &gen.Process{
			ServiceName: e.service,				ServiceName: e.service,
			Tags:        e.tags,
		},			},

```

ramonza:
@rakyll what do you think of taking a `gen.Process` directly, instead of `ServiceName` and `Tags` in the option struct?

rakyll:
Ah, got it. I think taking a gen.Process would be reasonable.

thebod:
I have added the Process to Options, with ServiceName being deprecated (and used as a fallback).
gen.Process itself is internal, and can not be used right away.

@ramonza Int32Tag is removed, the Tags is renamed to ProcessTags, and an example was added. I wasn't sure if a test is worth it, as it would only check that all tags have been copied from Tag to *gen.Tag?

Btw. I think it would be possible to solve this by using a wrapper which changes Exporter.bundler, however then we would need to safe onError on the Exporter to be able to call it.

Also: should I squash the commits or just keep pushing them one by one?

exporter/jaeger/jaeger.go
```diff
@@ -58,7 +58,11 @@ type Options struct {
 	Password string
 
 	// ServiceName is the Jaeger service name.
+	// deprecated: specify Process instead
 	ServiceName string
+
+	// Process contains the information about the exporting process
```
ramonza: Should end with a period

exporter/jaeger/jaeger.go
```diff
@@ -58,7 +58,11 @@ type Options struct {
 	Password string
 
 	// ServiceName is the Jaeger service name.
+	// deprecated: specify Process instead
```
ramonza: Should be capitalized and have a period:
```
Deprecated: Specify Process instead.
```

exporter/jaeger/example_test.go
```diff
@@ -39,7 +41,31 @@ func ExampleNewExporter_agent() {
...
+// on a Jaeger exporter. These tags will be added to the exported
+// Jaeger process.
+func ExampleNewExporter_processTags() {
+	// Register the Jaeger exporter to be able to retrieve
+	// the collected spans.
+	exporter, err := jaeger.NewExporter(jaeger.Options{
+		AgentEndpoint: "localhost:6831",
+		Process: jaeger.Process{
+			ServiceName: "trace-demo",
+			ProcessTags: []jaeger.Tag{
```
ramonza: I think now that we have `Process` it is redundant to also have the word `Process` in the name. This can just be `Tags` no IMO.

PR #874, opened 2018-08-21
Record should not block if the work queue is full

ramonza:


ramonza:
waiting for: https://github.com/census-instrumentation/opencensus-specs/pull/159

ramonza:
We decided against this approach.

PR #877, opened 2018-08-22
Add go.mod file

rakyll:
Users should be able to test if there are any problems
during 1.11's experimental cycle.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,20 @@
+module go.opencensus.io
+
+require (
+	git.apache.org/thrift.git v0.0.0-20180807212849-6e67faa92827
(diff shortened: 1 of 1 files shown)


rakyll:
Are you experiencing the same issue with 

```
 "contrib.go.opencensus.io/exporter/stackdriver"
 "contrib.go.opencensus.io/exporter/stackdriver/propagation"
 "go.opencensus.io/trace"
```

We moved the Stackdriver exporter to a different place, deleted the package and reverted it back. Don't use that package, it is about to be entirely removed. If depending on the contrib package doesn't fix the issue, please file a bug.

Issue #869, opened 2018-08-17
http.Request influenced sampling


toffaletti:
**Is your feature request related to a problem? Please describe.**
I would like to always sample requests with certain properties. For example, when the request has `?trace=1` and the `RemoteAddr` is on an internal network the trace will be sampled.

**Describe the solution you'd like**
```
type ochttp.Handler struct {
  ....
  StartOptions func(*http.Request) trace.StartOptions
}
```

**Describe alternatives you've considered**
1. Injecting a sampler in the `http.Handler` that `ochttp.Handler` wraps, using `trace.StartSpan(..., trace.Sampler(...))`. The problem with this is that the root span added by ochttp might not be sampled.

2. Wrapping `ochttp.Handler` itself and influencing the `ochttp.Handler.StartOptions.Sample` somehow. The problem with this is that there isn't really any good way to get context into a `trace.Sampler` except by closing over some state and we can't modify the closure for every request.

3. Wrapping `ochttp.Handler` and create an additional outer span where we can set a trace.Sampler via `StartSpan` TraceOptions that closes over the `*http.Request`. Adding a meaningless span just for this seems not ideal.

4. Duplicating all of the functionality of `ochttp.Handler` in another package just to add this. Essentially maintaining a fork.

5. Creating another `http.Server` on a different port with all of the same endpoint handlers except with a `AlwaysSample` sampler set. Opening another port isn't an option for us.

ramonza:
This seems like a good approach to me.

/cc @rakyll

rakyll:
Cases like this makes me wonder whether we should have provided a hook to customize the created span.

```
type ochttp.Handler struct {
  ....
  StartSpan func(*http.Request) *trace.Span
}
```

Then, FormatSpanName or the blacklisting/filtering capabilities we have been discussing can be implemented by the user with full flexibility.

ramonza:
Could you elaborate on the reasoning behind returning `trace.Span` rather than `trace.StartOptions` as @toffaletti proposed? I think either would work: `trace.Span` is more flexible (allows adding attributes for example) but `trace.StartOptions` seems simpler to implement.

rakyll:
There is a repeating pattern I see when it comes to the need of customizing request spans. We might keep adding different fields or should provide a way for the user to override the implementation entirely.

toffaletti:
[...]

PR #881, opened 2018-08-24
add HealthEndpoints map as public

sergeylanzman:
add public HealthEndpoints for set custom Health urls


ramonza:
@sergeylanzman could you explain a bit what prompted this change? what problem are you experiencing?

sergeylanzman:
@ramonza 
Sometimes application have health check on custom url not only "/healthz". Sometime health check intergrade from internal framework and more complicate(include check connection to DB, important services and etc). I think it's not relevant for tracing and need "clean" from tracing data/statistics. 
This PR add support options for add/replace other options for health check for delete from tracing data

sergeylanzman:
I think better solution create method "UnTrackebleEndpoints" and set internal map with entpoints.

ramonza:
What about having a per-path trace.Sampler as a `map[string]trace.Sampler` field on `ochttp.Handler`?

rakyll:
Can we achieve the blacklisting functionality via https://github.com/census-instrumentation/opencensus-go/issues/869#issuecomment-422496469?

We can return a NeverSample for health endpoints.

rakyll:
Once https://github.com/census-instrumentation/opencensus-go/pull/924 is merged, I will create a ochttp/ochttputil package with some filter implementations.

PR #887, opened 2018-08-30
add initial version of W3C tracestate

reyang:
This is the initial implementation of W3C tracestate in Go.

This is my first time writing program in Go, couple notes/questions:
1. I want to discuss if want to put thread safety guard on Tracestate.Fork(), or leave it as a responsibility for the API consumer.
2. Currently the Tracestate object will be copied every time when a new span got created, I plan to implement COW (copy on write) to improve perf.
3. My next pull request will implement the ochttp integration, which will propagate tracestate in HTTP headers.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -295,14 +295,6 @@ func (s *Span) SpanContext() SpanContext {
 	return s.spanContext
 }
 
-// Tracestate returns the Tracestate of the span context.
-func (s *Span) Tracestate() *Tracestate {
-	if s == nil {
-		return nil
-	}
-	return s.spanContext.Tracestate
-}
-
 // SetName sets the name of the span, if it is recording events.
 func (s *Span) SetName(name string) {
 	if !s.IsRecordingEvents() {
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -88,6 +88,7 @@ type SpanContext struct {
 	TraceID      TraceID
 	SpanID       SpanID
 	TraceOptions TraceOptions
+	Tracestate   *Tracestate
 }
 
 type contextKey struct{}
@@ -190,8 +191,11 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
 
 	cfg := config.Load().(*Config)
 
-	if !hasParent {
+	if hasParent {
+		span.spanContext.Tracestate = parent.Tracestate.Fork()
+	} else {
 		span.spanContext.TraceID = cfg.IDGenerator.NewTraceID()
+		span.spanContext.Tracestate = &Tracestate{}
 	}
 	span.spanContext.SpanID = cfg.IDGenerator.NewSpanID()
 	sampler := cfg.DefaultSampler
@@ -287,6 +291,14 @@ func (s *Span) SpanContext() SpanContext {
 		return SpanContext{}
 	}
 	return s.spanContext
+}
+
+// Tracestate returns the Tracestate of the span context.
+func (s *Span) Tracestate() *Tracestate {
+	if s == nil {
+		return nil
+	}
+	return s.spanContext.Tracestate
 }
 
 // SetName sets the name of the span, if it is recording events.
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -191,10 +191,12 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
 
 	cfg := config.Load().(*Config)
 
-	if hasParent {
+	if !hasParent {
+		span.spanContext.TraceID = cfg.IDGenerator.NewTraceID()
+	}
+	if hasParent && parent.Tracestate != nil {
 		span.spanContext.Tracestate = parent.Tracestate.Fork()
 	} else {
-		span.spanContext.TraceID = cfg.IDGenerator.NewTraceID()
 		span.spanContext.Tracestate = &Tracestate{}
 	}
 	span.spanContext.SpanID = cfg.IDGenerator.NewSpanID()
--- a/trace/trace_test.go
+++ b/trace/trace_test.go
@@ -108,6 +108,7 @@ func TestSampling(t *testing.T) {
 				TraceID:      tid,
 				SpanID:       sid,
 				TraceOptions: test.parentTraceOptions,
+				Tracestate:   &Tracestate{},
 			}
 			ctx, _ = StartSpanWithRemoteParent(context.Background(), "foo", sc, WithSampler(test.sampler))
 		} else if test.localParent {
@@ -189,6 +190,7 @@ func TestStartSpanWithRemoteParent(t *testing.T) {
 		TraceID:      tid,
 		SpanID:       sid,
 		TraceOptions: 0x0,
+		Tracestate:   &Tracestate{},
 	}
 	ctx, _ := StartSpanWithRemoteParent(context.Background(), "startSpanWithRemoteParent", sc)
 	if err := checkChild(sc, FromContext(ctx)); err != nil {
@@ -204,6 +206,7 @@ func TestStartSpanWithRemoteParent(t *testing.T) {
 		TraceID:      tid,
 		SpanID:       sid,
 		TraceOptions: 0x1,
+		Tracestate:   &Tracestate{},
 	}
 	ctx, _ = StartSpanWithRemoteParent(context.Background(), "startSpanWithRemoteParent", sc)
 	if err := checkChild(sc, FromContext(ctx)); err != nil {
@@ -229,6 +232,7 @@ func startSpan(o StartOptions) *Span {
 			TraceID:      tid,
 			SpanID:       sid,
 			TraceOptions: 1,
+			Tracestate:   &Tracestate{},
 		},
 		WithSampler(o.Sampler),
 		WithSpanKind(o.SpanKind),
@@ -299,6 +303,7 @@ func TestSpanKind(t *testing.T) {
 					TraceID:      tid,
 					SpanID:       SpanID{},
 					TraceOptions: 0x1,
+					Tracestate:   &Tracestate{},
 				},
 				ParentSpanID:    sid,
 				Name:            "span0",
@@ -316,6 +321,7 @@ func TestSpanKind(t *testing.T) {
 					TraceID:      tid,
 					SpanID:       SpanID{},
 					TraceOptions: 0x1,
+					Tracestate:   &Tracestate{},
 				},
 				ParentSpanID:    sid,
 				Name:            "span0",
@@ -333,6 +339,7 @@ func TestSpanKind(t *testing.T) {
 					TraceID:      tid,
 					SpanID:       SpanID{},
 					TraceOptions: 0x1,
+					Tracestate:   &Tracestate{},
 				},
 				ParentSpanID:    sid,
 				Name:            "span0",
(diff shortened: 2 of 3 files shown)


reyang:
@rakyll @ramonza, would you give me some hints? I'm getting the following AppVeyor build error which claims gcc compiler missing:

runtime/cgo
**exec: "gcc": executable file not found in %PATH%**
FAIL	go.opencensus.io/exporter/jaeger [build failed]
FAIL	go.opencensus.io/exporter/prometheus [build failed]
FAIL	go.opencensus.io/exporter/stackdriver/propagation [build failed]
FAIL	go.opencensus.io/exporter/zipkin [build failed]
FAIL	go.opencensus.io/internal [build failed]
FAIL	go.opencensus.io/plugin/ocgrpc [build failed]
FAIL	go.opencensus.io/plugin/ochttp [build failed]
FAIL	go.opencensus.io/plugin/ochttp/propagation/b3 [build failed]
FAIL	go.opencensus.io/plugin/ochttp/propagation/tracecontext [build failed]
FAIL	go.opencensus.io/stats [build failed]
FAIL	go.opencensus.io/stats/view [build failed]
FAIL	go.opencensus.io/tag [build failed]
FAIL	go.opencensus.io/trace [build failed]
FAIL	go.opencensus.io/trace/propagation [build failed]
FAIL	go.opencensus.io/zpages [build failed]
Command exited with code 2

trace/tracestate.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (ts *TracestateEntry) IsValid() bool {
+	return KEY_VALIDATION_RE.MatchString(ts.key) &&
+		VALUE_VALIDATION_RE.MatchString(ts.value)
+}
+
+// TraceState is a tracing-system specific context in a list of key-value pairs.
+// TraceState allows different vendors propagate additional information and
+// inter-operate with their legacy Id formats.
+type Tracestate struct {
```
ramonza: Why is this not just a map?

trace/tracestate.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+const (
+	KEY_WITHOUT_VENDOR_FORMAT = `[a-z][_0-9a-z\-\*\/]{0,255}`
+	KEY_WITH_VENDOR_FORMAT    = `[a-z][_0-9a-z\-\*\/]{0,240}@[a-z][_0-9a-z\-\*\/]{0,13}`
+	KEY_FORMAT                = `(` + KEY_WITHOUT_VENDOR_FORMAT + `)|(` + KEY_WITH_VENDOR_FORMAT + `)`
+	VALUE_FORMAT              = `[\x20-\x2b\x2d-\x3c\x3e-\x7e]{0,255}[\x21-\x2b\x2d-\x3c\x3e-\x7e]`
+)
+
+var KEY_VALIDATION_RE = regexp.MustCompile(`^(` + KEY_FORMAT + `)$`)
```
ramonza: Go convention for variables is camelCase

trace/tracestate.go
```diff
@@ -0,0 +1,100 @@
...
+
+package trace
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const (
+	KEY_WITHOUT_VENDOR_FORMAT = `[a-z][_0-9a-z\-\*\/]{0,255}`
```
ramonza: Go convention for consts is camelCase - I think golint will pick this up

trace/trace.go
```diff
@@ -289,6 +295,14 @@ func (s *Span) SpanContext() SpanContext {
 	return s.spanContext
 }
 
+// Tracestate returns the Tracestate of the span context.
+func (s *Span) Tracestate() *Tracestate {
```
ramonza: Why isn't this a method on SpanContext?

trace/tracestate.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+	return ""
+}
+
+func (ts *Tracestate) Remove(key string) string {
+	return ts.Set(key, "")
+}
+
+func (ts *Tracestate) Set(key string, value string) string {
```
ramonza: I don't think we should make any part of SpanContext mutable. Can we make Tracestate immutable for now? I think we can figure out how to make it possible to change the Tracestate in a subsequent PR - I think the focus of this initial work should just be on correct propagation.

trace/tracestate.go
```diff
@@ -0,0 +1,100 @@
...
+	key   string
+	value string
+}
+
+func (ts *TracestateEntry) IsValid() bool {
+	return KEY_VALIDATION_RE.MatchString(ts.key) &&
+		VALUE_VALIDATION_RE.MatchString(ts.value)
+}
+
+// TraceState is a tracing-system specific context in a list of key-value pairs.
```
ramonza: Inconsistent capitalization

trace/tracestate.go
basvanbeek: because order of key-value pairs is important, see: 
https://w3c.github.io/distributed-tracing/trace_context/HTTP_HEADER_FORMAT_RATIONALE.html#ordering-of-keys-in-tracestate

reyang: @ramonza - @basvanbeek is right, the W3C spec requires tracestate to following a special ordering rule - last modified key-value-pair should appear at the very beginning.

trace/tracestate.go
reyang: Got it, I'm fixing it.

trace/tracestate.go
reyang: Sure, I'd be glad to change to camelCase.

I did searched the entire repo to see if there is a convention I could follow, and I'm seeing inconsistent conventions. And golint is not part of CI validation (all I see is gofmt). e.g.

In stats/validation.go, we have
```
const (
	MaxNameLength = 255
)
```

in vendor\git.apache.org\thrift.git\lib\go\thrift\application_exception.go (although this is 3rd party dependency), we have
```
const (
	UNKNOWN_APPLICATION_EXCEPTION  = 0
	UNKNOWN_METHOD                 = 1
	INVALID_MESSAGE_TYPE_EXCEPTION = 2
	WRONG_METHOD_NAME              = 3
	BAD_SEQUENCE_ID                = 4
	MISSING_RESULT                 = 5
	INTERNAL_ERROR                 = 6
	PROTOCOL_ERROR                 = 7
)
```

trace/tracestate.go
basvanbeek: @reyang I think with @ramonza questioning the choice I think it makes sense to add a comment in the code so it is clear to everybody without needing to deep dive into the specs.

trace/trace.go
reyang: The Span.SpanContext implementation will construct a new SpanContext if it is nil, which makes it hard for consumer to tell the difference whether Tracestate is nil in the existing SpanContext because it is nil or because the SpanContext is nil.

trace/tracestate.go
reyang: Will fix.

trace/tracestate.go
reyang: @ramonza, I've commented out the Tracestate.Remove and Tracestate.Set for now.

Regarding SpanContext, I'm putting a pointer to Tracestate rather than a hardy copy for Tracestate as we would want to have COW (copy-on-write) Tracestate in the future for performance, taking a hard copy is too expensive.

ramonza: Would you mind removing them entirely for now? We in general never keep around commented code.

trace/tracestate.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	for _, entry := range ts.entries {
+		if !entry.IsValid() {
+			return false
+		}
+	}
+	return true
+}
+
+func (ts *Tracestate) Fork() *Tracestate {
```
ramonza: I think we should remove this method as well. It only makes sense if Tracestate is mutable.

trace/tracestate.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if value != "" {
+		ts.entries = append([]TracestateEntry{newEntry}, ts.entries...)
+	}
+	return retval
+}
+*/
+
+func (ts *Tracestate) String() string {
+	return fmt.Sprintf("tracestate%s", ts.entries)
```
ramonza: I don't think this will be a useful output for String() - I would recommend just removing the method

trace/trace_test.go
```diff
@@ -510,6 +520,7 @@ func TestSetSpanStatus(t *testing.T) {
 			TraceID:      tid,
 			SpanID:       SpanID{},
 			TraceOptions: 0x1,
+			Tracestate:   &Tracestate{},
```
ramonza: The zero value of *Tracestate (nil) should be a valid empty Tracestate in the spirit of making the zero value useful. So it should not be necessary to set it here.

trace/tracestate.go
```diff
@@ -0,0 +1,106 @@
...
+	return true
+}
+
+func (ts *Tracestate) Fork() *Tracestate {
+	retval := Tracestate{entries: ts.entries}
+	return &retval
+}
+
+func (ts *Tracestate) Get(key string) string {
+	for _, entry := range ts.entries {
```
ramonza: To make the zero value of `*Tracestate*` useful, let's add 
```
if ts == nil {
return ""
}
```

trace/tracestate.go
```diff
@@ -0,0 +1,106 @@
...
+var keyValidationRegExp = regexp.MustCompile(`^(` + keyFormat + `)$`)
+var valueValidationRegExp = regexp.MustCompile(`^(` + valueFormat + `)$`)
+
+type TracestateEntry struct {
+	key   string
+	value string
+}
+
+func (ts *TracestateEntry) IsValid() bool {
+	return keyValidationRegExp.MatchString(ts.key) &&
```
ramonza: Is an empty key valid? Is an empty value? It seems from the pattern that it is. Should it be? Ideally we would have a test for this.

trace/tracestate.go
ramonza: Should this method distinguish between an empty string value and the absence of a key? If so, perhaps it should return `(string, bool)`?

trace/tracestate.go
ramonza: I think it would be reasonable for `String()` to produce a Tracestate header value in the format from the spec.

reyang:
@ramonza, agree, that's what I've mentioned in the PR description "My next pull request will implement the ochttp integration, which will propagate tracestate in HTTP headers.", I will add the actual propagation and the test cases in the next PR.

ramonza:
I would prefer to have it in one PR because I don't think just this alone makes sense on its own -- it doesn't add any functionality and I'm worried that if users see it they will be confused as to why its there.

ramonza:
I am going to look into the AppVeyor failure, it is unrelated.

trace/tracestate.go
reyang: Please refer to https://github.com/census-instrumentation/opencensus-python/pull/238#issuecomment-407322735 where we agreed to separate the in-process representation and the actual HTTP header format.

trace/tracestate.go
reyang: Empty string means the key doesn't exist, as the W3C spec wouldn't allow empty string as the value.

trace/tracestate.go
reyang: Sure, I'll remove it.

reyang:
Close this as it is a dup with #890 which @rghetia is working on.

Issue #883, opened 2018-08-28
Implement tracestate reading and propagation


SergeyKanzhelev:
See https://github.com/w3c/distributed-tracing for the protocol details.

ramonza:
I don't think this should actually be closed. The linked PR does not actually implement propagation, so this still needs to be done.

songy23:
Propagation is done in https://github.com/census-instrumentation/opencensus-go/pull/901.

PR #890, opened 2018-08-31
Add Tracestate to SpanContext

rghetia:
This is the first step to resolve https://github.com/census-instrumentation/opencensus-go/issues/883
I will post another PR to do the second step ( to propagate tracestate)


rghetia:
This is duplicate of https://github.com/census-instrumentation/opencensus-go/pull/887
Didn't realize that @reyang was also working on it.

reyang:
@rghetia, are you willing to work on this? I would be happy to revoke my PR #887 given we're doing very similar things, and you have already implemented test cases!
Let me know what you think.

rghetia:
yes, I'll continue then. I will take care of some of the comments you
received as it applies to mine as well.

-Rahul

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+
+import (
+	"container/list"
+	"errors"
+	"fmt"
+)
+
+const (
+	// KeyMaxSize is the maximum characters allowed in the key.
+	KeyMaxSize = 256
```
rakyll: Do we need to export these values? User shouldn't be able to change them.

keyMaxSize = 256

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// ValueMaxSize is the maximum characters allowed in the value.
+	ValueMaxSize = 256
```
rakyll: ditto

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// MaxKeyValuePairs is the maximum number of key-value pairs allowed in the tracestate.
+	MaxKeyValuePairs = 32
```
rakyll: ditto

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+	return nil, errors.New("invalid parameters")
+}
+
+// CreateTracestate creates a Tracestate object and adds a key-value pair to the list.
+// If a non-empty parent is passed then key/value pair from the parent is copied
+// to a newly created Tracestate object.
+// If the key already exists in the parent then its value is replaced with the
+// value passed to this function. The key is also moved to the front of
+// the list. See Set func.
+func CreateTracestate(parent *Tracestate, key, value string) (*Tracestate, error) {
```
rakyll: NewTracestate.

In Go, constructors are named like NewXXX.

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+	ValueMaxSize = 256
+
+	// MaxKeyValuePairs is the maximum number of key-value pairs allowed in the tracestate.
+	MaxKeyValuePairs = 32
+)
+
+// Tracestate represent tracing-system specific context in a list of key-value pairs. Tracestate allows different
+// vendors propagate additional information and inter-operate with their legacy Id formats.
+type Tracestate struct {
+	stateList *list.List
```
rakyll: Don't use the container package, it is not maintained by anyone.

Why do you need a list?

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// Entry represent one key-value pair in a list of key-value pair of Tracestate.
+// Key is an opaque string up to 256 characters printable. It MUST begin with a lowercase letter,
+// and can only contain lowercase letters a-z, digits 0-9, underscores _, dashes -, asterisks *, and
+// forward slashes /.
+//
+// Value is an opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the
+// range 0x20 to 0x7E) except comma , and =.
+type Entry struct {
```
rakyll: Remove Entry, this type is not required for external APIs and not required.

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+// If the key already exists in the parent then its value is replaced with the
+// value passed to this function. The key is also moved to the front of
+// the list. See Set func.
+func CreateTracestate(parent *Tracestate, key, value string) (*Tracestate, error) {
+
+	tracestate := Tracestate{
+		stateList: list.New(),
+	}
+
+	// Insert all entries from the
```
rakyll: Incomplete sentence.

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+		kv = e.Value.(*Entry)
+		if kv.key == key {
+			return kv.value
+		}
+	}
+	return ""
+}
+
+// GetEntries retrieves all key/value Entry from Tracestate ts.
+func (ts *Tracestate) GetEntries() *list.List {
```
rakyll: Entries

In Go, accessors are not prefixed with Get. That's why we call them accessors, not getters.

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+	return ts.stateList
+}
+
+// Remove removes the key/value Entry the Tracestate ts.
+func (ts *Tracestate) Remove(key string) {
```
reyang: Do we want to follow the List.Remove signature - return the item that has been removed?

```
 func (l *List) Remove(e *Element) interface{} {
  	if e.list == l {
  		// if e.list == l, l must have been initialized when e was inserted
  		// in l or l == nil (e is a zero Element) and l.remove will crash
  		l.remove(e)
  	}
  	return e.Value
  }
```

trace/tracestate.go
```diff
@@ -0,0 +1,183 @@
...
+		return
+	}
+	elemWithKey := ts.findElem(key)
+	if elemWithKey != nil {
+		ts.stateList.Remove(elemWithKey)
+	}
+}
+
+// Set inserts a key/value Entry in the front if there isn't
+// and existing Entry with the same key. If there exists an Entry
```
reyang: minor: and -> an (typo?)

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+	return &tracestate, nil
+}
+
+// NewTracestateFronParent creates a Tracestate object and adds a key-value pair to the list.
+// If a non-empty parent is passed then key/value pair from the parent is copied
+// to a newly created Tracestate object.
+// If the key already exists in the parent then its value is replaced with the
+// value passed to this function. The key is also moved to the front of
+// the list. See add func.
+func NewTracestateFronParent(parent *Tracestate, key, value string) (*Tracestate, error) {
```
rghetia: Typo in function name. Also function could be NewFromParent (remove Tracestate?)

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+			ts.tracestateEntries = append(ts.tracestateEntries[:index], ts.tracestateEntries[index+1:]...)
+			break
+		}
+	}
+}
+
+func (ts *Tracestate) add(entry *TracestateEntry) error {
+	ts.remove(entry.Key())
+	if len(ts.tracestateEntries) >= maxKeyValuePairs {
+		return fmt.Errorf("Set failed: reached maximum key/value pairs limit of %d", maxKeyValuePairs)
```
reyang: We might need an API for truncating the tracestate (removing key-value-pairs) from the tail.
Consider the following scenario:
1. I got a tracestate with 31 key-value-pairs (which is maxKeyValuePairs  - 1).
2. The trace library I'm using would want to add 2 additional key-value-pairs.
3. What the library would want to do, is to **remove the last key-value-pair from the existing tracestate**, and prepend the 2 additional key-value-pairs.

There might be other scenarios, for example, remove the key-value-pairs from the tail until the total size (depending on the serialization detail) is under a certain limit (e.g. "The length of a combined header MUST be less than or equal to 512 bytes. If the length of a combined header is more than 512 bytes it SHOULD be ignored.").

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+			return entry.Key(), true
+		}
+		keyMap[entry.Key()] = 1
+	}
+	return "", false
+}
+
+// NewFromEntryArray creates a Tracestate object from an array of key-value pair.
+// nil is returned with with an error if
+//  1. If the len of the entries  > maxKeyValuePairs
```
reyang: Minor, there is an extra space before ">" (two spaces).

trace/trace_test.go
```diff
@@ -17,14 +17,17 @@ package trace
 import (
 	"context"
 	"fmt"
+	"go.opencensus.io/trace/tracestate"
```
reyang: Minor, put custom (go.opencensus.io) packages below the standard packages, e.g.:

```
import (
	"log"
	"net/http"

	"go.opencensus.io/plugin/ochttp/propagation/b3"
	"go.opencensus.io/stats/view"
	"go.opencensus.io/tag"
)
```

trace/tracestate/tracestate.go
rghetia: Maybe add 'truncate' as a bool argument to NewXXX. If set then remove pairs from the tail.

#2 - would require that NewXXX takes more than one key-value pair. Alternatively. you can call NewXXX repeatedly but that will not be efficient.

Regarding 512 bytes - should that be handled in propagation or should it be part of truncation?

Do you want me to add the above to this PR or open up a new one?

trace/tracestate/tracestate.go
rghetia: fixed.

trace/trace_test.go
rakyll: +1

trace/trace_test.go
```diff
@@ -17,14 +17,17 @@ package trace
...
 	"time"
 )
 
 var (
-	tid = TraceID{1, 2, 3, 4, 5, 6, 7, 8, 1, 2, 4, 8, 16, 32, 64, 128}
-	sid = SpanID{1, 2, 4, 8, 16, 32, 64, 128}
+	tid               = TraceID{1, 2, 3, 4, 5, 6, 7, 8, 1, 2, 4, 8, 16, 32, 64, 128}
+	sid               = SpanID{1, 2, 4, 8, 16, 32, 64, 128}
+	defaultTracestate = tracestate.Tracestate{}
+	tracestate1, _    = tracestate.NewFromParent(nil, "foo", "bar")
```
rakyll: Want to give this a descriptive name? testTracestate?

trace/tracestate/tracestate.go
reyang: We can leave it for the next PR since it's not a blocking issue.

The 512 bytes limit is specific to the HTTP header format, which should not be part of the tracestate itself.

I mention it here to discuss if we need a "truncate" function. (do we want to propagator to truncate the key-value-pairs by calling a tracestate method, or directly access the key-value-pairs and discard from the tail?)

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+	keyWithoutVendorFormat = `[a-z][_0-9a-z\-\*\/]{0,255}`
+	keyWithVendorFormat    = `[a-z][_0-9a-z\-\*\/]{0,240}@[a-z][_0-9a-z\-\*\/]{0,13}`
+	keyFormat              = `(` + keyWithoutVendorFormat + `)|(` + keyWithVendorFormat + `)`
+	valueFormat            = `[\x20-\x2b\x2d-\x3c\x3e-\x7e]{0,255}[\x21-\x2b\x2d-\x3c\x3e-\x7e]`
+)
+
+var keyValidationRegExp = regexp.MustCompile(`^(` + keyFormat + `)$`)
+var valueValidationRegExp = regexp.MustCompile(`^(` + valueFormat + `)$`)
+
+// Tracestate represent tracing-system specific context in a list of key-value pairs. Tracestate allows different
```
rakyll: Tracestate represents...

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// TracestateEntry represent one key-value pair in a list of key-value pair of Tracestate.
+// Key is an opaque string up to 256 characters printable. It MUST begin with a lowercase letter,
+// and can only contain lowercase letters a-z, digits 0-9, underscores _, dashes -, asterisks *, and
+// forward slashes /.
+//
+// Value is an opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the
+// range 0x20 to 0x7E) except comma , and =.
+type TracestateEntry struct {
```
rakyll: Do we need this type? We can represent everything as <string, string>.

In Go, introduction of new types are not encouraged if builtin types are sufficient. You can write the NewFromTracestateEntries as:

```
NewFromKeyVals(map[string]string) (*Tracestate, error)
```

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return "", false
+}
+
+// NewFromEntryArray creates a Tracestate object from an array of key-value pair.
+// nil is returned with with an error if
+//  1. If the len of the entries > maxKeyValuePairs
+//  2. If the entries contain duplicate keys
+func NewFromEntryArray(entries []*TracestateEntry) (*Tracestate, error) {
+
```
rakyll: Remove line.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Get retrieves value for a given key from Tracestate ts.
+// If the key is not found then false is returned with the value "" is returned.
+// If the key is found then true is returned with its value.
+func (ts *Tracestate) Get(key string) (string, bool) {
+	if len(ts.tracestateEntries) == 0 {
+		return "", false
+	}
+	for _, entry := range ts.tracestateEntries {
```
rakyll: Why isn't we keep all the entries in a map. This is an O(n) lookup.

trace/tracestate/tracestate.go
rghetia: with map[string]string order is not maintained.

rakyll: It was useful to know that order matters, thanks.

In Go, we don't stutter. Let's rename TracestateEntry to just Entry. Then it will look like tracestate.Entry at the call site.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+
+// TracestateEntry represent one key-value pair in a list of key-value pair of Tracestate.
+// Key is an opaque string up to 256 characters printable. It MUST begin with a lowercase letter,
+// and can only contain lowercase letters a-z, digits 0-9, underscores _, dashes -, asterisks *, and
+// forward slashes /.
+//
+// Value is an opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the
+// range 0x20 to 0x7E) except comma , and =.
+type TracestateEntry struct {
+	key   string
```
rakyll: What about keeping this type simple?

type Entry { Key, Value string }

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+	valueFormat            = `[\x20-\x2b\x2d-\x3c\x3e-\x7e]{0,255}[\x21-\x2b\x2d-\x3c\x3e-\x7e]`
+)
+
+var keyValidationRegExp = regexp.MustCompile(`^(` + keyFormat + `)$`)
+var valueValidationRegExp = regexp.MustCompile(`^(` + valueFormat + `)$`)
+
+// Tracestate represent tracing-system specific context in a list of key-value pairs. Tracestate allows different
+// vendors propagate additional information and inter-operate with their legacy Id formats.
+type Tracestate struct {
+	tracestateEntries []*TracestateEntry
```
rakyll: Just say entries, it is very mouthful to read tracestateEntiries. In Go, we prefer short and descriptive identifiers as much as possible.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+	for _, entry := range ts.tracestateEntries {
+		if entry.key == key {
+			return entry.value, true
+		}
+	}
+	return "", false
+}
+
+// TraceEntries returns a slice of TracestateEntry.
+func (ts *Tracestate) TraceEntries() []*TracestateEntry {
```
rakyll: s/TraceEntries/Entries

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func isValid(key, value string) bool {
+	return keyValidationRegExp.MatchString(key) &&
+		valueValidationRegExp.MatchString(value)
+}
+
+// NewTracestateEntry creates a TracestateEntry object with given key and value.
+// It returns error if either key or value is invalid.
+func NewTracestateEntry(key, value string) (*TracestateEntry, error) {
```
rakyll: You don't need that if you make Entry

```
type Entry struct{ Key, Value string }
```

Then, user can construct entries as Entry{Key: key, Value: value}

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+		keyMap[entry.Key()] = 1
+	}
+	return "", false
+}
+
+// NewFromEntryArray creates a Tracestate object from an array of key-value pair.
+// nil is returned with with an error if
+//  1. If the len of the entries > maxKeyValuePairs
+//  2. If the entries contain duplicate keys
+func NewFromEntryArray(entries []*TracestateEntry) (*Tracestate, error) {
```
rakyll: NewFromEnties(e ...Entry) 

Don't use a pointer to the entry but the concrete type here, then you immutability issues go away.

Use variadic arguments as much as possible, it makes it easy to pass a single entry.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+
+	if entries == nil {
```
rakyll: This case should be a noop if you ask my opinion.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, errors.New("Invalid parameter")
+	}
+	if len(entries) == 0 || len(entries) > maxKeyValuePairs {
+		return nil, fmt.Errorf("Invalid number of tracestateEntry (%d)", len(entries))
```
rakyll: In Go, error strings are not capitalized: 

fmt.Errorf("number of entries (%d) is larger than the max (%d)", len(entries), maxKeyValuePairs)

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if key, duplicate := containsDuplicateKey(entries); duplicate {
+		return nil, fmt.Errorf("Contains duplicate keys (%s)", key)
```
rakyll: Ditto.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+	return &tracestate, nil
+}
+
+// NewFromParent creates a Tracestate object and adds a key-value pair to the list.
+// If a non-empty parent is passed then key/value pair from the parent is copied
+// to a newly created Tracestate object.
+// If the key already exists in the parent then its value is replaced with the
+// value passed to this function. The key is also moved to the front of
+// the list. See add func.
+func NewFromParent(parent *Tracestate, key, value string) (*Tracestate, error) {
```
rakyll: Why not NewFromParent(parent *Tracestate, e ...Entry) ?

trace/trace_test.go
rghetia: fixed.

trace/trace_test.go
rghetia: done.

trace/tracestate/tracestate.go
rghetia: done.

trace/tracestate/tracestate.go
rghetia: We need to maintain ordered list of key-value pairs. Yes, map would be efficient in searching. We would need both. we can keep map[string]int for storing key-arrayIndex to improve the search. But the number of key-value pair is not expected to be high (based on comment from Java).
I can certainly make the change to add the key/arrayIndex for efficient search.

trace/tracestate/tracestate.go
rghetia: done.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,182 @@
...
+
```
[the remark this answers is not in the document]
rghetia: done.

trace/tracestate/tracestate_test.go
```diff
@@ -0,0 +1,290 @@
...
+// limitations under the License.
+
+package tracestate
+
+import (
+	"fmt"
+	"testing"
+)
+
+func init() {
```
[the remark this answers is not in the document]
rghetia: done.

trace/tracestate/tracestate_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+}
+
+func checkError(t *testing.T, tracestate *Tracestate, err error, testname, msg string) {
+	if err != nil {
+		t.Errorf("Test:%s: %s: tracestate=%v, error= %v", testname, msg, tracestate, err)
+	}
+}
+
+func expectError(t *testing.T, tracestate *Tracestate, err error, testname, msg string) {
```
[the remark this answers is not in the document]
rghetia: changed it to wantError.

trace/tracestate/tracestate_test.go
```diff
@@ -0,0 +1,290 @@
...
+	testname := "TestNilKey"
+
+	tracestate, err := NewFromParent(nil, "", "world")
+	expectError(t, tracestate, err, testname, "Create Tracesate did not err with nil key (\"\")")
+}
+
+func TestValueWithInvalidChar(t *testing.T) {
+	testname := "TestNilKey"
+
+	// Invalid characters comma, equal, slash, star
```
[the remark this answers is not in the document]
rghetia: removed the comments.

trace/tracestate/tracestate.go
rghetia: got it. done.

trace/tracestate/tracestate.go
rghetia: I chose pointer to make an inserting/removing entries in array efficient. Do you think simplicity out weighs the gain in efficiency here?

trace/tracestate/tracestate.go
rghetia: done.

trace/tracestate/tracestate.go
rghetia: With the current approach key/value is validated when Entry is created. If we go with simple route then in following constructor for Tracestate all keys would have to be validated.
NewFromArray([]Entry) *Tracestate {}

rakyll: What's wrong with that approach. That's the pattern we used for the tag package and we should follow the same pattern here to fight verbosity of error handling.

trace/tracestate/tracestate.go
rakyll: What do you mean? In most cases, Entry structs will be allocated on the stack if it is a concrete type.

trace/tracestate/tracestate.go
rghetia: ok. done.

trace/tracestate/tracestate.go
rghetia: ignore my comments. my understanding was wrong.
I'll change it to concrete type.

trace/tracestate/tracestate.go
rghetia: done.

trace/tracestate/tracestate.go
basvanbeek: I indeed don't expect this number to be high... ask yourself... will we expect more than a couple of different tracing systems collaborating in a request be a common use case? So I don't think we need additional code for more efficient search at this point (premature optimization).

trace/tracestate/tracestate.go
basvanbeek: @reyang can you explain the use case of needing to add 2 additional key-value pairs? 

Typically a tracing system should only encode a single key-value pair where key is their vendor label and value typically holds (embeds) multiple values important to that vendor's propagation format. 

Example: Zipkin old style B3 has multiple key-value pairs as separate headers but in W3C's tracestate it will use single value (thus multi embedded values in header value): `b3={x-b3-traceid}-{x-b3-spanid}-{if x-b3-flags 'd' else x-b3-sampled}-{x-b3-parentspanid}`

Also see spec example of 2 different tracing systems as it highlights single k/v per tracing system: https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#examples-of-http-headers-1

reyang: @basvanbeek I was thinking about the migration scenario:
1. You have service A, B and C. A and B are using a legacy protocol X for tracing, while B and C use another one Y.
2. B and C are owned by you (which means you can change the code and do deployments), while A is not (e.g. external service).
3. By upgrading the protocol between B and C to W3C distributed tracing, you want to carry the existing information in Tracestate during the transition period.
4. This would end up with X and Y in the Tracestate header for requests issued from service B to C.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,168 @@
...
+type Entry struct {
+	Key   string
+	Value string
+}
+
+// Get retrieves value for a given key from Tracestate ts.
+// If the key is not found then false is returned with the value "".
+// If the key is found then true is returned with its value.
+func (ts *Tracestate) Get(key string) (string, bool) {
+	if len(ts.entries) == 0 {
```
ramonza: Make the zero value of `*Tracestate`, `nil`, an empty `*Tracestate` by checking if `ts == nil` here. This way, if the user never uses Tracestate (probably the majority of the time), then we can avoid every allocating the struct.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil, true
+}
+
+// NewFromEntries creates a Tracestate object from an array of key-value pair.
+// nil is returned with an error if
+//  1. the len of the entries > maxKeyValuePairs
+//  2. the entries contain duplicate keys
+//  3. one or more entries are invalid.
+func NewFromEntries(entries []Entry) (*Tracestate, error) {
+	if len(entries) == 0 || len(entries) > maxKeyValuePairs {
```
ramonza: Why is 0 an invalid number of entries?

ramonza: I recommend that we consider `nil` to represent a valid, empty Tracestate as I mentioned in my previous comment. Then we can simply return `nil, nil`, if `len(entries) == 0`.

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// NewFromParent creates a Tracestate object from a parent and an entry (key-value pair).
+// Entries from the parent are copied if present and the entry passed to this function
+// is inserted in front of the list. If there exists any entry with the same key it is
+// removed. See add func.
+// An error is returned with nil Tracestate if
+//  1. the entry is invalid.
+//  2. the number of entries exceeds maxKeyValuePairs.
+func NewFromParent(parent *Tracestate, entry Entry) (*Tracestate, error) {
```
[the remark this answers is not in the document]
rghetia: 1. Should the 'New' be part of the function name as it is a constructor for tracestate?
2. when parent is not nil, should the entries be restricted to one entry? The reason is a tracing system should never be adding two kv-pair.

trace/tracestate/tracestate.go
rghetia: nil is already a valid Tracestate. I'll remove the len == 0 check.

trace/tracestate/tracestate.go
rghetia: I did not realize the you can call the method when ts is nil. I'll make the change.

trace/tracestate/tracestate_test.go
```diff
@@ -0,0 +1,320 @@
...
+func checkBack(t *testing.T, tracestate *Tracestate, wantKey, testname string) {
+	gotKey := tracestate.entries[len(tracestate.entries)-1].Key
+	if gotKey != wantKey {
+		t.Errorf("test:%s: last entry in the list: got %q want %q", testname, gotKey, wantKey)
+	}
+}
+
+func checkSize(t *testing.T, tracestate *Tracestate, wantSize int, testname string) {
+	gotSize := len(tracestate.entries)
+	if gotSize != wantSize {
```
rakyll: You can write these inline.

if got, want := len(tracestate.entries), wantSize  {}

trace/tracestate/tracestate.go
```diff
@@ -0,0 +1,142 @@
...
+// are inserted in front of those copied from the parent. If an entry copied from the
+// parent contains the same key as one of the entry in entries then the entry copied
+// from the parent is removed. See add func.
+//
+// An error is returned with nil Tracestate if
+//  1. one or more entry in entries is invalid.
+//  2. two or more entries in the input entries have the same key.
+//  3. the number of entries combined from the parent and the input entries exceeds maxKeyValuePairs.
+//     (duplicate entry is counted only once).
+func New(parent *Tracestate, entries []Entry) (*Tracestate, error) {
```
rakyll: Use entries as variadic arguments for syntactic sugar.

PR #896, opened 2018-09-04
zpages: fix countFormatter code and add tests

odeke-em:
Fix countFormatter code due to a bug that
I advertently introduced with a suggestion
in PR 894, due to change of signs:
from >= to <=
it should have just been the int64 cast.

However, also fix a long standing bug in the
original code for the case num >= 12, where
the divisor was copy-pasted as 1e9 instead
of 1e12 for Tera* values.

This change also adds tests to ensure such
problems could be caught in the first place
but that we never regress.

### DIFF
--- a/zpages/formatter_test.go
+++ b/zpages/formatter_test.go
@@ -1,0 +1,40 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


songy23:
FYI don't worry about the Appveyor build, it's a known issue.

odeke-em:
Thanks for the review @songy23! Hmm, same AppVeyor problem /cc @rakyll for help with that.

odeke-em:
Cool, thanks!

odeke-em:
@sipian please refetch the latest code from master.

Issue #892, opened 2018-09-01
ochttp and standard library docs inconsistency


nmiculinic:
From godoc documentation:

```
type RoundTripper interface {
        // RoundTrip executes a single HTTP transaction, returning
        // a Response for the provided Request.
        //
        // RoundTrip should not attempt to interpret the response. In
        // particular, RoundTrip must return err == nil if it obtained
        // a response, regardless of the response's HTTP status code.
        // A non-nil err should be reserved for failure to obtain a
        // response. Similarly, RoundTrip should not attempt to
        // handle higher-level protocol details such as redirects,
        // authentication, or cookies.
        //
        // RoundTrip should not modify the request, except for
        // consuming and closing the Request's Body. RoundTrip may
        // read fields of the request in a separate goroutine. Callers
        // should not mutate or reuse the request until the Response's
        // Body has been closed.
        //
        // RoundTrip must always close the body, including on errors,
        // but depending on the implementation may do so in a separate
        // goroutine even after RoundTrip returns. This means that
        // callers wanting to reuse the body for subsequent requests
        // must arrange to wait for the Close call before doing so.
        //
        // The Request's URL and Header fields must be initialized.
        RoundTrip(*Request) (*Response, error)
}
```

More exactly: ```RoundTrip should not modify the request, except for consuming and closing the Request's Body```

and ```&ochttp.Transport{}``` clearly modified the request by adding additional headers.

Thus I want to hear the rationale behind this decision (or go's standard library). Personally, I agree with ochttp's approach and perhaps go's documentation should indicate it's ok to change/add headers because there's no other way to implement client middleware that I know of. 

This issue is primarily to raise awareness of this inconsistency.

rakyll:
Thanks much for catching this bug. traceTransport should shallow copy the request and modify it rather than modifying the original request. RoundTrip should be safe for concurrent use for the same request object, that's why the interface enforces this constraint.

PR #898, opened 2018-09-05
ochttp.Transport should not modify request

ramonza:
Fixes: #892

### DIFF
--- a/plugin/ochttp/trace.go
+++ b/plugin/ochttp/trace.go
@@ -66,6 +66,16 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 
 	if t.format != nil {
+		// SpanContextToRequest will modify its Request argument, which is
+		// contrary to the contract for http.RoundTripper, so we need to
+		// pass it a copy of the Request.
+		// However, the Request struct itself was already copied by
+		// the WithContext calls above and so we just need to copy the header.
+		header := make(http.Header)
+		for k, v := range req.Header {
+			header[k] = v
+		}
+		req.Header = header
 		t.format.SpanContextToRequest(span.SpanContext(), req)
 	}
 
--- a/plugin/ochttp/trace_test.go
+++ b/plugin/ochttp/trace_test.go
@@ -79,6 +79,26 @@ func (t testPropagator) SpanContextToRequest(sc trace.SpanContext, req *http.Req
 	req.Header.Set("trace", hex.EncodeToString(buf.Bytes()))
 }
 
+func TestTransport_RoundTrip_Race(t *testing.T) {
+	// This tests that we don't modify the request in accordance with the
+	// specification for http.RoundTripper.
+	// We attempt to trigger a race by reading the request from a separate
+	// goroutine. If the request is modified by Transport, this should trigger
+	// the race detector.
+
+	transport := &testTransport{ch: make(chan *http.Request, 1)}
+	rt := &Transport{
+		Propagation: &testPropagator{},
+		Base:        transport,
+	}
+	req, _ := http.NewRequest("GET", "http://foo.com", nil)
+	go func() {
+		fmt.Println(*req)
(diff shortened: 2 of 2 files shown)


plugin/ochttp/trace.go
```diff
@@ -66,6 +66,11 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 
 	if t.format != nil {
+		header := make(http.Header)
```
rakyll: Copy paste the request cloning from https://github.com/golang/oauth2/blob/master/transport.go#L108:6.

ramonza: I intentionally did not copy the Request struct itself because it would already have been copied by the WithContext calls directly above this. I added a comment to this effect.

ramonza:
Comment added to clarify intent. @rakyll PTAL

plugin/ochttp/trace.go
```diff
@@ -66,6 +66,16 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 
 	if t.format != nil {
+		// SpanContextToRequest will modify its Request argument, which is
+		// contrary to the contract for http.RoundTripper, we therefore should
+		// create a copy of the Request.
+		// However, the Request struct itself would have already been copied by
```
rakyll: "have already been copied" or "is already copied"?

plugin/ochttp/trace_test.go
```diff
@@ -79,6 +79,26 @@ func (t testPropagator) SpanContextToRequest(sc trace.SpanContext, req *http.Req
...
+	// the race detector.
+
+	transport := &testTransport{ch: make(chan *http.Request, 1)}
+	rt := &Transport{
+		Propagation: &testPropagator{},
+		Base:        transport,
+	}
+	req, _ := http.NewRequest("GET", "http://foo.com", nil)
+	go func() {
+		fmt.Println(*req)
```
odeke-em: Sorry am late to the party but I don't think that our tests should be spamming stdout. If we'd like to do that, perhaps use t.Logf.

However, could we actually test that the headers are unmodified after the ochttp.Transport roundtrip?

ramonza: This doesn't actually go to stdout when running `go test` so I don't really see the harm. 

The advantage of this is that we test whether any part of the request (not just the headers) was modified. But in reality you're right that the headers are what actually was being modified.

odeke-em: >This doesn't actually go to stdout when running go test so I don't really see the harm.

The harm is spamming stdout. Currently you don't see it but that's because that goroutine is racy(i.e. no synchronization and it isn't guaranteed to run by the time the test exits) and did the race detector trip out at runtime without the fix?

If it were https://play.golang.org/p/Ur72IrWCTQY aka 
```go
package main

import (
	"fmt"
	"testing"
)

func TestFoo(t *testing.T) {
	done := make(chan bool)
	go func() {
		fmt.Println("foo")
		done <- true
	}()
	<-done
}
```
You'd def see the output. Please remove that fmt.Println and we can do
```go
_ = *req
```
if we really would like to keep it :)

Also perhaps for real race triggering at runtime ensure that that that goroutine is actually synchronized and runs concurrently with the code outside it.

ramonza: It only gets printed by "go test -v". Try it.

The code as-is repro'ed the race before I fixed it.

rakyll: We don't need to print this. Most users always use -v when running tests including our CI. We shouldn't pollute the output with arbitrary output.

ramonza: I don't really have strong feelings about this but this PR is merged. The right way to proceed is to open a new issue or a PR for this. Commenting on merged PRs is very difficult to track, I would prefer if we avoid this in future.

Issue #903, opened 2018-09-07
exporter/jaeger: ochttp trace status code is doubly quoted


rafaeljesus:
**Describe the bug**
Whether a request is traced via ochttp  the status.message value is always wrapped by `""OK""` in Jaeger UI. ATM I have to `status.message="\"OK"\"`

**Expected behavior**
In Jaeger UI I should be able to search traces filtering by `status.message=OK`

PR #905, opened 2018-09-08
Fix http status code string format

rafaeljesus:
Fix an issue when visualising a trace in Jaeger UI, the http.message field is equal to `""OK""` making harder to filter by tags `status.message="\"OK"\"`

Closes https://github.com/census-instrumentation/opencensus-go/issues/903


rafaeljesus:
I've noticed the unit tests failed, so now I am wondering if this behaviour is expected 🤔

PR #908, opened 2018-09-10
Fix the race in ochttp

rakyll:
Deep copy the headers, also create a racy test case
without having to print the request.

### DIFF
--- a/plugin/ochttp/trace_test.go
+++ b/plugin/ochttp/trace_test.go
@@ -94,9 +94,7 @@ func TestTransport_RoundTrip_Race(t *testing.T) {
 	req, _ := http.NewRequest("GET", "http://foo.com", nil)
 
 	go func() {
-		for i := 0; i < 1000; i++ {
-			req.Header.Add("trace", "another value")
-		}
+		fmt.Println(*req)
 	}()
 	rt.RoundTrip(req)
 
--- a/plugin/ochttp/trace_test.go
+++ b/plugin/ochttp/trace_test.go
@@ -76,7 +76,7 @@ func (t testPropagator) SpanContextToRequest(sc trace.SpanContext, req *http.Req
 	buf.Write(sc.TraceID[:])
 	buf.Write(sc.SpanID[:])
 	buf.WriteByte(byte(sc.TraceOptions))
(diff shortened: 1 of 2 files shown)


plugin/ochttp/trace_test.go
```diff
@@ -92,10 +92,14 @@ func TestTransport_RoundTrip_Race(t *testing.T) {
 		Base:        transport,
 	}
 	req, _ := http.NewRequest("GET", "http://foo.com", nil)
+
```
odeke-em: Could we add a synchronizer to ensure that the racer goroutine terminates
before we exit the test?

```go
var wg sync.WaitGroup
wg.Add(1)

go func() {
   defer wg.Done()
   
   for i := 0; i < 1000; i++ {
     req.Header.Add("trace", "another one")
   }
}()
rt.RoundTrip(req)

_ = <-transport.ch
wg.Wait()
```

plugin/ochttp/trace_test.go
```diff
@@ -92,10 +92,14 @@ func TestTransport_RoundTrip_Race(t *testing.T) {
...
 	go func() {
-		fmt.Println(*req)
+		for i := 0; i < 1000; i++ {
+			req.Header.Add("trace", "another value")
```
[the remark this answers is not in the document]
rakyll: This is a limitation of the RoundTripper contract. @bradfitz is the person who should respond to this question. I am using his reference implementation from 4 years ago. https://github.com/golang/oauth2/commit/a568078818bb97b96ee7c7829b4575b81d4684f1#diff-0fecf8d8a1c3b52ef05e62d6a1bf6effL149

plugin/ochttp/trace_test.go
rakyll: We still cannot guarantee if the gorotutine started before RoundTrip returned. We are trying to create a race that is hard to produce here.

ramonza:
@rakyll would you mind removing the failing test, I think it's fine that we just commit this without it

bogdandrutu:
@rakyll can you please update this PR to fix the test and probably rebase?

rakyll:
@bogdandrutu, rebased.

rakyll:
Closing due to inactivity, reopen if needed.

PR #915, opened 2018-09-17
Allow other endpoint urls than the default

achandras:
This will allow configuring Thrift endpoints that are not at Jaeger's default `/api/traces`

### DIFF
--- a/exporter/jaeger/jaeger.go
+++ b/exporter/jaeger/jaeger.go
@@ -39,6 +39,10 @@ type Options struct {
 	// For example, http://localhost:14268.
 	Endpoint string
 
+	// EndpointPath is the url path where spans are sent.
+	// For example, /api/traces
+	EndpointPath string
+
 	// AgentEndpoint instructs exporter to send spans to jaeger-agent at this address.
 	// For example, localhost:6831.
 	AgentEndpoint string
@@ -76,7 +80,12 @@ func NewExporter(o Options) (*Exporter, error) {
 	var client *agentClientUDP
 	var err error
 	if endpoint != "" {
-		endpoint = endpoint + "/api/traces?format=jaeger.thrift"
+		if o.EndpointPath != "" {
+			endpoint = endpoint + o.EndpointPath
+		} else {
+			endpoint = endpoint + "/api/traces"
+		}
+		endpoint = endpoint + "?format=jaeger.thrift"
 	} else {
 		client, err = newAgentClientUDP(o.AgentEndpoint, udpPacketMaxLength)
 		if err != nil {
--- a/exporter/jaeger/jaeger.go
+++ b/exporter/jaeger/jaeger.go
@@ -39,10 +39,9 @@ type Options struct {
 	// For example, http://localhost:14268.
 	Endpoint string
 
-	// EndpointPath is the url path where spans are sent.
-	// If this is not set, it will send to "/api/traces" as the default
-	// For example, /api/traces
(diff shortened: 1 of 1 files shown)


achandras:
I signed it!

exporter/jaeger/jaeger.go
```diff
@@ -39,6 +39,10 @@ type Options struct {
 	// For example, http://localhost:14268.
 	Endpoint string
 
+	// EndpointPath is the url path where spans are sent.
```
ramonza: Document the default value if empty.

exporter/jaeger/jaeger.go
```diff
@@ -39,6 +39,11 @@ type Options struct {
...
+	// If this is not set, it will send to "/api/traces" as the default
+	// For example, /api/traces
+	EndpointPath string
```
ramonza: After looking at the rest of the options, I think it is a little confusing that there is both `Endpoint` and `EndpointPath`. The question that occurred to me while looking at the new API is: why not include the path in the Endpoint option instead of having a separate `EndpointPath` option? 

Because of backwards compatibility, we cannot change the meaning of `Endpoint`. So instead I would like to recommend we introduce a new option: `CollectorEndpoint string`. If set, this will be the full URL (including the path) to the collector. We then deprecate `Endpoint` in favor of `CollectorEndpoint`. 

`CollectorEndpoint` is also more descriptive fits better with `AgentEndpoint`.

What do you think?

achandras: I agree that `CollectorEndpoint` seems a lot clearer than `Endpoint` + `EndpointPath`. I will make this change

Do you think it's a good idea to add a log message now that `Endpoint` is deprecated? Or should that be a change for later?

rakyll:
We have an Endpoint option? Shouldn't this setting allow us to only to configure the path?

rakyll:
Let's deprecate Endpoint as a part of this PR if that sounds good.

exporter/jaeger/jaeger.go
```diff
@@ -37,8 +37,13 @@ const defaultServiceName = "OpenCensus"
 type Options struct {
 	// Endpoint is the Jaeger HTTP Thrift endpoint.
 	// For example, http://localhost:14268.
+	// Deprecated: Use CollectorEndpoint instead.
```
rakyll: Add a new line before the deprecation notice to make it stand out on godoc.

```
// Endpoint is the Jaeger HTTP Thrift endpoint.
// For example, http://localhost:14268.		// For example, http://localhost:14268.
//
// Deprecated: Use CollectorEndpoint instead.
```

achandras: Added the new line

exporter/jaeger/jaeger.go
```diff
@@ -68,15 +72,17 @@ type Options struct {
...
+	if o.Endpoint == "" && o.CollectorEndpoint == "" && o.AgentEndpoint == "" {
 		return nil, errors.New("missing endpoint for Jaeger exporter")
 	}
 
+	var endpoint string
 	var client *agentClientUDP
 	var err error
-	if endpoint != "" {
-		endpoint = endpoint + "/api/traces?format=jaeger.thrift"
+	if o.Endpoint != "" {
```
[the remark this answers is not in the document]
ramonza: I nevermind its in Exporter.upload and it references the Exporter field "endpoint" which should be set correctly here.

Issue #873, opened 2018-08-21
Exemplar support


ramonza:
**Describe the solution you'd like**
Examplar support in view.Data for exporters.

See: https://github.com/census-instrumentation/opencensus-specs/pull/122

PR #917, opened 2018-09-17
Exemplar implementation

ramonza:
Fixes: #873

### DIFF
--- a/stats/doc.go
+++ b/stats/doc.go
@@ -21,35 +21,49 @@ aggregate the collected data, and export the aggregated data.
 
 Measures
 
-A measure represents a type of metric to be tracked and recorded.
+A measure represents a type of data point to be tracked and recorded.
 For example, latency, request Mb/s, and response Mb/s are measures
 to collect from a server.
 
-Each measure needs to be registered before being used. Measure
-constructors such as Int64 and Float64 automatically
+Measure constructors such as Int64 and Float64 automatically
 register the measure by the given name. Each registered measure needs
 to be unique by name. Measures also have a description and a unit.
 
-Libraries can define and export measures for their end users to
-create views and collect instrumentation data.
+Libraries can define and export measures. Application authors can then
+create views and collect and break down measures by the tags they are
+interested in.
 
 Recording measurements
 
 Measurement is a data point to be collected for a measure. For example,
 for a latency (ms) measure, 100 is a measurement that represents a 100ms
-latency event. Users collect data points on the existing measures with
+latency event. Measurements are created from measures with
 the current context. Tags from the current context are recorded with the
 measurements if they are any.
 
-Recorded measurements are dropped immediately if user is not aggregating
-them via views. Users don't necessarily need to conditionally enable/disable
+Recorded measurements are dropped immediately if no views are registered for them.
+There is usually no need to conditionally enable and disable
 recording to reduce cost. Recording of measurements is cheap.
 
-Libraries can always record measurements, and end-users can later decide
+Libraries can always record measurements, and applications can later decide
 on which measurements they want to collect by registering views. This allows
 libraries to turn on the instrumentation by default.
+
+Exemplars
+
+For a given recorded measurement, the associated exemplar is a diagnostic map
+that gives more information about the measurement.
+
+When aggregated using a Distribution aggregation, an exemplar is kept for each
+bucket in the Distribution. This allows you to easily find an example of a
+measurement that fell into each bucket.
+
+For example, if you also use the OpenCensus trace package and you
+record a measurement with a context that contains a sampled trace span,
+then the trace span will be added to the exemplar associated with the measurement.
+
+When exported to a supporting back end, you should be able to easily navigate
+to example traces that fell into each bucket in the Distribution.
+
 */
 package stats // import "go.opencensus.io/stats"
-
-// TODO(acetechnologist): Add a link to the language independent OpenCensus
-// spec when it is available.
--- a/stats/view/aggregation_data.go
+++ b/stats/view/aggregation_data.go
@@ -17,6 +17,8 @@ package view
 
 import (
 	"math"
+
+	"go.opencensus.io/exemplar"
 )
 
 // AggregationData represents an aggregated value from a collection.
@@ -24,7 +26,7 @@ import (
 // Mosts users won't directly access aggregration data.
 type AggregationData interface {
 	isAggregationData() bool
-	addSample(v float64)
+	addSample(e *exemplar.Exemplar)
 	clone() AggregationData
 	equal(other AggregationData) bool
 }
@@ -41,7 +43,7 @@ type CountData struct {
 
 func (a *CountData) isAggregationData() bool { return true }
 
-func (a *CountData) addSample(v float64) {
+func (a *CountData) addSample(_ *exemplar.Exemplar) {
 	a.Value = a.Value + 1
 }
 
@@ -68,8 +70,8 @@ type SumData struct {
 
 func (a *SumData) isAggregationData() bool { return true }
 
-func (a *SumData) addSample(f float64) {
-	a.Value += f
+func (a *SumData) addSample(e *exemplar.Exemplar) {
+	a.Value += e.Value
 }
 
 func (a *SumData) clone() AggregationData {
@@ -88,22 +90,30 @@ func (a *SumData) equal(other AggregationData) bool {
 // Distribution aggregation.
 //
 // Most users won't directly access distribution data.
+//
+// For a distribution with N bounds, the associated DistributionData will have
+// N+1 buckets.
 type DistributionData struct {
-	Count           int64     // number of data points aggregated
-	Min             float64   // minimum value in the distribution
-	Max             float64   // max value in the distribution
-	Mean            float64   // mean of the distribution
-	SumOfSquaredDev float64   // sum of the squared deviation from the mean
-	CountPerBucket  []int64   // number of occurrences per bucket
-	bounds          []float64 // histogram distribution of the values
+	Count           int64   // number of data points aggregated
+	Min             float64 // minimum value in the distribution
+	Max             float64 // max value in the distribution
+	Mean            float64 // mean of the distribution
+	SumOfSquaredDev float64 // sum of the squared deviation from the mean
+	CountPerBucket  []int64 // number of occurrences per bucket
+	// ExemplarsPerBucket is slice the same length as CountPerBucket containing
+	// an exemplar for the associated bucket, or nil.
+	ExemplarsPerBucket []*exemplar.Exemplar
+	bounds             []float64 // histogram distribution of the values
 }
 
 func newDistributionData(bounds []float64) *DistributionData {
+	bucketCount := len(bounds) + 1
 	return &DistributionData{
-		CountPerBucket: make([]int64, len(bounds)+1),
-		bounds:         bounds,
-		Min:            math.MaxFloat64,
-		Max:            math.SmallestNonzeroFloat64,
+		CountPerBucket:     make([]int64, bucketCount),
+		ExemplarsPerBucket: make([]*exemplar.Exemplar, bucketCount),
+		bounds:             bounds,
+		Min:                math.MaxFloat64,
+		Max:                math.SmallestNonzeroFloat64,
 	}
 }
 
@@ -119,7 +129,8 @@ func (a *DistributionData) variance() float64 {
 
 func (a *DistributionData) isAggregationData() bool { return true }
 
-func (a *DistributionData) addSample(f float64) {
+func (a *DistributionData) addSample(e *exemplar.Exemplar) {
+	f := e.Value
 	if f < a.Min {
 		a.Min = f
 	}
@@ -127,7 +138,7 @@ func (a *DistributionData) addSample(f float64) {
 		a.Max = f
 	}
 	a.Count++
-	a.incrementBucketCount(f)
+	a.addToBucket(e)
 
 	if a.Count == 1 {
 		a.Mean = f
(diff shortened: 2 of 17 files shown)


ramonza:
@odeke-em still waiting on the spec to be merged (https://github.com/census-instrumentation/opencensus-specs/pull/179) but what do you think of the approach?

stats/doc.go
```diff
@@ -21,35 +21,49 @@ aggregate the collected data, and export the aggregated data.
...
 Measurement is a data point to be collected for a measure. For example,
 for a latency (ms) measure, 100 is a measurement that represents a 100ms
-latency event. Users collect data points on the existing measures with
+latency event. Measurements are created from measures with
 the current context. Tags from the current context are recorded with the
 measurements if they are any.
 
-Recorded measurements are dropped immediately if user is not aggregating
-them via views. Users don't necessarily need to conditionally enable/disable
+Recorded measurements are dropped immediately if no views are registered for then.
```
odeke-em: s/then/them/

but we could improve that sentence by saying
```go
Recorded measurements are dropped immediately if they don't have any associated views.
```

stats/view/aggregation_data.go
```diff
@@ -89,18 +92,35 @@ func (a *SumData) equal(other AggregationData) bool {
...
-	Mean            float64   // mean of the distribution
-	SumOfSquaredDev float64   // sum of the squared deviation from the mean
-	CountPerBucket  []int64   // number of occurrences per bucket
-	bounds          []float64 // histogram distribution of the values
+	Count           int64   // number of data points aggregated
+	Min             float64 // minimum value in the distribution
+	Max             float64 // max value in the distribution
+	Mean            float64 // mean of the distribution
+	SumOfSquaredDev float64 // sum of the squared deviation from the mean
+	// Deprecated: Replaced by Count in Buckets.
```
odeke-em: Perhaps let's put this field at the bottom and a blank space before it to draw attention to it like this:
```go
     bounds []float64 // histogram distribution of the values
     Buckets []Bucket
     
     // Deprecated: Replaced by Count in Buckets.
     CountPerBucket []int64 // Number of occurances per bucket
```

stats/view/aggregation_data.go
```diff
@@ -139,26 +159,61 @@ func (a *DistributionData) addSample(f float64) {
...
 
-	for i, b := range a.bounds {
-		if f < b {
-			a.CountPerBucket[i]++
-			return
+	// Heuristic to pick the "better" exemplar: first keep the one with a
+	// sampled trace attachment, if neither have a trace attachment, pick the
+	// one with more attachments.
+	_, haveTraceID := e.Attachments["TraceID"]
+	if haveTraceID || len(e.Attachments) >= len(b.Exemplar.Attachments) {
```
odeke-em: I believe we meant to use `>` instead of `>=` otherwise it'll always create a new Exempler if `!haveTraceID && len(e.Attachments) == len(b.Exemplar.Attachments)`

stats/view/aggregation_data.go
```diff
@@ -139,26 +159,61 @@ func (a *DistributionData) addSample(f float64) {
...
 }
 
 func (a *DistributionData) clone() AggregationData {
+	c := *a
 	counts := make([]int64, len(a.CountPerBucket))
 	copy(counts, a.CountPerBucket)
-	c := *a
 	c.CountPerBucket = counts
+	buckets := make([]Bucket, len(a.Buckets))
+	copy(buckets, a.Buckets)
```
odeke-em: This only changes the slices but the underlying ExemplarData per bucket is the same
since the ExemplarData pointer is unchanged so this is a shallow copy. Copying CountsPerBucket copies int64s which are just values.

trace/exemplar.go
```diff
@@ -0,0 +1,52 @@
...
+	ExemplarKeyTraceOptions = "TraceOptions"
+)
+
+func init() {
+	exemplar.RegisterAttachmentExtractor(attachSpanContext)
+}
+
+func attachSpanContext(ctx context.Context, attachments map[string]string) map[string]string {
+	span := FromContext(ctx)
+	sc := span.SpanContext()
```
odeke-em: This might result in a nil dereference on the first span(if SpanContext doesn't check that it is nil), but anyways we don't use it if span == nil. Please move it below after we've checked `if span == nil`

trace/exemplar_test.go
```diff
@@ -0,0 +1,114 @@
...
+		t.Fatalf("bucket.Count = %d; want %d", got, want)
+	}
+	exemplar := bucket.Exemplar
+	if exemplar == nil {
+		t.Fatal("Expected exemplar")
+	}
+	if got, want := exemplar.Value, 1.5; got != want {
+		t.Fatalf("exemplar.Value = %v; got %v", got, want)
+	}
+	if _, ok := exemplar.Attachments["TraceID"]; !ok {
```
odeke-em: Could we please use ExemplarTraceID and ExemplerSpanID, etc.. as the keys in these tests?

exemplar/extractor.go
```diff
@@ -0,0 +1,72 @@
...
+	"time"
+)
+
+// Exemplar represents an abstract set of supplemental information associated with
+// a measurement, for the purpose of providing an example of the type of
+// situation that resulted in the measurement.
+// For example, a trace exemplar contains key-value pairs representing a
+// SpanContext and points to an example span that resulted in a given
+// measurement.
+type Exemplar struct {
```
odeke-em: Am curious why we have two very similar structs `Exemplar` and `ExemplarData` which vary by only one field `Value`, but also this definition of Exemplar seems only suited to traces yet we expect them to work for both traces and metrics.

stats/view/aggregation_data.go
```diff
@@ -89,18 +92,35 @@ func (a *SumData) equal(other AggregationData) bool {
...
+	SumOfSquaredDev float64 // sum of the squared deviation from the mean
+	// Deprecated: Replaced by Count in Buckets.
+	CountPerBucket []int64   // number of occurrences per bucket
+	bounds         []float64 // histogram distribution of the values
+	Buckets        []Bucket
+}
+
+// ExemplarData is an example data point associated with each bucket.
+type ExemplarData struct {
+	Value       float64           // the value that was recorded
```
odeke-em: Nice! I was actually thinking to comment about making this a float64, but you already did it.

stats/view/aggregation_data.go
```diff
@@ -139,26 +159,61 @@ func (a *DistributionData) addSample(f float64) {
...
-		a.CountPerBucket[0]++
+func (a *DistributionData) addToBucket(val float64, ex *exemplar.Exemplar) {
+	var (
+		count  *int64
+		bucket *Bucket
+	)
+
+	for i, b := range a.bounds {
+		if val < b {
+			count = &a.CountPerBucket[i]
```
odeke-em: I now wonder why bounds, Bucket aren't of type `map[int]float64`.
I believe it would save on:
* Potential runtime crashes from out-of-bounds index access with slices
* Having to understand that the zeroth bucket is for the case when we don't have any bounds
* Save on the pointer manipulation
* Save on upfront memory allocations e.g. if we had a program using lots of views with DistributionData
* [index] access would automatically create zero values

To illustrate this point example this function would simply become
```go
func (a *DistributionData) addToBucket(val float64, ex *exemplar.Exemplar) {
        valBucketIndex := len(a.bounds)
        for i, b := range a.bounds {
                if val < b {
                        valBucketIndex = i 
                        break
                }
        }

        bucket := a.Buckets[valBucketIndex]
        bucket.Count += 1
        a.CountPerBucket[valBucketIndex] = bucket.Count
        bucket.maybeRetainExemplar(val, ex)
}
```

Although perhaps it is too late to change the types of Count* and Bucket* but it is something to keep in mind for future designs :)

stats/doc.go
ramonza: I think that the current phrasing is better since it is very specific in saying "registered" (which is the name of the function) while "associated" is more vague.

stats/view/aggregation_data.go
ramonza: Yeah, the idea is that we replace the existing exemplar with the newer one if they are equally "good" under the assumption that newer is better.

stats/view/aggregation_data.go
ramonza: Yeah but we don't expect anyone to mutate the map at this stage (it's only mutated when the exemplary is first created) so I think shallow copying in this case is fine.

trace/exemplar_test.go
ramonza: These are the keys from the spec. I think since they map is already inside the exemplar, repeating the work "exemplar" doesn't really add anything.

exemplar/extractor.go
ramonza: Done

stats/view/aggregation_data.go
ramonza: I created a type for attachments and added this as a note on the type godoc.

exemplar/extractor.go
```diff
@@ -0,0 +1,77 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package exemplar implements support for exemplars. Exemplars are additional
+// data associated with each measurement.
+//
+// Their purpose it to provide an example of the kind of thing
+// (request, RPC, trace span, etc.) that resulted in that measurement.
+package exemplar
```
rakyll: Is this a public API or should it be internal for now?

What are the public use cases?

ramonza: The Exemplar type is needed for exporters and the AttachmentExtractor is meant to be an extensible mechanism (see comments on the spec PR about ways to extract attachments from Tracestate).

exemplar/extractor.go
```diff
@@ -0,0 +1,77 @@
...
+	extractors = append(extractors, e)
+}
+
+// NewFromContext extracts exemplars from the given context.
+// Each registered AttachmentExtractor (see RegisterAttachmentExtractor) is called in an
+// unspecified order to add attachments to the exemplar.
+func AttachmentsFromContext(ctx context.Context) Attachments {
+	var a Attachments
+	for _, extractor := range extractors {
+		a = extractor(ctx, a)
```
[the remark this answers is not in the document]
ramonza: We may have more than one extractor return valid attachments. For example, if there are tags in the context as well as a sampled trace.

exemplar/extractor.go
```diff
@@ -0,0 +1,77 @@
...
+	"time"
+)
+
+const (
+	KeyTraceID   = "trace_id"
+	KeySpanID    = "span_id"
+	KeyPrefixTag = "tag:"
+)
+
+// Exemplar is an example data point associated with each bucket.
```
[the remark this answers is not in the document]
ramonza: Yes, it is only available for distribution aggregation. I will make that more explicit.

exemplar/extractor.go
```diff
@@ -0,0 +1,77 @@
...
+type Exemplar struct {
+	Value       float64     // the value that was recorded
```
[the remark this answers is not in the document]
ramonza: This is how it is in the metrics proto: https://github.com/census-instrumentation/opencensus-proto/blob/a74b002d33b4471aff7288afdb41393f6ff02ca1/src/opencensus/proto/metrics/v1/metrics.proto

stats/view/aggregation_data.go
```diff
@@ -17,14 +17,16 @@ package view
...
+	"go.opencensus.io/exemplar"
 )
 
 // AggregationData represents an aggregated value from a collection.
 // They are reported on the view data during exporting.
 // Mosts users won't directly access aggregration data.
 type AggregationData interface {
 	isAggregationData() bool
-	addSample(v float64)
+	addSample(e *exemplar.Exemplar)
```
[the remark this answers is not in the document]
ramonza: I agree this is a bit odd. The reason for this diff (removing the v argument) is that it is repeated inside the exemplar.

Regarding the presence of exemplars in the core APIs (it being a top-level package), I am open to moving it under "stats" or even "stats/view". But then we would have a direct dependency from the trace package to a package under "stats".

stats/view/aggregation_data.go
```diff
@@ -89,18 +91,30 @@ func (a *SumData) equal(other AggregationData) bool {
...
+	bounds []float64 // histogram distribution of the values
+
+	// Deprecated: Replaced by Count in Buckets.
+	CountPerBucket []int64 // number of occurrences per bucket
+}
+
+// Bucket represents a single bucket value in a distribution.
+type Bucket struct {
+	Count    int64              // Number of recorded measurements falling into this bucket.
+	Exemplar *exemplar.Exemplar // Exemplar of this bucket.
```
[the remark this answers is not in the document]
ramonza: Yes, see Bucket in: https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto

stats/view/aggregation_data.go
ramonza: ok, I think this makes sense especially given that most Exporters won't care about exemplars.

ramonza:
@rakyll updated, PTAL

stats/view/worker_commands.go
```diff
@@ -151,7 +155,12 @@ func (cmd *recordReq) handleCommand(w *worker) {
 		}
 		ref := w.getMeasureRef(m.Measure().Name())
 		for v := range ref.views {
-			v.addSample(cmd.tm, m.Value())
+			e := &exemplar.Exemplar{
```
songy23: Creating a new `exemplar` for every `measurement` is very wrong. Going to revert this change.

PR #918, opened 2018-09-18
Add stats.RecordWithTags

### LINKED ISSUE #834: Provide a method to record with explicit tags
See: https://github.com/census-instrumentation/opencensus-specs/issues/140

> Contexts are currently the only way to provide tag values (AFAICT), which makes sense for the majority of metrics, which are scoped to a some kind of request and certainly for traces.

> For metrics there are many cases though, where one just wants to measure some state of application internals, e.g. the length of a queue.
It would be helpful if one could initialize a static set of tags somehow (e.g. queue identifier) without forcing a context object into all APIs for the sole purpose of setting a metric tag. It may actually force developers to significantly refactor and break their APIs if they want to add OC instrumentation to existing code.

> I think similar issues would arise in practice for an OC logging API.

rakyll:
Fixes #834.

### DIFF
--- a/plugin/ocgrpc/stats_common.go
+++ b/plugin/ocgrpc/stats_common.go
@@ -142,18 +142,21 @@ func handleRPCEnd(ctx context.Context, s *stats.End) {
 
 	latencyMillis := float64(elapsedTime) / float64(time.Millisecond)
 	if s.Client {
-		ctx, _ = tag.New(ctx,
-			tag.Upsert(KeyClientMethod, methodName(d.method)),
-			tag.Upsert(KeyClientStatus, st))
-		ocstats.Record(ctx,
+		ocstats.RecordWithTags(ctx,
+			[]tag.Mutator{
+				tag.Upsert(KeyClientMethod, methodName(d.method)),
+				tag.Upsert(KeyClientStatus, st),
+			},
 			ClientSentBytesPerRPC.M(atomic.LoadInt64(&d.sentBytes)),
 			ClientSentMessagesPerRPC.M(atomic.LoadInt64(&d.sentCount)),
 			ClientReceivedMessagesPerRPC.M(atomic.LoadInt64(&d.recvCount)),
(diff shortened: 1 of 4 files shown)


plugin/ocgrpc/stats_common.go
```diff
@@ -142,18 +142,21 @@ func handleRPCEnd(ctx context.Context, s *stats.End) {
 
 	latencyMillis := float64(elapsedTime) / float64(time.Millisecond)
 	if s.Client {
-		ctx, _ = tag.New(ctx,
-			tag.Upsert(KeyClientMethod, methodName(d.method)),
-			tag.Upsert(KeyClientStatus, st))
-		ocstats.Record(ctx,
+		ocstats.RecordWithTags(ctx,
```
rakyll: Not sure if the change helps these cases. RecordWithTags is going to return without recording if there is an error.

stats/record.go
```diff
@@ -52,3 +52,17 @@ func Record(ctx context.Context, ms ...Measurement) {
 	}
 	recorder(tag.FromContext(ctx), ms)
 }
+
+// RecordWithTags records one or multiple measurements at once.
+//
+// Measurements will be tagged with the tags in the context mutated by the mutators.
+// RecordWithTags is useful if you want to record with tag mutations but don't want
+// to propagate the mutations in the context.
+func RecordWithTags(ctx context.Context, mutators []tag.Mutator, ms ...Measurement) error {
```
ramonza: I don't really mind but I think in this case it isn't worth having varargs for the `ms` argument. I think since we are requiring the user to explicitly pass a `tag.Mutator` slice, we might as well also require an explicit `ms` slice. 

The small advantage is that we could in future introduce a new varargs argument if we ever found the need for that in a backwards-compatible way. 

This is just a thought, I don't really mind if we leave it like this either.

rakyll: I like it as it is because the APIs are more symmetric.

PR #920, opened 2018-09-19
[travis] Run test suite against 386 arch

### LINKED ISSUE #868: tests: provide mode to run on various architectures
This feature request is a suggestion that might be very helpful: this repository is critical to many infrastructure projects and we are getting to the point where it would be beneficial to have code tested on various architectures to avoid surprises such as:
* https://github.com/census-instrumentation/opencensus-go/issues/587
* https://github.com/census-instrumentation/opencensus-go/issues/865
which were reported after users' programs crashed. Those issues manifest from bugs in Go themselves but it would be nice to catch this during development beforehand and not from users

Go allows us to test on various architectures out of the box by using `GOARCH` for say
* amd64
* 386
* arm
* arm64
etc

For example both #587 and #865 could have trivially been caught during development and by CI, by running:
```shell
GOARCH=arm go test ./...
```
and 
```shell
GOARCH=386 go test ./...
```
respectively, but also perhaps on the various architectures if we can get access to real hardware.

diogenes:
Notice: Build time for this change is increasing: `Ran for 4 min 21 sec`. Previously it was about 3min.

Besides that, we have a better coverage for problems targeting a 386 architecture.

[Closes #868]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,5 +22,6 @@ script:
   - if [ -n "$(gofmt -s -l $GO_FILES)" ]; then echo "gofmt the following files:"; gofmt -s -l $GO_FILES; exit 1; fi
   - go vet ./...
   - go test -v -race $PKGS            # Run all the tests with the race detector enabled
+  - GOARCH=386 go test -v $PKGS       # Run all tests against a 386 architecture
   - 'if [[ $TRAVIS_GO_VERSION = 1.8* ]]; then ! golint ./... | grep -vE "(_mock|_string|\.pb)\.go:"; fi'
   - go run internal/check/version.go


.travis.yml
```diff
@@ -22,5 +22,6 @@ script:
   - if [ -n "$(gofmt -s -l $GO_FILES)" ]; then echo "gofmt the following files:"; gofmt -s -l $GO_FILES; exit 1; fi
   - go vet ./...
   - go test -v -race $PKGS            # Run all the tests with the race detector enabled
+  - GOOS=linux GOARCH=386 go test -v $PKGS # Run all tests against a 386 architecture
```
ramonza: is explicitly setting GOOS necessary?

diogenes: In fact, it seems it's not needed! Thanks for the heads up, @ramonza!

PR #922, opened 2018-09-21
stats/view: implement Flush

### LINKED ISSUE #862: Make it possible for opencensus to pass all remaining data to registered exporters when program exits.
Problem:

When exiting a program running opencensus monitoring, some of the data are not passed to exporter. This happens because as one can see in following code, reportUsage() is called only by timer, so when program exists, all data passed to opencensus after last timer firing event are simply lost.

```
func (w *worker) start() {
	for {
		select {
		case cmd := <-w.c:
			cmd.handleCommand(w)
		case <-w.timer.C:
			w.reportUsage(time.Now())
		case <-w.quit:
			w.timer.Stop()
			close(w.c)
			w.done <- true
			return
		}
	}
}
```

Solution proposal:
1. Make `func (w *worker) stop()` to be (indirectly) callable by client of package.
2. Make `func (w *worker) start()` to call `w.reportUsage()` before returning when `<-w.quit` is triggered.

odeke-em:
Flush is a global function that immediately reports
all collected data by the worker, regardless of the
reporting duration or buffering.

Added tests but also ensured that calling (*worker).stop()
then Flush doesn't block and returns ASAP if the state
is "quit". The required protected w.quit from double close
but also using a select to detect cases of forever waiting
channel `w.quit` which was closed.

Fixes #862

### DIFF
--- a/stats/view/view_measure_test.go
+++ b/stats/view/view_measure_test.go
@@ -8,6 +8,8 @@ import (
 )
 
 func TestMeasureFloat64AndInt64(t *testing.T) {
+	restart()
+
 	// Recording through both a Float64Measure and Int64Measure with the
 	// same name should work.
 
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -17,6 +17,7 @@ package view
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	"go.opencensus.io/stats"
@@ -43,6 +44,8 @@ type worker struct {
 	timer      *time.Ticker
 	c          chan command
 	quit, done chan bool
+	flushCh    chan bool
+	quitOnce   sync.Once
 }
 
 var defaultWorker *worker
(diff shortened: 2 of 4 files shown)


odeke-em:
The Travis CI test seems unrelated to this change, locally running
```shell
$ go test -run=TestEndToEnd_Single -count=100
PASS
ok  	go.opencensus.io/plugin/ocgrpc	0.153s
```
passes

ramonza:
@odeke-em if that is the case we should skip the test and file an issue

ramonza:
I think we need specs agreement before we can proceed here: https://github.com/census-instrumentation/opencensus-specs/issues/152

rakyll:
Ah, I implemented before seeing this PR. Feel free to void it. https://github.com/census-instrumentation/opencensus-go/pull/923

bogdandrutu:
@odeke-em I think (RetrieveData)[https://github.com/census-instrumentation/opencensus-go/blob/master/stats/view/worker.go#L94] is your answer for this. See how I used it https://github.com/census-instrumentation/opencensus-service/blob/master/observability/observabilitytest/observabilitytest.go#L105

PR #923, opened 2018-09-21
Add Flush to force aggregate in-memory buffer

### LINKED ISSUE #862: Make it possible for opencensus to pass all remaining data to registered exporters when program exits.

rakyll:
Fixes #862.

### DIFF
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -40,9 +40,9 @@ type worker struct {
 	views      map[string]*viewInternal
 	startTimes map[*viewInternal]time.Time
 
-	timer      *time.Ticker
-	c          chan command
-	quit, done chan bool
+	timer                   *time.Ticker
+	c                       chan command
+	quitCh, doneCh, flushCh chan struct{}
 }
 
 var defaultWorker *worker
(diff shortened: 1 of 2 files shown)


odeke-em:
@rakyll thanks for also working on this! I had mailed #922 sometime back. However, be ware that calling Flush after any future stops shouldn't block :)

bogdandrutu:
@rakyll I think (RetrieveData)[https://github.com/census-instrumentation/opencensus-go/blob/master/stats/view/worker.go#L94] is your answer for this. See how I used it https://github.com/census-instrumentation/opencensus-service/blob/master/observability/observabilitytest/observabilitytest.go#L105

Do you think that is enough?

wdhnl:
Can this be looked at again?

rakyll:
Closing this due to inactivity, please open another PR if it needs to be addressed.

PR #924, opened 2018-09-21
Allow user to set start options per HTTP request

### LINKED ISSUE #869: http.Request influenced sampling

rakyll:
Fixes #869.

### DIFF
--- a/plugin/ochttp/client.go
+++ b/plugin/ochttp/client.go
@@ -47,6 +47,10 @@ type Transport struct {
 	// for spans started by this transport.
 	StartOptions trace.StartOptions
 
+	// GetStartOptions allows to set start options per request. If set,
+	// StartOptions is going to be ignored.
+	GetStartOptions func(*http.Request) trace.StartOptions
+
 	// NameFromRequest holds the function to use for generating the span name
 	// from the information found in the outgoing HTTP Request. By default the
 	// name equals the URL Path.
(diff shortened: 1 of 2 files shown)


rakyll:
Also created https://github.com/census-instrumentation/opencensus-specs/pull/182.

plugin/ochttp/client.go
```diff
@@ -47,6 +47,10 @@ type Transport struct {
 	// for spans started by this transport.
 	StartOptions trace.StartOptions
```
ramonza: Should we deprecate this?

rakyll: If we deprecate this GetStartOptions will be a bad choice of naming. GetX only makes sense when X already exists.

PR #928, opened 2018-09-23
http out test cases 

SergeyKanzhelev:
This is example how shared http out test cases can be used across repositories. The JSON can be either downloaded directly from specs repo or copied into this repo periodically with some script.

This PR also makes two changes in actual code:

1. Do not add `http.user_agent` attribute when it's empty
2. Set `error` attribute to spans in case of error like DNS error according to https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table


plugin/ochttp/http-out-test-cases.json
```diff
@@ -0,0 +1,258 @@
+[
```
ramonza: Usually this type of file is in a `testdata` folder: https://dave.cheney.net/2016/05/10/test-fixtures-in-go

plugin/ochttp/trace.go
```diff
@@ -83,6 +83,7 @@ func (t *traceTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	resp, err := t.base.RoundTrip(req)
 	if err != nil {
 		span.SetStatus(trace.Status{Code: trace.StatusCodeUnknown, Message: err.Error()})
+		span.AddAttributes(trace.BoolAttribute("error", true))
```
ramonza: I'm not sure what the point of this is? Is this required by the specs?

plugin/ochttp/trace.go
```diff
@@ -149,11 +150,21 @@ func spanNameFromURL(req *http.Request) string {
...
-	return []trace.Attribute{
-		trace.StringAttribute(PathAttribute, r.URL.Path),
-		trace.StringAttribute(HostAttribute, r.URL.Host),
-		trace.StringAttribute(MethodAttribute, r.Method),
-		trace.StringAttribute(UserAgentAttribute, r.UserAgent()),
+	userAgent := r.UserAgent()
+
+	if userAgent != "" {
+		return []trace.Attribute{
+			trace.StringAttribute(PathAttribute, r.URL.Path),
```
ramonza: Why are we doing this?

plugin/ochttp/http-out-test-cases.json
SergeyKanzhelev: What's your opinion here - is it better to take a copy from specs or download directly from there? Second approach will fail all incompliant SDKs compilation on any spec modification. But fixes will be implemented faster.

plugin/ochttp/trace.go
SergeyKanzhelev: This is the way to indicate that this unknown is clearly an error. Attribute is taken from open tracing spec. Any suggestions here? Should there be UNKNOWN_ERROR status and UNKNOWN_UNKNOWN?

plugin/ochttp/trace.go
SergeyKanzhelev: just to omit empty `http.user_agent` attribute. If we have a spec we need to agree on single implementation. Python never collects it, Go - always. In C# I want to omit including empty-values attribute to save space. I'll add a note to spec in specs PR.

Do you disagree with intent or just curious why?

plugin/ochttp/http-out-test-cases.json
ramonza: Unit tests should never make network calls (except localhost). Better to put it in the repo like you have it here.

plugin/ochttp/trace.go
SergeyKanzhelev: it's from open tracing specs, not open census =). I've added a mention of it in my specs PR

SergeyKanzhelev: the problem is that `UNKNOWN` is ambiguous. Is it good unknown or bad? When we know for sure - it will be good to indicate it

ramonza: I don't think it is ambiguous. The error message will clarify what went wrong. We have spoken about having an OpenTracing compatibility layer and hopefully we will get that working eventually and then that would be an appropriate place to put this I think. For now, I would prefer not to add it or if you really feel strongly about it, please add it to OpenCensus specs and get approval there and then we can implement it.

plugin/ochttp/trace.go
ramonza: I don't disagree with the intent but I prefer not to change things unless there is a reason: it's always possible someone was relying on the existing behavior in a way we didn't anticipate and we create some problems for them.

An additional concern was the code duplication that was introduced here, which we should avoid regardless of the outcome.

I am happy to make this change if it brings Go into better alignment with the other implementations, since it is pretty small. In that case please refactor to remove the duplicate code if possible.

plugin/ochttp/http-out-test-cases.json
SergeyKanzhelev: I made the change

plugin/ochttp/trace.go
SergeyKanzhelev: ok. 

> I would prefer not to add it or if you really feel strongly about it, please add it to OpenCensus specs and get approval there and then we can implement it.

exactly what I'm doing =) this is a POC for test cases proposed there. 

I think I'll pull this `error` thing out for now (maybe later tonight) to separate two discussions

plugin/ochttp/trace.go
SergeyKanzhelev: In `Go` - will it be better to return a slice of fixed length array to implement this kind of variability?

SergeyKanzhelev: Right now as I said it's not aligned at all. So this is an attempt to do it. I'd like to make this change instead of introducing empty `user_agent` property in other languages. I figured impact here should be small.

I definitely share the desire to minimize change

ramonza: You can do something like:
```go
attrs := make([]trace.Attribute, 0, 4)
attrs = append(attrs, 
  trace.StringAttribute(HostAttribute, r.URL.Host),
  trace.StringAttribute(MethodAttribute, r.Method),
  ...)

if ua != "" {
attrs = append(attrs, trace.StringAttribute(UserAgentAttribute, ua))
}
```
I think that will avoid an extra allocation in the common case where there is a user agent.

plugin/ochttp/trace.go
SergeyKanzhelev: error attribute was removed from the spec

SergeyKanzhelev:
@ramonza would you mind to look at it, I've refreshed it

plugin/ochttp/trace_test.go
```diff
@@ -517,6 +522,164 @@ func TestResponseAttributes(t *testing.T) {
...
+				if sp.SpanKind == trace.SpanKindClient {
+					client = sp
+				}
+			}
+
+			if client.Name != tt.SpanName {
+				t.Errorf("span names don't match: expected: %s, actual: %s", tt.SpanName, client.Name)
+			}
+
+			codeToStr := map[int32]string{
```
rghetia: how about using codeToStr from [here](https://github.com/census-instrumentation/opencensus-go/blob/ebb7978abd3f971d98f7b2b2be19e7e374409b0b/plugin/ochttp/trace.go#L199-L217)

PR #930, opened 2018-09-24
Add constructor functions for Handler and Transport

ramonza:
These set the default propagation format to TraceContext.

### DIFF
--- a/plugin/ochttp/client.go
+++ b/plugin/ochttp/client.go
@@ -25,9 +25,7 @@ import (
 // Transport is an http.RoundTripper that instruments all outgoing requests with
 // OpenCensus stats and tracing.
 //
-// The zero value is intended to be a useful default, but for
-// now it's recommended that you explicitly set Propagation, since the default
-// for this may change.
+// Use NewTransport to create a Transport with the default configuration.
 type Transport struct {
 	// Base may be set to wrap another http.RoundTripper that does the actual
 	// requests. By default http.DefaultTransport is used.
@@ -36,8 +34,9 @@ type Transport struct {
 	// the returned round tripper will be cancelable.
 	Base http.RoundTripper
 
-	// Propagation defines how traces are propagated. If unspecified, a default
-	// (currently B3 format) will be used.
+	// Propagation defines how traces are propagated. NewTransport will configure
+	// W3C TraceContext as the default format. For backwards compatibility, if
+	// this field is set to nil, B3 propagation format will be used.
 	Propagation propagation.HTTPFormat
 
 	// StartOptions are applied to the span started by this Transport around each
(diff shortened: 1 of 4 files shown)


plugin/ochttp/client.go
```diff
@@ -69,7 +77,7 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	// TODO: remove excessive nesting of http.RoundTrippers here.
 	format := t.Propagation
 	if format == nil {
-		format = defaultFormat
+		format = compatibilityFormat
```
rakyll: Why this change?

plugin/ochttp/client.go
```diff
@@ -25,9 +25,7 @@ import (
 // Transport is an http.RoundTripper that instruments all outgoing requests with
 // OpenCensus stats and tracing.
 //
-// The zero value is intended to be a useful default, but for
-// now it's recommended that you explicitly set Propagation, since the default
-// for this may change.
+// Use NewTransport to create a Transport with the default configuration.
```
rakyll: I think this is a confusing choice. Zero-values are expected to support the default configuration.

We should deprecate the use of Transport zero value if this is the way to go.

plugin/ochttp/server.go
```diff
@@ -68,6 +72,17 @@ type Handler struct {
...
 }
 
+// NewHandler wraps the given handler to add stats and tracing instrumentation.
+// By default, W3C TraceContext propagation format is used and the returned
+// handler is assumed to be a public endpoint.
+func NewHandler(next http.Handler) *Handler {
+	return &Handler{
+		Handler:          next,
+		Propagation:      defaultFormat,
+		IsPublicEndpoint: true,
```
rakyll: It is less likely that a server is a a public endpoint.

plugin/ochttp/client.go
ramonza: I just renamed the variable `defaultFormat` to `compatibilityFormat` to make it clear that we intend the new default to be TraceContext.

plugin/ochttp/client.go
ramonza: I would like to deprecate use of the zero value. Just not exactly sure how to go about this, except for what I have done here which is to document that NewTransport should be used.

ramonza:
Waiting for: https://github.com/census-instrumentation/opencensus-specs/pull/185

plugin/ochttp/server.go
ramonza: I would guess that it's really a lot *more* likely to be public for most users, especially the smaller users that are not running complex microservice architectures. It is also a safer default.

ramonza:
Closing because I believe we decided to make a backwards incompatible change to TraceContext instead.

PR #931, opened 2018-09-24
improve W3C Trace Context compliance

reyang:
Trying to improve the compliance by fixing:
1. Current version allows trace options to be optional, which is against https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#traceparent-field.
2. Current implementation doesn't combine multiple headers with the same name.
3. Version compatibility.

### DIFF
--- a/plugin/ochttp/propagation/tracecontext/propagation.go
+++ b/plugin/ochttp/propagation/tracecontext/propagation.go
@@ -51,43 +51,49 @@ func (f *HTTPFormat) SpanContextFromRequest(req *http.Request) (sc trace.SpanCon
 		return trace.SpanContext{}, false
 	}
 	sections := strings.Split(h, "-")
-	if len(sections) < 3 {
+	if len(sections) < 4 {
 		return trace.SpanContext{}, false
 	}
 
+	if len(sections[0]) != 2 {
+		return trace.SpanContext{}, false
+	}
 	ver, err := hex.DecodeString(sections[0])
 	if err != nil {
 		return trace.SpanContext{}, false
 	}
-	if len(ver) == 0 || int(ver[0]) > supportedVersion || int(ver[0]) > maxVersion {
+	version := int(ver[0])
+	if version > maxVersion {
 		return trace.SpanContext{}, false
 	}
 
+	if version == 0 && len(sections) != 4 {
+		return trace.SpanContext{}, false
+	}
+
+	if len(sections[1]) != 32 {
+		return trace.SpanContext{}, false
+	}
 	tid, err := hex.DecodeString(sections[1])
 	if err != nil {
 		return trace.SpanContext{}, false
 	}
-	if len(tid) != 16 {
-		return trace.SpanContext{}, false
-	}
 	copy(sc.TraceID[:], tid)
 
+	if len(sections[2]) != 16 {
+		return trace.SpanContext{}, false
+	}
 	sid, err := hex.DecodeString(sections[2])
 	if err != nil {
 		return trace.SpanContext{}, false
 	}
-	if len(sid) != 8 {
-		return trace.SpanContext{}, false
-	}
 	copy(sc.SpanID[:], sid)
 
-	if len(sections) == 4 {
-		opts, err := hex.DecodeString(sections[3])
-		if err != nil || len(opts) < 1 {
-			return trace.SpanContext{}, false
-		}
-		sc.TraceOptions = trace.TraceOptions(opts[0])
+	opts, err := hex.DecodeString(sections[3])
+	if err != nil || len(opts) < 1 {
+		return trace.SpanContext{}, false
 	}
+	sc.TraceOptions = trace.TraceOptions(opts[0])
 
 	// Don't allow all zero trace or span ID.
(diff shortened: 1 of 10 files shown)


plugin/ochttp/propagation/tracecontext/propagation.go
```diff
@@ -98,13 +104,26 @@ func (f *HTTPFormat) SpanContextFromRequest(req *http.Request) (sc trace.SpanCon
 	return sc, true
 }
 
+func getRequestHeader(req *http.Request, name string, commaSeparated bool) (hdr string, ok bool) {
+	ok = false
```
ramonza: It's not really clear what "ok" means here. I would expect it to mean that the header is present, but that doesn't seem to be the case since it can be false even when the header is present. Perhaps it should have a different name, or be documented.

rakyll: Remove, zero value of a boolean is false already.

plugin/ochttp/propagation/tracecontext/propagation.go
```diff
@@ -98,13 +104,26 @@ func (f *HTTPFormat) SpanContextFromRequest(req *http.Request) (sc trace.SpanCon
...
+	for k, v := range req.Header {
+		if strings.EqualFold(k, name) {
```
ramonza: Instead of looping through all the headers, you can canonicalize `name` using `textproto.CanonicalMIMEHeaderKey` and index `req.Header` directly.

plugin/ochttp/propagation/tracecontext/propagation.go
```diff
@@ -98,13 +104,26 @@ func (f *HTTPFormat) SpanContextFromRequest(req *http.Request) (sc trace.SpanCon
...
+			if commaSeparated || len(v) == 1 {
+				ok = true
+			}
+			return strings.Join(v, ","), ok
```
rakyll: Can you give an example when this case happens?

A header key might have more than one value if user in the process mutates the header. Do we care about these cases?

reyang: https://tools.ietf.org/html/rfc7230#section-3.2.2

This is to make sure the ochttp tracecontext propagation could handle duplicated header names. It works for tracestate with either single key-value pair or multiple pairs.

```
traceparent: 00-12345678901234567890123456789012-1234567890123456-00
tracestate: rojo=1,congo=2
tracestate: foo=3,bar=4
tracestate: baz=5
```

After string.Join, the tracestate string would be rojo=1,congo=2,foo=3,bar=4,baz=5

plugin/ochttp/propagation/tracecontext/propagation.go
reyang: Got it, thanks.

plugin/ochttp/propagation/tracecontext/propagation.go
reyang: Is this conventional (which might break later) or something we could rely on?

ramonza: It is documented here: https://golang.org/pkg/net/http/?#Header.Get

reyang: That was what I was looking at, and it's not working.

According to the doc:

Get gets the **first** value associated with the given key.
...
To **access multiple values of a key**, or to use non-canonical keys, **access the map directly**.

plugin/ochttp/propagation/tracecontext/propagation.go
```diff
@@ -98,13 +104,26 @@ func (f *HTTPFormat) SpanContextFromRequest(req *http.Request) (sc trace.SpanCon
...
+}
+
 // TODO(rghetia): return an empty Tracestate when parsing tracestate header encounters an error.
 // Revisit to return additional boolean value to indicate parsing error when following issues
 // are resolved.
 // https://github.com/w3c/distributed-tracing/issues/172
 // https://github.com/w3c/distributed-tracing/issues/175
 func tracestateFromRequest(req *http.Request) *tracestate.Tracestate {
-	h := req.Header.Get(tracestateHeader)
+	h, _ := getRequestHeader(req, tracestateHeader, true)
```
ramonza: It's kind of weird that we join headers with commas only to then split by comma again below (line 131). Seems like we could avoid this unnecessary back and forward conversion by keeping the headers separate and splitting each one separately and then concatenating all the results after splitting.

plugin/ochttp/propagation/tracecontext/propagation.go
reyang: Thanks for the explanation, makes sense.

reyang:
Messed up my current working branch. I've created a new PR #934 with the review comments addressed.
Sorry for the inconvenience.

plugin/ochttp/propagation/tracecontext/propagation.go
reyang: I intended to keep the HTTP layer and TraceContext separate.
Combining duplicated headers using COMMA is mentioned by the HTTP RFC.
Splitting the COMMA separated key-value pair is required by W3C TraceContext.

Issue #935, opened 2018-09-26
exporter/prometheus: Avoid missing metrics


negz:
**Describe the bug**
As best I can tell the Go Prometheus metrics exporter does not export views as lines in the Prometheus exposition format until something actually happens to the view's measurement. Put otherwise, a view with a count aggregation metric will never expose an unlabelled counter at 0; each counter will spring into existence when it is first incremented. This goes against [Prometheus's recommendation](https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics) that instrumentation 'avoid missing metrics'. (I debated raising this as a feature request, given it _is_ only a guideline.)

**To Reproduce**
1. Download [the Prometheus exporter example code](https://raw.githubusercontent.com/census-instrumentation/opencensus-go/140879bf49cb17276aaf28b5f6a04e251ff98473/exporter/prometheus/example/main.go).
2. Apply the following diff:
```diff
--- ogmain.go   2018-09-26 01:37:35.000000000 -0700
+++ main.go     2018-09-26 01:37:10.000000000 -0700
@@ -17,9 +17,7 @@
 package main
 
 import (
-       "context"
        "log"
-       "math/rand"
        "net/http"
        "time"
 
@@ -36,8 +34,6 @@
 )
 
 func main() {
-       ctx := context.Background()
-
        exporter, err := prometheus.NewExporter(prometheus.Options{})
        if err != nil {
                log.Fatal(err)
@@ -68,13 +64,6 @@
        // Set reporting period to report data at every second.
        view.SetReportingPeriod(1 * time.Second)
 
-       // Record some data points...
-       go func() {
-               for {
-                       stats.Record(ctx, videoCount.M(1), videoSize.M(rand.Int63()))
-                       <-time.After(time.Millisecond * time.Duration(1+rand.Intn(400)))
-               }
-       }()
 
        addr := ":9999"
        log.Printf("Serving at %s", addr)
```
3. `go run main.go`
4. Attempt to curl the metrics:
```bash
$ curl -v http://localhost:9999/metrics
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 9999 (#0)
> GET /metrics HTTP/1.1
> Host: localhost:9999
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Length: 0
< Content-Type: text/plain; version=0.0.4; charset=utf-8
[...]

PR #937, opened 2018-09-27
wip: prometheus metrics should be given a default zero value

cstyan:
Let me know if this is completely wrong.

Based on @rakyll 's comment in #935 I tried removing the return if `len(vd.Rows) == 0`, but it seems we also need to then create and send a metric with a 0 value in `Collect`. Should we be checking the aggregation type of `vd`?

WIP, thought this would be a good first issue :)

### DIFF
--- a/exporter/prometheus/prometheus.go
+++ b/exporter/prometheus/prometheus.go
@@ -124,9 +124,9 @@ func (o *Options) onError(err error) {
 // to Untyped Metric, CountData will be a Counter Metric,
 // DistributionData will be a Histogram Metric.
 func (e *Exporter) ExportView(vd *view.Data) {
-	if len(vd.Rows) == 0 {
-		return
-	}
+	// if len(vd.Rows) == 0 {
+	// 	return
+	// }
 	e.c.addViewData(vd)
 }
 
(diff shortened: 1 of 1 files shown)


exporter/prometheus/prometheus.go
```diff
@@ -185,20 +185,27 @@ func (c *collector) Describe(ch chan<- *prometheus.Desc) {
...
 	for _, vd := range viewData {
 		sig := viewSignature(c.opts.Namespace, vd.View)
 		c.registeredViewsMu.Lock()
 		desc := c.registeredViews[sig]
 		c.registeredViewsMu.Unlock()
-
-		for _, row := range vd.Rows {
-			metric, err := c.toMetric(desc, vd.View, row)
+		if len(vd.Rows) == 0 {
+			metric, err := c.toMetric(desc, vd.View, &view.Row{Data: &view.CountData{}})
```
rakyll: How does this work when you switch to a different metric descriptor as soon as there is available data?

I don't know the canonical best approach here. OpenCensus is a dynamic instrumentation library which means you can introduce new aggregations in runtime. As far as I understand from your issue, Prometheus expects metric descriptors to be available even though there is no data collected on them. Maybe @fabxc can help us with https://github.com/census-instrumentation/opencensus-go/issues/935.

cstyan: The way I understand it the metric descriptor should not/does not change? Prometheus client libraries export default 0 values for metrics without labels. So I think the issue is that the metric is not sent during `Collect` if it doesn't have and data rows yet.

The Prometheus metrics should only be registered with the view once as far as I can tell.

fabxc: The point of the initialization recommendation is that all valid label combinations exist already. This will just expose a series with the same name but no labels combinations set at all. This is more harmful than not exposing any series since you know have a metric that changes its signature at runtime once data is observed.

See @negz second comment on the issue: https://github.com/census-instrumentation/opencensus-go/issues/935#issuecomment-424643888

fabxc:
Please see my comment on the main issue: https://github.com/census-instrumentation/opencensus-go/issues/935#issuecomment-425355874

ramonza:
Seems from the comments on https://github.com/census-instrumentation/opencensus-go/issues/935 like we are in agreement that this is not the way to go.

PR #938, opened 2018-10-03
Don't rely on global trace.ApplyConfig in tests

ramonza:
Some tests are relying on previous tests changing the global sampler
leading to random failures when tests are run in isolation.

Set an explicit sampler rather than relying on globals.


songy23:
Please fix the build.

PR #941, opened 2018-10-04
Provide ways to avoid propagating trace and tag metadata

ramonza:
In certain situations it is undesirable to propagate trace or tag
metadata on outbound requests. Nevertheless, it may still be
desirable to instrument these requests using ocgrpc or ochttp so
that stats and in-process spans may be collected.

This change provides ways to accomplish this for both ocgrpc and
ochttp.

### DIFF
--- a/plugin/ocgrpc/client_stats_handler.go
+++ b/plugin/ocgrpc/client_stats_handler.go
@@ -25,7 +25,7 @@ import (
 )
 
 // statsTagRPC gets the tag.Map populated by the application code, serializes
-// its tags into the GRPC metadata in order to be sent to the server.
+// its tags into the gRPC metadata in order to be sent to the server.
 func (h *ClientHandler) statsTagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
 	startTime := time.Now()
 	if info == nil {
--- a/plugin/ocgrpc/grpc_test.go
+++ b/plugin/ocgrpc/grpc_test.go
@@ -68,9 +68,9 @@ func TestClientHandler(t *testing.T) {
 
 func TestServerHandler(t *testing.T) {
 	tests := []struct {
-		name         string
-		newTrace     bool
-		expectTraces int
+		name           string
+		publicEndpoint bool
+		expectTraces   int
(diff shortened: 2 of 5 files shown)


rakyll:
In-process instrumentation still may need dimensions to be propagated properly.

ramonza:
@rakyll I'm not sure I understand your comment. This PR does not prevent in-process propagation.

plugin/ocgrpc/client.go
```diff
@@ -29,6 +29,11 @@ type ClientHandler struct {
 	// StartOptions.SpanKind will always be set to trace.SpanKindClient
 	// for spans started by this handler.
 	StartOptions trace.StartOptions
+
+	// NoPropagate may be set to true to avoid propagating tags and tracing
+	// on outbound calls. This is appropriate to set on gRPC calls to external
+	// services.
+	NoPropagate bool
```
rakyll: I think I didn't understand the goal of this feature initially. As far as I understand now, not propagating the tags is a security feature. We don't want to propagate tags to untrusted endpoints.

What about having a field here called `IsPublicEndpoint` and not propagation if it is true? So it is more symmetric to the handler field.

ramonza: I don't think "public endpoint" makes sense in the context of a client, only a server. I would be open to deprecating/renaming `IsPublicEndpoint` to `NoPropagate` -- I think that makes sense for both, what do you think?

rakyll: Why not? The outgoing URL can be called a public/insecure endpoint. It makes less sense for a server because it is not an endpoint but more of a starting point.

ramonza: That doesn't agree with my understanding of the word "endpoint", but anyway I removed this part of the PR. I think it is better handled by having custom propagation format support for gRPC (like we do for HTTP) and instead using the same technique of a `None()` propagation format that can be configured.  This will make gRPC and HTTP align better.

ramonza:
Waiting on: https://github.com/census-instrumentation/opencensus-specs/pull/201

PR #943, opened 2018-10-08
Always end execution tracer task

### LINKED ISSUE #942: Execution tracer tasks not not marked as done when Span is not sampled
**Describe the bug**

When using Go 1.11, each `Span` created with the `go.opencensus.io/trace` package also creates a `runtime/trace.Task` as long as the execution tracer is enabled, whether or not the `Span` is sampled or not. Each of these `Task`s has an `End` method, which is put into the `Span.executionTracerTaskEnd` field, that should be called once the `Task` has ended.

The `Task.End` method / `Span.executionTracerTaskEnd` function is called in `Span.End`, but only when the `Span` itself is sampled. Since unsampled `Span`s can still have associated `Task`s, this can lead to unclosed `Task`s.

https://github.com/census-instrumentation/opencensus-go/blob/fb92c347f423413c4522dc97dcd0d464ab30e469/trace/trace.go#L245-L253

**To Reproduce**
1. Add println() calls before the `trace.NewTask` and `Task.End` calls, like this:

```diff
diff --git a/trace/trace_go11.go b/trace/trace_go11.go
index b7d8aaf..91de855 100644
--- a/trace/trace_go11.go
+++ b/trace/trace_go11.go
@@ -27,6 +27,7 @@ func startExecutionTracerTask(ctx context.Context, name string) (context.Context
                // runtime/trace is not enabled.
                return ctx, func() {}
        }
+       println("started")
        nctx, task := t.NewTask(ctx, name)
-       return nctx, task.End
+       return nctx, func() { println("stopped"); task.End() }
 }
```

2. Create an unsampled span while the execution tracer is running.

```go
package main

import (
	"context"
	"os"
	"runtime/trace"
	"time"

	octrace "go.opencensus.io/trace"
)

func main() {
	f, err := os.Create("trace.out")
	if err != nil {
		panic(err)
	}
	defer f.Close()

	if err := trace.Start(f); err != nil {
		panic(err)
	}
	defer trace.Stop()

	ctx := context.TODO()

	ctx, span := octrace.StartSpan(ctx, "example", octrace.WithSampler(octrace.NeverSample()))
	defer span.End()

	for i := 0; i < 10; i++ {
		println(i)
		time.Sleep(time.Duration(i) * 100 * time.Millisecond)
	}
}
```

**Expected behavior**
`Task.End` function should be called. With the above example only the "started" line will be printed. If I change the `octrace.NeverSample()` to `octrace.AlwaysSample()` the "stopped" line will be printed.

Note that even if the `Task.End` method is not called the resulting trace will still be valid and work with the `go tool trace`.

**Additional context**

rakyll:
Fixes #942.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -243,13 +243,16 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
 
 // End ends the span.
 func (s *Span) End() {
+	if s == nil {
+		return
+	}
+	if s.executionTracerTaskEnd != nil {
+		s.executionTracerTaskEnd()
+	}
 	if !s.IsRecordingEvents() {
 		return
 	}
 	s.endOnce.Do(func() {
-		if s.executionTracerTaskEnd != nil {
-			s.executionTracerTaskEnd()
-		}
 		exp, _ := exporters.Load().(exportersMap)
 		mustExport := s.spanContext.IsSampled() && len(exp) > 0
 		if s.spanStore != nil || mustExport {
--- a/trace/trace_test.go
+++ b/trace/trace_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


rakyll:
Added tests, PTAL.

PR #948, opened 2018-10-14
set countFormatter param to int64

### LINKED ISSUE #895: zpages: statsSnapshots counts are of type int with limited ranges and will easily overflow
After examining #893 and #894 I am concerned about the choice of int for fields that counts events, most especially for high traffic applications.

This issue is to discuss that design choice(perhaps if there is a reason why we used int) and perhaps recommend that we switch to uint64 which has a way higher range of 18446744073709551615 which is still representable even on 32bit architectures.

### Data points to illustrate the problem and solution
To put that into perspective, it is trivial to imagine an app with more than 2147483647 events being received by zPages and moreover since we'll have folks deploying zPages into high traffic apps for months and perhaps even years.

If say you have 5,000 QPS(e.g. clicks registered) and each query conservatively generates say 5 RPCs = >=25,000 events counted per second

Throughput (T) = 25,000/s

State|Max range(M)|Time to overflow = M/T (s)|Hours to overflow (hr)
---|---|---|---
Currently|(1<<31)-1 = 2147483647|2147483647/25000 = 85899.34588|23.8609
Proposed|(1<<64)-1 = 18446744073709551615|18446744073709551615/25000 = 737869762948382.1|204963823041.21725

### Summary
Currently for a 5K QPS app that generates say 5RPCs:
* it would take less than 1 day to overflow an int counting field i.e. 23.8609 hours = 0.9942 days
* it would take about 8,540,159,293 hrs or 23,397,696.69420288 years which are waaay more than sufficient for any application

feloy:
Fixes #895

### DIFF
--- a/zpages/formatter_test.go
+++ b/zpages/formatter_test.go
@@ -19,7 +19,7 @@ import "testing"
 
 func TestCountFormatter(t *testing.T) {
 	tests := []struct {
-		in   int
+		in   int64
 		want string
 	}{
 		{-1, " "},
@@ -30,6 +30,9 @@ func TestCountFormatter(t *testing.T) {
 		{1e6, "1.000 M "},
 		{1e9, "1.000 G "},
 		{1e8 + 2e9, "2.100 G "},
+		{1e12, "1.000 T "},
+		{1e15, "1.000 P "},
+		{1e18, "1.000 E "},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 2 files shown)


odeke-em:
Thanks for this PR @feloy and welcome to OpenCensus!

So int64 is alright but in the issue I had recommended uint64. We also forgot to update the struct fields themselves from int. Anyways I've mailed out #952 to fix #895, please take a look.

PR #952, opened 2018-10-20
zpages: fix Snapshot int->*int64 changes + tests on count

### LINKED ISSUE #895: zpages: statsSnapshots counts are of type int with limited ranges and will easily overflow

### LINKED ISSUE #951: zpages: fix the count formatter argument type and StatsSnapshot field types
PR https://github.com/census-instrumentation/opencensus-go/pull/948 changed the signature of countFormatter's argument from int to int64 however the fields of Snapshot that are passed into count during template execution are still int. This issue was reported by a user of the opencensus-service as per https://github.com/census-instrumentation/opencensus-service/issues/123
```shell
zpages: executing template: template: rpcz:31:41: executing "rpcz" at <count>: wrong type for value; expected int64; got int
```

and in fact the test output also reported the runtime template error in
https://ci.appveyor.com/project/opencensusgoteam/opencensus-go/builds/19498802
screen shot 2018-10-20 at 3 53 53 pm
and
https://travis-ci.org/census-instrumentation/opencensus-go/builds/441324783#L1183
screen shot 2018-10-20 at 3 55 06 pm

The proper way to test those templates is by passing in a buffer and invoking execute then checking the output and errors such as
```go
        if err := tmpl.Execute(buf, data); err != nil {
                t.Fatalf("Failed to execute template: %v", err)
        }
```

In the issue https://github.com/census-instrumentation/opencensus-go/issues/895 I had recommended uint64, but we went with int64 which while still large, we have the ability to take advantage of the extra precision as I had described and also ensure graceful wrapping back to 0 on overflow in a couple of millenia

It would also be nice to make the zpages functions testable but also the formatters testable so that we can catch such innocent regressions early enough.

odeke-em:
The signature for countFormatter's arguments was changed
in PR #948 from int to int64 but unfortunately we forgot
to update the int fields from int to the new type.

Issue #948's recommendation was to use uint64 since that
affords us more time to overflow e.g. even at 25QPs
it'd take a couple of millenia to overflow with uint64.

This change completes the change of types of the int variables
but also adds tests that make use of the template functions
and a snapshot and then compare the output or fail. Previously the
tests complained but since they were HTTP tests not attached
to an output buffer that could be inspected, the bug innocently
crept away in standard output and the tests incorrectly passed as in:
* https://ci.appveyor.com/project/opencensusgoteam/opencensus-go/builds/19498802#L700
* https://travis-ci.org/census-instrumentation/opencensus-go/builds/441324783#L1183

This problem was reported by an OpenCensus-Service user in
* https://github.com/census-instrumentation/opencensus-service/issues/123

Fixes #895
Fixes #951

### DIFF
--- a/zpages/formatter_test.go
+++ b/zpages/formatter_test.go
@@ -19,10 +19,9 @@ import "testing"
 
 func TestCountFormatter(t *testing.T) {
 	tests := []struct {
-		in   int64
+		in   uint64
 		want string
 	}{
-		{-1, " "},
 		{0, " "},
 		{1, "1"},
 		{1024, "1024"},
--- a/zpages/rpcz.go
+++ b/zpages/rpcz.go
@@ -170,9 +170,9 @@ type statSnapshot struct {
 	// TODO: compute hour/minute values from cumulative
 	Method           string
 	Received         bool
-	CountMinute      int
-	CountHour        int
-	CountTotal       int
+	CountMinute      uint64
+	CountHour        uint64
+	CountTotal       uint64
 	AvgLatencyMinute time.Duration
 	AvgLatencyHour   time.Duration
 	AvgLatencyTotal  time.Duration
@@ -185,9 +185,9 @@ type statSnapshot struct {
 	OutputRateMinute float64
 	OutputRateHour   float64
 	OutputRateTotal  float64
-	ErrorsMinute     int
-	ErrorsHour       int
-	ErrorsTotal      int
+	ErrorsMinute     uint64
+	ErrorsHour       uint64
+	ErrorsTotal      uint64
 }
 
 type methodKey struct {
@@ -267,7 +267,7 @@ func (s snapExporter) ExportView(vd *view.Data) {
 			}
 			for _, tag := range row.Tags {
 				if tag.Key == ocgrpc.KeyClientStatus && tag.Value != "OK" {
-					s.ErrorsTotal += int(count)
+					s.ErrorsTotal += uint64(count)
 				}
 			}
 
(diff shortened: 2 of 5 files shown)


odeke-em:
Thanks for the reviews @helly25 @ramonza!

odeke-em:
Ahh, Github's consistency problems somehow merged this PR in https://github.com/census-instrumentation/opencensus-go/commit/486d66aed0ffe5d13ed9f0d59f96fa761cd0abcf but yet it hasn't been officially merged.

PR #955, opened 2018-10-22
Bump version to 0.19.0

ramonza:


ramonza:
AppVeyor failures seem like an appveyor issue

winmillwill:
FWIW, there's still no v0.19.0 release on github. It was just weird to see this in the commits to master since v0.18.0 and have that still be the latest release. Module-aware `go get` also seems to be ignorant of any v0.19.0 and thinks v0.18.0 is latest.

So, is it intended that this version is some other notion of version, or is it intended that this is the same notion of version, but it's not really released yet, or something else?

PR #960, opened 2018-10-25
exporter/zipkin: option to set RemoteEndpoint

### LINKED ISSUE #959: exporter/zipkin: allow NewExporter to set remote endpoint
Currently the zipkin exporter was implemented with only the ability to create a localEndpoint in the call
https://godoc.org/go.opencensus.io/exporter/zipkin#NewExporter

https://github.com/census-instrumentation/opencensus-go/blob/96e75b88df843315da521168a0e3b11792088728/exporter/zipkin/zipkin.go#L34-L50

We need the ability to add a remote endpoint if need be. That is because RemoteEndpoint https://github.com/openzipkin/zipkin-go/blob/70244c9703277c13566aac289c530a1143b14cf5/model/span.go#L39-L40 is essential when creating the service graph of elements by Zipkin later.

I am currently wrapping up the Zipkin interceptor for the OpenCensus Agent/Service in https://github.com/census-instrumentation/opencensus-service/pull/111 and the only thing lacking here is the ability to create an OpenCensus exporter that also adds the endpoint.

Since the exporter seems to have suggested optionality of the RemoteEndpoint, perhaps let's add an option `WithRemoteEndpoint` and with that always transmit one when uploading Zipkin spans to a backend.

odeke-em:
Provide an option `WithRemoteEndpoint` that allows
setting the remote endpoint of a Zipkin exporter.
This change is necessary because the constructor
NewExporter only takes in a local endpoint, implying
that remoteEndpoint is optional but when necessary,
we need to send this information along since it is
used by Zipkin to construct a service graph.
I found this issue while working on the OpenCensus service/agent.

Fixes #959

### DIFF
--- a/exporter/zipkin/options.go
+++ b/exporter/zipkin/options.go
@@ -1,0 +1,36 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package zipkin
+
+import "github.com/openzipkin/zipkin-go/model"
+
+type Option interface {
+	mutateExporter(*Exporter)
+}
+
+type remoteEndpoint struct {
+	endpoint *model.Endpoint
+}
+
+var _ Option = (*remoteEndpoint)(nil)
+
+// WithRemoteEndpoint sets the remote endpoint of the exporter.
+func WithRemoteEndpoint(endpoint *model.Endpoint) Option {
+	return &remoteEndpoint{endpoint: endpoint}
+}
+
+func (re *remoteEndpoint) mutateExporter(exp *Exporter) {
+	exp.remoteEndpoint = re.endpoint
+}
--- a/exporter/zipkin/zipkin.go
+++ b/exporter/zipkin/zipkin.go
@@ -27,8 +27,9 @@ import (
 // Exporter is an implementation of trace.Exporter that uploads spans to a
 // Zipkin server.
 type Exporter struct {
-	reporter      reporter.Reporter
-	localEndpoint *model.Endpoint
+	reporter       reporter.Reporter
+	localEndpoint  *model.Endpoint
+	remoteEndpoint *model.Endpoint
 }
 
 // NewExporter returns an implementation of trace.Exporter that uploads spans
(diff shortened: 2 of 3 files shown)


exporter/zipkin/options.go
```diff
@@ -0,0 +1,36 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package zipkin
+
+import "github.com/openzipkin/zipkin-go/model"
+
+type Option interface {
```
ramonza: I think this is a lot of ceremony to accomplish this. I know @rakyll also dislikes the functional options pattern. What about just adding an exported field to the exporter and then users can do:
```go
e := zipkin.NewExporter(...)
e.RemoteEndpoint = "..."
```

exporter/zipkin/options.go
```diff
@@ -0,0 +1,36 @@
...
+	mutateExporter(*Exporter)
+}
+
+type remoteEndpoint struct {
+	endpoint *model.Endpoint
+}
+
+var _ Option = (*remoteEndpoint)(nil)
+
+// WithRemoteEndpoint sets the remote endpoint of the exporter.
```
ramonza: We should document what the remote endpoint is. It is not obvious.

exporter/zipkin/options.go
odeke-em: Unfortunately that makes for an awkward API -- why is localEndpoint an unexported field but RemoteEndpoint is exported? RemoteEndpoint being exported makes it seem then like the important field yet localEndpoint has been the seemingly most mandatory one.

Considering a case by case basis this is the need for an option because:
a) the exporter has been started and localEndpoint which is the more important field isn't exposed
b) for the 1+ year that this exporter has existed, no one has needed remoteEndpoint and am the first, hence the "optionality"

odeke-em:
Thanks for the first pass of review @ramonza! 

>I'm not sure functional options are the right approach here. Consider just adding a field instead - it's much simpler.

It makes for an awkward API if localEndpoint is added in the constructor and then RemoteEndpoint is a field. Perhaps either we'd have to create a new constructor(am not sure that's advisable) or break the signature of the exporter that has been used for the past 1 year. As I had previously mentioned, am using functional options because RemoteEndpoint has been treated as "optional" for the past 1+ years and I am on the edge of those folks that need it, am not sure that'd warrant a field. The Stackdriver exporter constructor uses a struct with fields so that use case matches properly but this one doesn't.

odeke-em:
>his seems like it would be a per-span thing, rather than something that could be set globally on the exporter. For zipkin-go, it is also a per-span field. So I'm a little confused as to why we are we adding it to the exporter - am I missing something here?

Right, it is a per span field and so is LocalEndpoint. according to:

Resource|URL
---|---
Zipkin-Go|[Span definition](https://github.com/openzipkin/zipkin-go/blob/70244c9703277c13566aac289c530a1143b14cf5/model/span.go#L32-L43)
Zipkin's data model|https://zipkin.io/zipkin-api/#/default/post_spans

With the OpenCensus service, I am not using one exporter, I have to create a unique exporter per localEndpoint-remoteEndpoint combination in order to use this exporter. Even as is I still need to make many exporters since spans intercepted from Zipkin producing services exist off different localEndpoints.

I think if we were to redesign this exporter:
a) Allow LocalEndpoint and RemoteEndpoint to be attributes in a span e.g. "zipkin.localEndpoint.*", "zipkin.remoteEndpoint.*"
b) if unset we could retrieve the LocalEndpoint from the localIP and host name

Evidently even the exporter's ExportSpan method passes on localEndpoint https://github.com/census-instrumentation/opencensus-go/blob/96e75b88df843315da521168a0e3b11792088728/exporter/zipkin/zipkin.go#L53-L55

I was contemplating just implementing a fresh exporter to accomodate the per-span localEndpoint+remoteEndpoint needs but I'd prefer we consolidate that work in one place. With my newly proposed design, I'd be able to use just one exporter and it could multiplex on each localEndpoint-remoteEndpoint combination.

basvanbeek:
The Zipkin datamodel indeed describes both local endpoint and remote endpoint as per span setable items. Since most instrumented services are single homed and not proxying/annotating on behalf of others we typically allow a default local endpoint to be set on the tracer level (ex. Zipkin-go) and can be overridden by spans if needed. 

This works great for local endpoints as it's basically self describing their own node in the graph but this won't work for remote endpoints as they typically aren't confined to a single remote service.

For reference I'm adding an issue I raised some time ago on OpenCensus specs to have a better solution for this in OC: https://github.com/census-instrumentation/opencensus-specs/issues/135

rghetia:
@odeke-em should this be closed until there is consensus on how to fix this by resolving spec issue first (https://github.com/census-instrumentation/opencensus-specs/issues/135).

songy23:
Closing because zipkin exporter has been moved to https://github.com/census-ecosystem/opencensus-go-exporter-zipkin.

PR #962, opened 2018-10-30
exporter/prometheus: allow labels to be configured

### LINKED ISSUE #961: exporter/prometheus: Allow configuring of constant labels in exporter
The Prometheus exporter allows a namespace to be specified, but Prometheus often recommends using labels to differentiate between services for example (https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels), while a namespace will make it harder to track metrics that are shared between services while maintaining a separation of other metrics.

The exporter options https://github.com/census-instrumentation/opencensus-go/blob/4f7fcb478fc0e34627b4da801e1502ee05a87127/exporter/prometheus/prometheus.go#L46 should allow a slice of constant labels to be configured, which are registered with each view. These would be used as input to https://github.com/census-instrumentation/opencensus-go/blob/4f7fcb478fc0e34627b4da801e1502ee05a87127/exporter/prometheus/prometheus.go#L83

Using namespaces to differentiate between service. Has the downside of making it harder to combine metrics for shared packages.

**Additional context**
Ran into this while creating and exploring Prometheus metrics for shared packages.

johanbrandhorst:
Allows the user to configure labels consistent across
all views for the exporter.

Fixes #961

### DIFF
--- a/exporter/prometheus/prometheus.go
+++ b/exporter/prometheus/prometheus.go
@@ -44,9 +44,10 @@ type Exporter struct {
 
 // Options contains options for configuring the exporter.
 type Options struct {
-	Namespace string
-	Registry  *prometheus.Registry
-	OnError   func(err error)
+	Namespace   string
+	Registry    *prometheus.Registry
+	OnError     func(err error)
+	ConstLabels prometheus.Labels // ConstLabels will be set as labels on all views.
 }
 
 // NewExporter returns an exporter that exports stats to Prometheus.
@@ -80,7 +81,7 @@ func (c *collector) registerViews(views ...*view.View) {
 				viewName(c.opts.Namespace, view),
 				view.Description,
 				tagKeysToLabels(view.TagKeys),
-				nil,
+				c.opts.ConstLabels,
 			)
 			c.registeredViewsMu.Lock()
 			c.registeredViews[sig] = desc
--- a/exporter/prometheus/prometheus_test.go
+++ b/exporter/prometheus/prometheus_test.go
@@ -342,3 +342,94 @@ func TestCumulativenessFromHistograms(t *testing.T) {
 		t.Fatalf("\ngot:\n%s\n\nwant:\n%s\n", got, want)
 	}
 }
+
+func TestConstLabelsIncluded(t *testing.T) {
+	constLabels := prometheus.Labels{
+		"service": "spanner",
+	}
+	measureLabel, _ := tag.NewKey("method")
+
+	exporter, err := NewExporter(Options{
+		ConstLabels: constLabels,
+	})
+	if err != nil {
+		t.Fatalf("failed to create prometheus exporter: %v", err)
+	}
+	view.RegisterExporter(exporter)
+	defer view.UnregisterExporter(exporter)
+
+	names := []string{"foo", "bar", "baz"}
+
+	var measures mSlice
+	for _, name := range names {
+		measures.createAndAppend("tests/"+name, name, "")
+	}
+
+	var vc vCreator
+	for _, m := range measures {
+		vc.createAndAppend(m.Name(), m.Description(), []tag.Key{measureLabel}, m, view.Count())
+	}
+
+	if err := view.Register(vc...); err != nil {
+		t.Fatalf("failed to create views: %v", err)
+	}
+	defer view.Unregister(vc...)
+
+	view.SetReportingPeriod(time.Millisecond)
+
+	ctx, _ := tag.New(context.Background(), tag.Upsert(measureLabel, "issue961"))
+	for _, m := range measures {
+		stats.Record(ctx, m.M(1))
+	}
+
+	srv := httptest.NewServer(exporter)
+	defer srv.Close()
+
+	var i int
+	var output string
+	for {
+		time.Sleep(10 * time.Millisecond)
+		if i == 1000 {
+			t.Fatal("no output at /metrics (10s wait)")
+		}
+		i++
+
+		resp, err := http.Get(srv.URL)
+		if err != nil {
+			t.Fatalf("failed to get /metrics: %v", err)
+		}
+
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatalf("failed to read body: %v", err)
+		}
+		resp.Body.Close()
+
+		output = string(body)
+		if output != "" {
+			break
+		}
+	}
+
+	if strings.Contains(output, "collected before with the same name and label values") {
+		t.Fatal("metric name and labels being duplicated but must be unique")
+	}
+
+	if strings.Contains(output, "error(s) occurred") {
+		t.Fatal("error reported by prometheus registry")
+	}
+
+	want := `# HELP tests_bar bar
+# TYPE tests_bar counter
+tests_bar{method="issue961",service="spanner"} 1
+# HELP tests_baz baz
+# TYPE tests_baz counter
+tests_baz{method="issue961",service="spanner"} 1
+# HELP tests_foo foo
+# TYPE tests_foo counter
+tests_foo{method="issue961",service="spanner"} 1
+`
+	if output != want {
+		t.Fatal("output differed from expected")
+	}
+}


exporter/prometheus/prometheus.go
```diff
@@ -47,6 +47,7 @@ type Options struct {
 	Namespace string
 	Registry  *prometheus.Registry
 	OnError   func(err error)
+	Labels    prometheus.Labels
```
ramonza: I think we should call this "ConstLabels" (to distinguish them from tag-derived labels) and document that these will be set on all views.
```suggestion
	ConstLabels    prometheus.Labels // ConstLabels will be set as labels on all views.
```

johanbrandhorst: Done! Couldn't accept the suggestion because it wanted to make a new commit and it'd only fix part of it :(.

johanbrandhorst:
Fixed CI failure :grimacing:

odeke-em:
@johanbrandhorst perhaps a test too?

odeke-em:
To avoid blocking you and others waiting for this feature since it has been 4+ days, I can merge this with the premise of you sending a follow-up PR for the test, as the fix is straightforward but we'll need the test to lock-in this behavior.

johanbrandhorst:
I'll gladly submit a test for this, sorry I hadn't added one already.

johanbrandhorst:
It looks straight forward, give me 30 minutes.

odeke-em:
>It looks straight forward, give me 30 minutes.

No rush and thank you!

johanbrandhorst:
Will that do?

exporter/prometheus/prometheus_test.go
```diff
@@ -342,3 +342,84 @@ func TestCumulativenessFromHistograms(t *testing.T) {
 		t.Fatalf("\ngot:\n%s\n\nwant:\n%s\n", got, want)
 	}
 }
+
+func TestConstLabelsIncluded(t *testing.T) {
+	key, label := "service", "spanner"
+	exporter, err := NewExporter(Options{
+		ConstLabels: prometheus.Labels{
+			key: label,
```
odeke-em: How is it possible that the literal's name is `key` i.e. unexported, instead of `Key`?
Did you run this test?

exporter/prometheus/prometheus_test.go
```diff
@@ -342,3 +342,84 @@ func TestCumulativenessFromHistograms(t *testing.T) {
...
+
+	if err := view.Register(vc...); err != nil {
+		t.Fatalf("failed to create views: %v", err)
+	}
+	defer view.Unregister(vc...)
+
+	view.SetReportingPeriod(time.Millisecond)
+
+	for _, m := range measures {
+		stats.Record(context.Background(), m.M(1))
```
odeke-em: Since we are testing for the interaction between const labels and other labels, it'd be useful to record some tags in the context before recording these measurements e.g.
```go
ctx, _ := tag.New(context.Background(), tag.Upsert(keyExtraKey, "issue961"))
for _, m := range measures {
    stats.Record(ctx, m.M(1))
}
```

exporter/prometheus/prometheus_test.go
```diff
@@ -342,3 +342,84 @@ func TestCumulativenessFromHistograms(t *testing.T) {
...
+	}
+
+	srv := httptest.NewServer(exporter)
```
odeke-em: I don't think we need to run a server, we can just use `httptest.ResponseRecorder` https://golang.org/pkg/net/http/httptest/#NewRecorder

exporter/prometheus/prometheus_test.go
```diff
@@ -342,3 +342,84 @@ func TestCumulativenessFromHistograms(t *testing.T) {
...
+	var i int
+	var output string
+	for {
+		time.Sleep(10 * time.Millisecond)
+		if i == 1000 {
+			t.Fatal("no output at /metrics (10s wait)")
+		}
+		i++
+
+		resp, err := http.Get(srv.URL)
```
odeke-em:
```go
resrec := httptest.NewRecorder()
req, _ := http.NewRequest("GET", "https://example.org", nil)
exporter.ServeHTTP(resrec, req)
res := resrec.Result()
body := ioutil.ReadAll(res.Body)
```
should suffice instead of a server

exporter/prometheus/prometheus_test.go
```diff
@@ -342,3 +342,84 @@ func TestCumulativenessFromHistograms(t *testing.T) {
...
+
+	if strings.Contains(output, "collected before with the same name and label values") {
+		t.Fatal("metric name and labels being duplicated but must be unique")
+	}
+
+	if strings.Contains(output, "error(s) occurred") {
+		t.Fatal("error reported by prometheus registry")
+	}
+
+	for _, name := range names {
```
odeke-em: It'd be nice to have a static enumeration of the output e.g.
```go
want := `tests_foo{....
```
with the values clearly inspectable

exporter/prometheus/prometheus_test.go
```diff
@@ -342,3 +342,84 @@ func TestCumulativenessFromHistograms(t *testing.T) {
...
+		}
+		resp.Body.Close()
+
+		output = string(body)
+		if output != "" {
+			break
+		}
+	}
+
+	t.Log(output)
```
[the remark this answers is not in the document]
johanbrandhorst: Doh! Good spot.

exporter/prometheus/prometheus_test.go
```diff
@@ -342,3 +342,84 @@ func TestCumulativenessFromHistograms(t *testing.T) {
...
+	key, label := "service", "spanner"
+	exporter, err := NewExporter(Options{
+		ConstLabels: prometheus.Labels{
+			key: label,
+		},
+	})
+	if err != nil {
+		t.Fatalf("failed to create prometheus exporter: %v", err)
+	}
+	view.RegisterExporter(exporter)
```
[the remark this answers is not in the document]
johanbrandhorst: I'll go ahead and add this to the other test where I copied this from too.

exporter/prometheus/prometheus_test.go
johanbrandhorst: Yes, I have run the test. `prometheus.Labels` is a map type.

exporter/prometheus/prometheus_test.go
johanbrandhorst: Sure, can do!

exporter/prometheus/prometheus_test.go
johanbrandhorst: I'll replace the other test where I copied this from too.

exporter/prometheus/prometheus_test.go
johanbrandhorst: OK, can do!

exporter/prometheus/prometheus_test.go
johanbrandhorst: I'm not sure this _will_ suffice. I think we have to ping it a couple of times for metrics to be reported, which is why the loop was there (note the reporting period is 1ms). Is it okay if I leave this in? It breaks the test as suggested.

exporter/prometheus/prometheus_test.go
odeke-em: Oh, aha gotcha. Let's make that obvious then, I think static assignments take away such ambiguities and also below I added a suggestion to enumerate the output in a statically created string, that way we don't have to dynamically create the output

johanbrandhorst:
Thanks for the suggestions, I hope the new changes should resolve the issues.

exporter/prometheus/prometheus_test.go
```diff
@@ -342,3 +342,88 @@ func TestCumulativenessFromHistograms(t *testing.T) {
...
+
+	if strings.Contains(output, "collected before with the same name and label values") {
+		t.Fatal("metric name and labels being duplicated but must be unique")
+	}
+
+	if strings.Contains(output, "error(s) occurred") {
+		t.Fatal("error reported by prometheus registry")
+	}
+
+	for _, name := range names {
```
odeke-em: I actually meant literally 
want := `tests_foo{method="issue961",service="spanner"} 1
tests_bar{method="issue961",service="spanner"} 1`
instead of this iterations on the names
and then wholly compare the outputs i.e.
```go
if output != want {
  t.Fatal
}
```

johanbrandhorst: Ah, I see!

johanbrandhorst: I've removed the iteration, hopefully this is now what you were expecting :).

PR #964, opened 2018-11-01
ochttp: add DefaultAttributes to each span

yancl:
Just like what the [PR](https://github.com/opencensus-integrations/ocsql/pull/4) want to do, thanks.

### DIFF
--- a/plugin/ochttp/client.go
+++ b/plugin/ochttp/client.go
@@ -61,6 +61,9 @@ type Transport struct {
 	// httptrace package.
 	NewClientTrace func(*http.Request, *trace.Span) *httptrace.ClientTrace
 
+	// DefaultAttributes will be set to each span as default.
+	DefaultAttributes []trace.Attribute
+
 	// TODO: Implement tag propagation for HTTP.
 }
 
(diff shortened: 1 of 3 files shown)


basvanbeek:
I'm not sure if we need it here. I personally would use the NewClientTrace hook. Not only does it allow you to add static default annotations like this PR tries to provide but also have annotations be extracted from the HTTP request payload.

yancl:
Oops, thanks @basvanbeek. you are right. I didn't notice the NewClientTrace hook. Although i didn't see the same mechanism for the ochttp server, but i think it is enough for me now.

Thanks again for your help.

basvanbeek:
@yancl I apologize for actually rushing the answer.

The NewClientTrace could indeed be used for this but it is not meant for it as it specifically wants to allow for the `net/http/httptrace' package to be used for annotating spans. It feels dirty to abuse it.

That said I would prefer a hook on both client and server end which would provide us access to the http Request data and the current span so not only static annotation payloads can be automatically inserted. 

Question is if we need it. It is also possible to add middlewares for this for those that want to adjust span data. A bit more tedious to do but possible.

@ramonza @odeke-em @rakyll any insights?

yancl:
Thank you @basvanbeek . That would be much better if we could add a hook or middleware for manipulating span data:)

basvanbeek:
@yancl I've quickly set-up an example how to do server side middleware:
https://gist.github.com/basvanbeek/6a486ddf13fe3e7ac9537c742a96e525

yancl:
Thanks @basvanbeek, you are so nice!! The gist looks great to me and works well except some small changes as follows:

From:
```
func annotatingMiddleware(next http.Handler) http.Handler {
	return http.HandlerFunc(
...
```

To:
```
func annotatingMiddleware(next http.Handler) http.Handler {
	return &ochttp.Handler{
            Handler: http.HandlerFunc(
...
```

Since the middleware way works well, i will close the PR. Thanks again for your great review and help, @basvanbeek :)

Issue #940, opened 2018-10-04
Add Metrics APIs as per "metrics.proto" 


mayurkale22:
Add Metrics package and below classes based on [metrics.proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto). Although Metrics proto is still being finalized, We think it is safe to start working on value classes like LabelKey, TimeSeries, Point etc.

- [x] Metrics: Add LabelKey and LabelValue.
- [x] Metrics: Add Value and Point.
- [x] Metrics: Add TimeSeries.
- [x] Metrics: Add MetricDescriptor.
- [x] Metrics: Add Metric.
- [x] Metrics: Add Summary and Distribution.
- [x] Metrics: Produce and store Metrics in Stats impl.(#1078)

Counterpart classes available in Java for reference : https://github.com/census-instrumentation/opencensus-java/tree/master/api/src/main/java/io/opencensus/metrics

ramonza:
At the moment, I am leaning towards relying directly on the protobuf generated classes rather than re-implementing them unless anyone has objections.

bogdandrutu:
@ramonza we should not add a protobuf dependency on the core library (that was our decision).

johanbrandhorst:
Will this include Go specific metrics like `gc_pause` time distributions?

bogdandrutu:
Yes we can have that as a "contrib" package.

rghetia:
all tasks are complete.

PR #969, opened 2018-11-08
Metric and gauge support

ramonza:
Updates: #940


ramonza:
cc @fabxc

rakyll:
Can you split this PR into multiple ones? It would be easier to have a PR only with APIs and once we agree on them, implementation can follow.

ramonza:
Going to split this up as requested

PR #975, opened 2018-11-15
Add metric API

ramonza:


ramonza:
The follow-ups are here: https://github.com/ramonza/opencensus-go/pulls

ramonza:
@rakyll split up the PR, PTAL

metric/producer.go
```diff
@@ -0,0 +1,84 @@
...
+	ind     uint64
+}
+
+var _ Producer = (*Registry)(nil)
+
+// NewRegistry creates a new Registry.
+func NewRegistry() *Registry {
+	m := &Registry{
+		sources: make(map[*uintptr]Producer),
+		ind:     0,
```
fabxc: Unused remainder from before the `uintptr` trick :)?

metric/doc.go
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package metric contains a data model and exporter support for metrics.
+// The data model is derived from the metric protocol buffers:
+//   https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto
```
rakyll: User doesn't have to know about the proto, please remove this.

metric/metric.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"time"
+
+	"go.opencensus.io/resource"
+)
+
+// LabelValue represents the value of a label. A missing value (nil) is distinct
+// from an empty string value.
+type LabelValue *string
```
rakyll: This is a unlikely type. Why not simply `type LabelValue { Value string }`?

metric/metric.go
```diff
@@ -0,0 +1,55 @@
...
+
+// NewLabelValue creates a new non-nil LabelValue that represents the given string.
```
rakyll: This is redundant.

LabelValue("val") would just work.

metric/point.go
```diff
@@ -0,0 +1,237 @@
...
+// distribution.
+type BucketOptions struct {
+	// ExplicitBoundaries specifies a set of buckets with arbitrary upper-bounds.
+	// This defines len(bounds) + 1 (= N) buckets. The boundaries for bucket
+	// index i are:
+	//
+	// [0, bucket_bounds[i]) for i == 0
+	// [bucket_bounds[i-1], bucket_bounds[i]) for 0 < i < N-1
+	// [bucket_bounds[i-1], +infinity) for i == N-1
+	ExplicitBoundaries []float64
```
rakyll: Just Boundaries?

metric/point.go
```diff
@@ -0,0 +1,237 @@
...
+// Value types. New value types may be added.
+const (
+	ValueTypeDouble ValueType = iota
+	ValueTypeInt64
+	ValueTypeDistribution
+	ValueTypeSummary
+)
+
+// NewDoublePoint creates a new Point holding a float64 value.
+func NewDoublePoint(t time.Time, val float64) Point {
```
rakyll: NewFloat64Point

metric/point.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+//go:generate stringer -type ValueType
+
+// ValueType is the type of value held in a point.
+type ValueType int
+
+// Value types. New value types may be added.
+const (
+	ValueTypeDouble ValueType = iota
```
rakyll: There is no doubles in Go, if the name is dictated by the spec, it should be flex-ed.

metric/producer.go
```diff
@@ -0,0 +1,84 @@
...
+	// The returned metrics should be unique for each combination of name and
+	// resource.
+	Read() []*Metric
+}
+
+// Registry maintains a set of metric producers for exporting. Most users will
+// rely on the DefaultRegistry.
+type Registry struct {
+	mu      sync.RWMutex
+	sources map[*uintptr]Producer
```
rakyll: sources map[Producer]struct{}?

metric/producer.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	return ms
+}
+
+// AddProducer adds a producer to this registry.
+func (m *Registry) AddProducer(source Producer) (remove func()) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if source == m {
+		panic("attempt to add registry to itself")
```
rakyll: Why is this statically possible?

metric/metric.go
```diff
@@ -0,0 +1,55 @@
...
+	Description string   // human-readable description
+	Unit        Unit     // units for the measure
+	Type        Type     // type of measure
+	LabelKeys   []string // label keys
+}
+
+// Metric represents a quantity measured against a resource with different
+// label value combinations.
+type Metric struct {
+	Descriptor *Descriptor        // metric descriptor
```
rakyll: Descriptor Descriptor 

No need for a pointer.

metric/point.go
```diff
@@ -0,0 +1,237 @@
...
+	VisitInt64Value(int64)
+	VisitDistributionValue(*Distribution)
+	VisitSummaryValue(*Summary)
+}
+
+// ReadValue accepts a ValueVisitor and calls the appropriate method with the
+// value of this point.
+// Consumers of Point should use this in preference to switching on the type
+// of the value directly, since new value types may be added.
+func (p Point) ReadValue(vv ValueVisitor) {
```
rakyll: Is this something that exporters would use? If yes, maybe we shouldn't have it in this package.

metric/unit.go
```diff
@@ -0,0 +1,11 @@
+package metric
```
rakyll: Missing license.

metric/producer_test.go
```diff
@@ -0,0 +1,29 @@
+package metric
```
rakyll: Missing license.

metric/metric.go
ramonza: The intention here is to be able to represent no value (nil) and empty string value distinctly. The protobuf does this, so we need to as well to be compatible. Using a type like you suggest would not allow us to represent these two distinctly.

metric/point.go
ramonza: I was on the fence about this. I agree it looks very strange. Will change it.

metric/point.go
ramonza: Yes, it is. This whole package is mostly for use by exporters.

metric/producer.go
ramonza: Producer may not be suitable as a map key. This way, we don't have to care.

metric/producer.go
ramonza: *Registry implements Producer

ramonza:
Updated @rakyll @rghetia PTAL

rghetia:
LGTM

metric/metric.go
rakyll: Protobuf types are not an idiomatic due to the limitations and code-generation nature of the problem. In handcrafted libraries, we are doing better than proto.

metric/producer.go
rakyll: The question is it might be broken to make Registry implementing Read() []*Metric.

Can't we make it implement ReadAll() []*Metric to differentiate it from the producers? Otherwise, we should introduce a MultiProducer a la io.MultiReader instead of this type.

metric/producer.go
rakyll: In what case? Also why a pointer to uintptr?

metric/point.go
rakyll: Is there any end to end example usage for the new APIs here? It is not clear what the package is for. If it is mainly for the exporters, we shouldn't use a top-level package like "metric"  where we would like to put the developer-facing APIs to.

metric/metric.go
ramonza: We proposed to replace with this with `struct { Value string, HasValue bool }` to differentiate between "missing" and an empty string value.

metric/producer.go
ramonza: Will rename

Issue #963, opened 2018-10-31
Drop negative bucket bounds and negative values in record API


mayurkale22:
Negative values are currently not supported by any of the backends that we support and have histograms (Stackdriver and Prometheus). We have decided to implement a backwards incompatible change and drop negative values in the Record API (We will start with logging an error for few versions then throw an exception).

This could be a breaking change if users are recording negative value for any measure. We planned to implement this change asap in all the libraries and explain this behaviour to the current users very clearly.

Counterpart in Java: census-instrumentation/opencensus-java#1499 and census-instrumentation/opencensus-java#1525

PR #979, opened 2018-11-21
Return an error if negative bucket bounds are provided

ramonza:
Fixes: #963

### DIFF
--- a/stats/view/view.go
+++ b/stats/view/view.go
@@ -17,6 +17,7 @@ package view
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
@@ -69,6 +70,8 @@ func (v *View) same(other *View) bool {
 		v.Measure.Name() == other.Measure.Name()
 }
 
+var ErrNegativeBucketBounds = errors.New("negative bucket bounds not supported")
+
 // canonicalize canonicalizes v by setting explicit
 // defaults for Name and Description and sorting the TagKeys
 func (v *View) canonicalize() error {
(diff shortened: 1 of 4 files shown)


stats/view/view.go
```diff
@@ -90,6 +93,12 @@ func (v *View) canonicalize() error {
 	sort.Slice(v.TagKeys, func(i, j int) bool {
 		return v.TagKeys[i].Name() < v.TagKeys[j].Name()
 	})
+	sort.Float64s(v.Aggregation.Buckets)
+	for _, b := range v.Aggregation.Buckets {
+		if b < 0 {
+			return ErrNegativeBucketBounds
```
mayurkale22: We just want to log a warning in case of negative bounds and silently remove 0 bounds, counterpart: https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/stats/BucketBoundaries.java#L81-L90.

stats/view/view_test.go
```diff
@@ -444,3 +446,31 @@ func TestRegisterAfterMeasurement(t *testing.T) {
 		t.Error("View should have data")
 	}
 }
+
+func TestViewRegister_negativeBucketBounds(t *testing.T) {
+	m := stats.Int64("TestViewRegister_negativeBucketBounds", "", "")
+	v := &View{
+		Measure:     m,
+		Aggregation: Distribution(-1, 2),
```
mayurkale22: Please add tests for 0 bounds also.

mayurkale22:
Please fix the build.

stats/view/view.go
bogdandrutu: I think it is good to return an error if the API already had the error.

ramonza:
@mayurkale22 are you okay with leaving the error as is? PTAL

mayurkale22:
> @mayurkale22 are you okay with leaving the error as is? PTAL

Ok, sure.

PR #988, opened 2018-11-28
Allow views on measurements without required tags to succeed

johanbrandhorst:
See #659. So far this just adds a failing test, but I intend to submit a fix for this once the correct behaviour has been decided.

### DIFF
--- a/exporter/prometheus/prometheus_test.go
+++ b/exporter/prometheus/prometheus_test.go
@@ -433,3 +433,78 @@ tests_foo{method="issue961",service="spanner"} 1
 		t.Fatal("output differed from expected")
 	}
 }
+
+func TestViewMeasureWithoutTag(t *testing.T) {
+	exporter, err := NewExporter(Options{})
+	if err != nil {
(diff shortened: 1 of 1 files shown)


exporter/prometheus/prometheus_test.go
```diff
@@ -433,3 +433,78 @@ tests_foo{method="issue961",service="spanner"} 1
...
+		t.Fatal("metric name and labels being duplicated but must be unique")
+	}
+
+	if strings.Contains(output, "error(s) occurred") {
+		t.Fatal("error reported by prometheus registry")
+	}
+
+	want := `# HELP tests_foo foo
+# TYPE tests_foo counter
+tests_foo{key=""} 1
```
johanbrandhorst: This is assuming that we would just create an empty value label for missing tags.

johanbrandhorst:
Label this as `do not merge` please (even though CI should fail).

bogdandrutu:
Working on a fix.

bogdandrutu:
Fixed in https://github.com/census-instrumentation/opencensus-go/pull/989

PR #989, opened 2018-11-28
Fix prometheus exporter when tags are missing.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-go/issues/659

### DIFF
--- a/exporter/prometheus/prometheus.go
+++ b/exporter/prometheus/prometheus.go
@@ -208,7 +208,7 @@ func (c *collector) Collect(ch chan<- prometheus.Metric) {
 func (c *collector) toMetric(desc *prometheus.Desc, v *view.View, row *view.Row) (prometheus.Metric, error) {
 	switch data := row.Data.(type) {
 	case *view.CountData:
-		return prometheus.NewConstMetric(desc, prometheus.CounterValue, float64(data.Value), tagValues(row.Tags)...)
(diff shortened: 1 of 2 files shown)


exporter/prometheus/prometheus.go
```diff
@@ -255,14 +255,6 @@ func tagKeysToLabels(keys []tag.Key) (labels []string) {
 	return labels
 }
 
-func tagsToLabels(tags []tag.Tag) []string {
-	var names []string
-	for _, tag := range tags {
-		names = append(names, internal.Sanitize(tag.Key.Name()))
```
odeke-em: Why did you remove this? It is imperative that we sanitize keys for Proemtheus otherwise "org/foo" will fail for Prometheus which expects "org_foo".

bogdandrutu: This is unused as you can see I did not change any code that calls it, and this seems to be a package private function.

odeke-em: Gotcha, but it is noisy in this PR and unrelated to this change.

bogdandrutu: I see your point, but dead code is the thing that I "like" the most :) happy to do a different PR if you feel strong.

bogdandrutu: Reverted and I would send a different PR.

PR #994, opened 2018-12-04
Add metricexport package with metric API

ramonza:


metric/metricexport/point.go
```diff
@@ -0,0 +1,239 @@
...
+	TypeGaugeFloat64
+	TypeGaugeDistribution
+	TypeCumulativeInt64
+	TypeCumulativeFloat64
+	TypeCumulativeDistribution
+	TypeSummary
+)
+
+// IsGuage returns true if the metric type represents a gauge-type value.
+func (t Type) IsGuage() bool {
```
bogdandrutu: Why is this exported? Can we postpone to export it until is necessary?

metric/metricexport/point.go
```diff
@@ -0,0 +1,239 @@
...
+
+	"go.opencensus.io/exemplar"
+)
+
+// Point is a single data point of a time series.
+type Point struct {
+	// Time is the point in time that this point represents in a time series.
+	Time time.Time
+	// ValueType is the type of value held in this point.
+	ValueType ValueType
```
bogdandrutu: To avoid users writing a switch statement and force them to use the ReadValue, is it not better to not export this type?

metric/metricexport/metric.go
```diff
@@ -0,0 +1,58 @@
...
+
+// LabelValue represents the value of a label. A missing value (nil) is distinct
+// from an empty string value.
+type LabelValue struct {
+	Value   string
+	Present bool
+}
+
+// NewLabelValue creates a new non-nil LabelValue that represents the given string.
+func NewLabelValue(val string) LabelValue {
```
bogdandrutu: How do people get access to a nil LabelValue? Should we create a constant NilLabelValue?

metric/metricexport/point.go
ramonza: I'm going to remove ValueType completely since it's not needed.

metric/metricexport/metric.go
ramonza: They can just use the zero value: LabelValue{}

ramonza: I will document this.

metric/metricexport/point.go
ramonza: I will remove it

ramonza:
I think it would be better to have at least one use of the metrics API committed along with it, so I am going to combine this with the Gauge API PR. That will make it clearer why we need the Producer (Gauge is a Producer).

PR #999, opened 2018-12-14
Add metric package containing Gauge support

ramonza:


metric/gauge.go
```diff
@@ -0,0 +1,208 @@
...
+	keys     []string
+	newEntry func() gaugeEntry
+}
+
+type gaugeEntry interface {
+	read(t time.Time) metricdata.Point
+}
+
+// Read returns the current values of the gauge as a metric for export.
+func (g *Gauge) Read() []*metricdata.Metric {
```
bogdandrutu: Why is this public? Not sure I understand the design of these package. I was expecting to have a MetricsProducer (name is probably not the best) that implements the metricsexport.Producer, when create a new Gauge add it to a list of created Gauges in the MetricsProducer.

metric/gauge.go
```diff
@@ -0,0 +1,208 @@
...
+				LabelKeys:   keys,
+			},
+		},
+	}
+	return g
+}
+
+// Set sets the current gauge value.
+func (g *Float64Gauge) Set(val float64, labelVals ...metricdata.LabelValue) {
+	ge := g.getEntry(labelVals).(*float64Entry)
```
bogdandrutu: getEntry seems to be expensive:
* need to serialize the entries
* possible lock contention on the sync.Map

This were the reason we exposed the "Entry" (in java LongPoint) to the user to allow them to cache the entry and avoid all the possible performance implications.

metric/gauge.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func (v *int64GaugeValue) read(t time.Time) metricdata.Point {
+	return metricdata.NewInt64Point(t, atomic.LoadInt64(&v.val))
+}
+
+// NewInt64Gauge creates a new int64-valued gauge and adds it to the
+// given registry.
+func NewInt64Gauge(name, description string, unit metricdata.Unit, keys ...string) *Int64Gauge {
+	g := &Int64Gauge{
```
bogdandrutu: Shouldn't we keep a reference to the newly created Gauge in order to be able to read/export the value for this Gauge?

metric/gauge.go
ramonza: I see, so basically you want it to work like it does in Java? I thought this would be a more natural Go API where you directly add the Gauge as a metricexport.Producer (this is why this method is public). I will make it work like Java instead by adding an intermediate MetricManager (or something like that).

metric/gauge.go
ramonza: I thought it would be worth not exposing it because it would make the API more complex. I will change it to expose it directly.

metric/gauge.go
ramonza: I envisioned that the Gauge would be added to a metricexport.Registry object as in the example. But I will make it work like Java and probably have a MetricManager that can make Gauges instead of having them be independent objects like this.

ramonza: Done

ramonza:
@bogdandrutu PTAL

I also removed the metricexporter.Registry to free up the name "Registry". I plan to incorporate its functionality into metric.Registry when I submit the stats-to-metrics PR.

Kuqd:
Hello @ramonza I think I need this :), do you have more documentation on the usage ?

ramonza:
@Kuqd the design is largely based off the Java implementation. There are some notes here as well: https://github.com/census-instrumentation/opencensus-specs/blob/master/metrics/Metrics.md

@bogdandrutu needs to review and approve before this can be merged.

Issue #1007, opened 2019-01-03
Get SpanData of spans


smthpickboy:
We want to implement log correlation for go(#950), so we need to access Span.data for parent span id and span name etc, and log them.

Is there a way to access Span.data? We can access it in Exporters, but only sampled trace/span can reach Exporter, while we want to log span info even if trace/span is NOT sampled.

Thanks!

basvanbeek:
For log correlation you only need the SpanContext which holds items like TraceID and SpanID which is sufficient for correlation purposes and also propagated in cases when a Trace is not sampled. 

Span.data does not hold any required information to allow for log correlation and is only available for Spans that are sampled.

smthpickboy:
I think Parent Span ID is very useful for rebuilding structure of traces/spans from logs when traces are not sampled. Span Name further adds useful info to the rebuilded results.

rghetia:
if log data is used to rebuild traces then why not sample the trace?

smthpickboy:
@rghetia Because whether a trace will be sampled is determined at the time when the trace/span is created, but that's not where logging happens. When later the program decides to do some logging, there's no way to make sure this trace will be sampled and exported.

bogdandrutu:
I am not sure I completely understand this,

bogdandrutu:
My bad of pressing close button.

rghetia:
> @rghetia Because whether a trace will be sampled is determined at the time when the trace/span is created, but that's not where logging happens. When later the program decides to do some logging, there's no way to make sure this trace will be sampled and exported.

unless log is created for every span rebuilding trace from log is not going to be useful. At that point it is as good as turning on always-sampling mode. Logging every span is probably more expensive then turning on sampling.

smthpickboy:
Always-sampling would be a waste of resources apparently. But when error happens, it would be nice to recover/rebuild more information as much as possible. Of course we can't recover the whole trace/span structure, but still, things like span name can provide valuable info for diagnosing problems, since not all the people in a team are familiar with the source code. And, we are planning to relate span names to the metrics names, so we can process and diagnose problems automatically in the future.

NogginBops:
Any update on this?
Is there currently no way to get the SpanContext from a span?
I can't seem to find any way to get SpanID or TraceID from a span either.

This would be really useful to have so that one could implement log correlation yourself.

bogdandrutu:
[...]

PR #1008, opened 2019-01-03
Add get data of span method

yore-new:
return the data of span to allow accessing all info of a span.
just return the value of data instead a pointer to avoid changing by user.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -286,6 +286,14 @@ func (s *Span) makeSpanData() *SpanData {
 	return &sd
 }
 
+// SpanData returns the data of the span
+func (s *Span) SpanData() SpanData {
+	if s == nil || s.data == nil {
+		return SpanData{}
+	}
+	return *s.data
+}
+
 // SpanContext returns the SpanContext of the span.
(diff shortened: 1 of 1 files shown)


basvanbeek:
As explained in #1007 there is no need to provide access to the entire span data payload for correlation purposes. This is what SpanContext is for. The other problem is you return a shallow copy of span data which would allow for mutating for instance the Attributes map (and get some concurrent access nasties).
For now, I do not see a reason for allowing access to even a full copy of span data.

rakyll:
Why do we need to give SetForceGenerateSpanData as an option to the user?

@bogdandrutu, do we have any requirements about immutability of the span data?

sydnash:
> As explained in #1007 there is no need to provide access to the entire span data payload for correlation purposes. This is what SpanContext is for. The other problem is you return a shallow copy of span data which would allow for mutating for instance the Attributes map (and get some concurrent access nasties).
> For now, I do not see a reason for allowing access to even a full copy of span data.

but i cann't get parent span id from SpanContext, could add a parent span id to the SpanContext?

PR #1015, opened 2019-01-24
Add option to configure limits for span data

rghetia:
for attributes, annotations message events and links.


rghetia:
> LGTM, please fix the build.

Should this [version](https://github.com/census-instrumentation/opencensus-go/blob/2b5032d79456124f42db6b7eb19ac6c155449dc2/opencensus.go#L20) be now 0.20.0? If that is the case then it is quite confusing.

PR #1029, opened 2019-02-12
RFC: add SpanContext.LocalRootSpanID

axw:
This is a companion to https://github.com/census-instrumentation/opencensus-specs/issues/229. These are the changes that would be necessary in order for Elastic APM to implement an exporter, as in https://github.com/elastic/apm-agent-go/compare/master...axw:opencensus.

I'd appreciate feedback either here or on the spec issue, indicating whether this might be acceptable.

### DIFF
--- a/trace/trace_test.go
+++ b/trace/trace_test.go
@@ -232,27 +232,27 @@ func TestStartSpanWithRemoteParent(t *testing.T) {
 
 func TestLocalRootSpanID(t *testing.T) {
 	ctx, span1 := StartSpan(context.Background(), "span1")
-	if span1.spanContext.LocalRootSpanID == (SpanID{}) {
-		t.Errorf("exporting root span: expected nonzero LocalRootSpanID")
+	if span1.localRootSpanID == (SpanID{}) {
+		t.Errorf("exporting root span: expected nonzero localRootSpanID")
 	}
 
 	_, span2 := StartSpan(ctx, "span2")
 	if err := checkChild(span1.spanContext, span2); err != nil {
 		t.Error(err)
 	}
-	if got, want := span2.spanContext.LocalRootSpanID, span2.spanContext.LocalRootSpanID; got != want {
-		t.Errorf("span2.LocalRootSpanID=%q; want %q (span1.LocalRootSpanID)", got, want)
+	if got, want := span2.localRootSpanID, span2.localRootSpanID; got != want {
+		t.Errorf("span2.localRootSpanID=%q; want %q (span1.localRootSpanID)", got, want)
 	}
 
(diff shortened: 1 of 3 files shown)


trace/trace_test.go
```diff
@@ -230,6 +230,32 @@ func TestStartSpanWithRemoteParent(t *testing.T) {
...
+	ctx, span1 := StartSpan(context.Background(), "span1")
+	if span1.localRootSpanID == (SpanID{}) {
+		t.Errorf("exporting root span: expected nonzero localRootSpanID")
+	}
+
+	_, span2 := StartSpan(ctx, "span2")
+	if err := checkChild(span1.spanContext, span2); err != nil {
+		t.Error(err)
+	}
+	if got, want := span2.localRootSpanID, span2.localRootSpanID; got != want {
```
rghetia: want should be span1.localRootSpanID.

trace/trace_test.go
```diff
@@ -230,6 +230,32 @@ func TestStartSpanWithRemoteParent(t *testing.T) {
...
+
+	_, span3 := StartSpanWithRemoteParent(context.Background(), "span3", span2.SpanContext())
+	if err := checkChild(span3.spanContext, span2); err != nil {
+		t.Error(err)
+	}
+	if span3.localRootSpanID == (SpanID{}) {
+		t.Errorf("exporting span with remote parent: expected nonzero localRootSpanID")
+	}
+	if got, want := span3.localRootSpanID, span2.localRootSpanID; got == want {
+		t.Errorf("span2.localRootSpanID=%q; expected different value to span1.localRootSpanID, got same", got)
```
rghetia: Error message should refer to span3 (got) and span2 (different from).

trace/trace_test.go
axw: Thanks - fixed. I rushed the tests a little bit.

(The tests aren't too important though, I mostly just put this up for comments on the approach.)

trace/trace_test.go
axw: Thanks, fixed.

axw:
Thanks @rghetia.

I hadn't intended for this to be merged before https://github.com/census-instrumentation/opencensus-specs/issues/229 was resolved, but if you're happy for this to go in then I'm happy.

rghetia:
> Thanks @rghetia.
> 
> I hadn't intended for this to be merged before [census-instrumentation/opencensus-specs#229](https://github.com/census-instrumentation/opencensus-specs/issues/229) was resolved, but if you're happy for this to go in then I'm happy.

It looks fine to me however I should not have merged until the spec is modified first. I'll roll it back unless we decide to update the spec based on this PR.

tsloughter:
Is this being added to the actual proto/spec? Right now it means the opencensus go deviates from the other implementations and they won't be able to work elastic.

PR #1032, opened 2019-02-21
Add support for tag metadata.

rghetia:


rghetia:
This should be merged only after [spec](https://github.com/census-instrumentation/opencensus-specs/pull/233) is updated.

tag/map.go
```diff
@@ -126,13 +174,35 @@ func Update(k Key, v string) Mutator {
...
+	}
+}
+
+// UpsertWithMetadata returns a mutator that upserts the
+// value of the tag associated with k with v. It also updates
+// Metadata md. It inserts the value and the metadata if k
+// doesn't exist already. It mutates the value and the
+// Metadata if k already exists.
+// If metadata md is nil then it defaults to NonPropagatingMetadata
+func UpsertWithMetadata(k Key, v string, md *Metadata) Mutator {
```
bogdandrutu: Maybe consider to use the same pattern as StartOption from StartSpan in trace.

```go
type metadatas struct {
	ttl int
}

type Metadata func(*metadatas)

// Not exported for the moment because we want only TTLNoPropagation and TTLUnlimitedPropagation
func withTTL(ttl int) Metadata {
  return func(*metadatas) {
    
  }
}

var (
  TTLNoPropagation = withTTL(valueTTLNoPropagation)
  TTLUnlimitedPropagation = withTTL(valueTTLUnlimitedPropagation)
}

func Upsert(k Key, v string, mds ...Metadata) Mutator {
    metas metadatas
    for _, md := range mds {
        op(&metas)
    }
}
```

Users will use this:
```go
tag.Upsert(key, "my_value", TTLNoPropagation, other_metadata)
```

What do you think? @rakyll do you have any suggestion?

tag/map.go
```diff
@@ -62,29 +67,29 @@ func (m *Map) String() string {
 	return buffer.String()
 }
 
-func (m *Map) insert(k Key, v string) {
+func (m *Map) insert(k Key, v string, md *Metadata) {
```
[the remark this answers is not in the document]
rghetia: metadata may or may not exists or given kv pair. Also metadata could be reused for several kv pair, In that case it can be reused.

tag/map.go
```diff
@@ -126,13 +174,35 @@ func Update(k Key, v string) Mutator {
...
 				return nil, errInvalidValue
 			}
-			m.upsert(k, v)
+			m.upsert(k, v, PropagatingMetadata)
+			return m, nil
+		},
+	}
+}
+
+// UpsertWithMetadata returns a mutator that upserts the
```
rakyll: UpsertWithMetadata is a very verbose name for a core API.

Please consider short names. If not, user will never think about UpsertWithMetadata is a replacement for Upsert.

tag/metadata.go
```diff
@@ -0,0 +1,37 @@
+// Copyright 2017, OpenCensus Authors
```
songy23: nit: s/2017/2019

tag/metadata.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// Metadata represents a tag Metadata.
+type Metadata struct {
+	ttl int
+}
+
+var (
+	// NonPropagatingMetadata has non propagating property. It is predefined for convenience.
+	NonPropagatingMetadata = &Metadata{}
```
songy23: Should this be `NonPropagatingMetadata = &Metadata{TagTTLNoPropagation}`?

rghetia:
@bogdandrutu, @songy23 PTAL.

tag/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+func withTTL(ttl int) Metadata {
+	return func(m *metadatas) {
+		m.ttl = ttl
+	}
+}
+
+var (
+	// TTLNoPropagation applies metadata with ttl value of valueTTLNoPropagation.
+	// It is predefined for convenience.
+	TTLNoPropagation = withTTL(valueTTLNoPropagation)
```
songy23: Not sure if we want to expose these constants here. See the comment at https://github.com/census-instrumentation/opencensus-java/pull/1768#discussion_r264796776.

tag/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+	// valueTTLUnlimitedPropagation allows tag to propagate without any limits on number of hops.
+	valueTTLUnlimitedPropagation = -1
+)
+
+type metadatas struct {
+	ttl int
+}
+
+// Metadata applies metadatas specified by the function.
+type Metadata func(*metadatas)
```
songy23: Curious why this needs to be `func`?

tag/metadata.go
rghetia: The intention was to restrict from using any value other than two TTL values. But this is not the right approach as TTLNoPropation can be modified by the user. Let me look at how to do this differently.

tag/metadata.go
rghetia: sort of like a builder (similar to startOption for startSpan).

tag/metadata.go
```diff
@@ -0,0 +1,47 @@
...
+
+// WithTTLNoPropagation applies metadata with ttl value of valueTTLNoPropagation.
+// It is predefined for convenience.
+func WithTTLNoPropagation() Metadata {
```
bogdandrutu: What do you think about this?
```go
type TTL struct {
  value int
}

var (
NoPropagation = &TTL{value= 0}
UnlimitedPropagation = &TTL{value= -1}
)

func WithTTL(TTL ttl) Metadata {
	return func(m *metadatas) {
		m.ttl = ttl
	}
}
     

```

rghetia: you could inadvertently modify UnlimitedPropagation and WithTTL would allow any value of TTL. I thought we don't want that at this point

rghetia: @rakyll would you be worried if someone by mistake sets UnlimitedPropagation to nil?

bogdandrutu: Let's do it this way.

rghetia: @bogdandrutu I have already made the change and uploaded. If everything looks fine then can you please approve it?

rghetia:
@bogdandrutu , friendly reminder.

bogdandrutu:
Why closed? Do I still need to look at this?

rghetia:
closed it by mistake.  I'll reopen it.

rghetia:
> Why closed? Do I still need to look at this?

I have made all the changes you requested. You can approve if it looks good.

rghetia:
@bogdandrutu friendly reminder.

songy23:
Please post a new PR to the `deb` branch. This one is to `master`.

rghetia:
opening a new PR on dev.

PR #1036, opened 2019-02-26
Use Host from Request instead of Request.URL

### LINKED ISSUE #992: blank HTTP host variable is logged
I made a similar ticket on the stackdriver code but this is happening here, as well. https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/issues/62

Blank http.host values are being sent because this exporter uses the r.URL.Host variable (that is often empty) and doesn't log the r.Host (which is often provided) on http.Request objects.

The reason why is described in this stackoverflow answer: https://stackoverflow.com/questions/42921567/what-is-the-difference-between-host-and-url-host-for-golang-http-request

I've not found a case in which the r.URL.Host was the desired one since HTTP/1.1 and newer (which require Host headers) is standard now.

philhug:
Request.URL contains the value sent in the HTTP request and not in the Host header.

This fixes #992


bogdandrutu:
Please rebase.

hug-abacus:
rebased

PR #1037, opened 2019-02-26
treat 0 bucket bound as an error.

rghetia:
Enforcing [spec](https://github.com/census-instrumentation/opencensus-proto/blob/c9cc5de7251f8a320dd6cb4f92e5698ba86f284c/src/opencensus/proto/metrics/v1/metrics.proto#L222-L225)

### DIFF
--- a/stats/view/view.go
+++ b/stats/view/view.go
@@ -70,7 +70,7 @@ func (v *View) same(other *View) bool {
 		v.Measure.Name() == other.Measure.Name()
 }
 
-var ErrNegativeBucketBounds = errors.New("negative bucket bounds not supported")
+var ErrNonPositiveBucketBounds = errors.New("non-positive bucket bounds not supported")
 
 // canonicalize canonicalizes v by setting explicit
 // defaults for Name and Description and sorting the TagKeys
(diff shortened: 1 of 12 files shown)


stats/view/view.go
```diff
@@ -95,7 +95,7 @@ func (v *View) canonicalize() error {
 	})
 	sort.Float64s(v.Aggregation.Buckets)
 	for _, b := range v.Aggregation.Buckets {
-		if b < 0 {
+		if b <= 0 {
 			return ErrNegativeBucketBounds
```
songy23: This should be renamed "ErrNonPositiveBucketBounds" now.

rghetia: done.

rghetia: I changed it back as I am silently dropping 0 bound bucket.

songy23: OK, silently dropping it makes more sense.

stats/view/view_test.go
```diff
@@ -459,18 +459,6 @@ func TestViewRegister_negativeBucketBounds(t *testing.T) {
 	}
 }
 
-func TestViewRegister_zeroBucketBounds(t *testing.T) {
```
songy23: Consider adding a test on verifying the leading 0 bucket bounds are dropped.

rghetia: added the test.

odeke-em:
Oh oh this is a little concerning and going to be a breaking change for many people and all our libraries, examples and integrations too. For more than a year we've had buckets of `[0, 10` etc acceptable. Enforcing a spec is great but should be gradual IMHO.

rghetia:
@odeke-em it will still accept 0 value bucket but will silently drop. So it will not break the api.

PR #1039, opened 2019-02-28
add metric producer manager.

rghetia:


metric/producer/manager.go
```diff
@@ -0,0 +1,80 @@
...
+
+package producer
+
+import (
+	"sync"
+)
+
+type manager struct {
+	mu       sync.RWMutex
+	producers []Producer
```
songy23: Consider using a `map[string]Producer`.

metric/producer/manager.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// GetAll returns all producer registered with the manager. It is typically
+// used by exporter to read metrics from producers.
+func GetAll() []Producer{
+	pm := getManager()
+	return pm.producers
+}
+
+func (pm *manager) contains(producer Producer) bool {
```
songy23: Suggestion: if `contains` is only used in `add`, maybe it's not worth adding a separate method; otherwise `contains` should be guarded with `mu`.

metric/producer/manager.go
```diff
@@ -0,0 +1,80 @@
...
+func Delete(producer Producer) {
+	pm := getManager()
+	pm.delete(producer)
+}
+
+// GetAll returns all producer registered with the manager. It is typically
+// used by exporter to read metrics from producers.
+func GetAll() []Producer{
+	pm := getManager()
+	return pm.producers
```
songy23: Consider returning a copy of producer slice, otherwise this may introduce synchronization issue. In Java we always return [an unmodifiable snapshot](https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/metrics/export/MetricProducerManagerImpl.java#L32).

metric/producer/manager.go
rghetia: i had considered that before but map cannot be unmodifiable. However slices are unmodifiable implicitly.

metric/producer/manager.go
rghetia: slices are unmodifiable implicitly. I have added a test for it.

metric/producer/manager.go
rghetia: removed contains.

metric/producer/manager.go
bogdandrutu: You can implement your own immutable map/slice (also not sure that slices are immutable) by cloning the current map/slice and apply the mutation operation on the clone then swap the pointers to the map/slice.

metric/producer/manager.go
bogdandrutu: That is not true:
I think this is a good article to read https://medium.com/@nitishmalhotra/uh-ohs-in-go-slice-of-pointers-c0a30669feee

bogdandrutu: I gave you the wrong link. The idea is that Slice is a struct {len, cap, pointerToData}. If passed by value all of these will get copied but the underlying memory where  pointerToData points is shared between multiple copies of the slice.

rghetia: I got it. It is NOT immutable. I verified by a different test.

rghetia: I will change the implementation.

rghetia:
@bogdandrutu and @songy23 PTAL again.

metric/producer/manager.go
```diff
@@ -0,0 +1,87 @@
...
+
+func getManager() *manager {
+	once.Do(func() {
+		prodMgr = &manager{}
+		prodMgr.producers = make(map[Producer]struct{})
+	})
+	return prodMgr
+}
+
+// Add adds the producer to the manager if it is not already present.
```
bogdandrutu: Maybe define what is the "Manager"?

metric/producer/manager.go
```diff
@@ -0,0 +1,87 @@
...
+// the producers.
+func GetAll() []Producer {
+	pm := getManager()
+	return pm.getAll()
+}
+
+func (pm *manager) getAll() []Producer {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+	producers := make([]Producer, len(pm.producers))
```
bogdandrutu: producers := make([]Producer, len(pm.producers))
copy(producers, pm.producers)

https://github.com/golang/go/wiki/SliceTricks#copy

metric/metricexport/producer.go
```diff
@@ -19,6 +19,7 @@ import (
 )
 
 // Producer is a source of metrics.
+// Deprecated in lieu of go.opencensus.io.metric.producer.Producer
```
bogdandrutu: Simply delete it, we don't use this anywhere correct?

bogdandrutu:
@rghetia ping me when this is ready for review.

metric/metricexport/producer.go
rghetia: deleted.

metric/producer/manager.go
rghetia: done.

metric/producer/manager.go
rghetia: producers is slice while pm.producers is a map. So copy won't work.

rghetia:
> @rghetia ping me when this is ready for review.

It is ready.

metric/producer/manager.go
```diff
@@ -0,0 +1,89 @@
...
+		prodMgr = &manager{}
+		prodMgr.producers = make(map[Producer]struct{})
+	})
+	return prodMgr
+}
+
+// Add adds the producer to the manager if it is not already present.
+// The manager maintains the list of active producers. It provides
+// this list to a reader to read metrics from each producer and then export.
+func Add(producer Producer) {
```
bogdandrutu: I would like to have these function on a "ProducerManager" struct and have a "GetGlobalProducerManager". This way the exporters will get a "ProducerManager" as an option, this is a bit more flexible. We can do that in a separate PR just to make sure we make progress, but I think is a better design.

rghetia: sure.

odeke-em:
Hey folks, thanks for the work on this @rghetia! However, I am a little lost on what these interfaces are for? Including #1045 this seems very Java-esque. What's going on?

songy23:
@odeke-em This is for the OpenCensus-Go main library to produce `Metric`s. Before we produce `ViewData` and `ViewData` is going to be replaced by `Metric`s.

PR #1056, opened 2019-03-11
refactor ReadAll to Read and fix concurrency issue.

rghetia:


metric/registry.go
```diff
@@ -15,15 +15,18 @@
...
+
+	"go.opencensus.io/metric/metricdata"
 )
 
 // Registry creates and manages a set of gauges.
 // External synchronization is required if you want to add gauges to the same
 // registry from multiple goroutines.
 type Registry struct {
+	mu     sync.RWMutex
 	gauges map[string]*gauge
```
bogdandrutu: Maybe consider to use sync.Map? Please let me know if that matches your requirements.

rghetia: sync.Map would work.

metric/registry.go
```diff
@@ -53,8 +53,9 @@ func (r *Registry) AddInt64Gauge(name, description string, unit metricdata.Unit,
 }
 
 func (r *Registry) initGauge(g *gauge, labelKeys []string, name string, description string, unit metricdata.Unit) *gauge {
-	existing, ok := r.gauges[name]
+	val, ok := r.gauges.Load(name)
 	if ok {
+		existing := val.(*gauge)
 		if existing.isFloat != g.isFloat {
```
bogdandrutu: As we discussed we need to return an error if it exists not only if it has a different type. But happy to see that in a different PR.

metric/registry.go
```diff
@@ -67,15 +68,17 @@ func (r *Registry) initGauge(g *gauge, labelKeys []string, name string, descript
...
+	r.gauges.Store(name, g)
 	return g
 }
 
-// ReadAll reads all gauges in this registry and returns their values as metrics.
-func (r *Registry) ReadAll() []*metricdata.Metric {
-	ms := make([]*metricdata.Metric, 0, len(r.gauges))
-	for _, g := range r.gauges {
+// Read reads all gauges in this registry and returns their values as metrics.
+func (r *Registry) Read() []*metricdata.Metric {
```
bogdandrutu: Why change this name? I don't feel strongly about it but was curios.

metric/registry.go
rghetia: I will create a different PR for this.

metric/registry.go
rghetia: Producer [interface](https://github.com/census-instrumentation/opencensus-go/blob/2eaaf3ad5aeaf6d5f513040ace947fad79d9da63/metric/metricproducer/producer.go#L27) is Read(). Since there is only going to be one Read there is no point in naming ReadAll. (short is better as long as it is clear). 
If ReadAll is preferable I can change the producer interface.

PR #1057, opened 2019-03-12
During gauge creation return error instead of panic.

rghetia:


metric/gauge.go
```diff
@@ -24,19 +25,19 @@ import (
 	"go.opencensus.io/metric/metricdata"
 )
 
-// gauge represents a quantity that can go up an down, for example queue depth
+// gauge represents a quantity that can go up an down, for example queueInt64 depth
```
songy23: > queueInt64

Typo?

metric/error_const.go
```diff
@@ -0,0 +1,20 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metric
+
+const (
+	errGaugeExistsWithDiffType = "gauge with same name exists with a different type"
```
songy23: You can use error constants instead:
```go
var (
  errGaugeExistsWithDiffType = errors.New("gauge with same name exists with a different type")
 ...
)
```

rghetia: SG.

metric/gauge.go
rghetia: yes. I'll fix it.

Issue #649, opened 2018-03-28
Prometheus exporter produces non cumulative histogram buckets


rozaliev:
I did some testing with opencensus-go and prometheus library, it seems that opencensus produces non cumulative buckets. 

How to reproduce:
```
exporter, err := prometheus.NewExporter(prometheus.Options{})
if err != nil {
   log.Fatal(err)
}
view.RegisterExporter(exporter)
if err := view.Subscribe(ochttp.DefaultServerViews...); err != nil {
   log.Fatal(err)
}
```

Expected result 
```
opencensus_opencensus_io_http_server_latency_bucket{le="0"} 0
opencensus_opencensus_io_http_server_latency_bucket{le="1"} 1
opencensus_opencensus_io_http_server_latency_bucket{le="2"} 1
opencensus_opencensus_io_http_server_latency_bucket{le="3"} 1
opencensus_opencensus_io_http_server_latency_bucket{le="4"} 1
opencensus_opencensus_io_http_server_latency_bucket{le="5"} 1
opencensus_opencensus_io_http_server_latency_bucket{le="6"} 1
opencensus_opencensus_io_http_server_latency_bucket{le="8"} 1
opencensus_opencensus_io_http_server_latency_bucket{le="10"} 1
opencensus_opencensus_io_http_server_latency_bucket{le="13"} 939126
opencensus_opencensus_io_http_server_latency_bucket{le="16"} 974422
opencensus_opencensus_io_http_server_latency_bucket{le="20"} 974609
opencensus_opencensus_io_http_server_latency_bucket{le="25"} 974609
opencensus_opencensus_io_http_server_latency_bucket{le="30"} 974609
opencensus_opencensus_io_http_server_latency_bucket{le="40"} 974609
opencensus_opencensus_io_http_server_latency_bucket{le="50"} 974609
opencensus_opencensus_io_http_server_latency_bucket{le="65"} 974609
opencensus_opencensus_io_http_server_latency_bucket{le="80"} 974609
opencensus_opencensus_io_http_server_latency_bucket{le="100"} 974609
opencensus_opencensus_io_http_server_latency_bucket{le="130"} 974611
opencensus_opencensus_io_http_server_latency_bucket{le="160"} 974611
opencensus_opencensus_io_http_server_latency_bucket{le="200"} 974611
opencensus_opencensus_io_http_server_latency_bucket{le="250"} 974611
opencensus_opencensus_io_http_server_latency_bucket{le="300"} 974611
[...]

PR #1061, opened 2019-03-12
Exporter/Prometheus: Simplify histogram creation.

songy23:
And resolve a TODO.

With https://github.com/census-instrumentation/opencensus-go/pull/979 and https://github.com/census-instrumentation/opencensus-go/pull/1037, bucket bounds are guaranteed to be positive and strictly increasing.

### DIFF
--- a/exporter/prometheus/prometheus_test.go
+++ b/exporter/prometheus/prometheus_test.go
@@ -288,7 +288,7 @@ func TestCumulativenessFromHistograms(t *testing.T) {
 	// We want the results that look like this:
 	// 1:   [0.25]      		| 1 + prev(i) = 1 + 0 = 1
 	// 5:   [1.45]			| 1 + prev(i) = 1 + 1 = 2
-	// 10:	[]			| 1 + prev(i) = 1 + 2 = 3
+	// 10:	[7.69]			| 1 + prev(i) = 1 + 2 = 3
 	// 20:  [12]			| 1 + prev(i) = 1 + 3 = 4
 	// 50:  []			| 0 + prev(i) = 0 + 4 = 4
 	// 100: []			| 0 + prev(i) = 0 + 4 = 4
@@ -358,9 +358,9 @@ func TestHistogramUnorderedBucketBounds(t *testing.T) {
 		Description: "this is a test",
 		Measure:     m,
 
-		// Intentionally used unordered elements in the ascending distribution
+		// Intentionally used unordered and duplicated elements in the distribution
 		// to ensure unordered bucket bounds are handled.
-		Aggregation: view.Distribution(10, 5, 1, 50, 20, 100, 250),
+		Aggregation: view.Distribution(10, 5, 1, 1, 50, 5, 20, 100, 250),
 	}
 
 	if err := view.Register(v); err != nil {
@@ -375,7 +375,7 @@ func TestHistogramUnorderedBucketBounds(t *testing.T) {
 	// We want the results that look like this:
 	// 1:   [0.25]      		| 1 + prev(i) = 1 + 0 = 1
 	// 5:   [1.45]			| 1 + prev(i) = 1 + 1 = 2
-	// 10:	[]			| 1 + prev(i) = 1 + 2 = 3
+	// 10:	[7.69]			| 1 + prev(i) = 1 + 2 = 3
 	// 20:  [12]			| 1 + prev(i) = 1 + 3 = 4
 	// 50:  []			| 0 + prev(i) = 0 + 4 = 4
 	// 100: []			| 0 + prev(i) = 0 + 4 = 4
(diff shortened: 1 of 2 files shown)


odeke-em:
If a user defines `v := view.Distribution(0, 10, 9, 20, 40)` what does OpenCensus-Go do? Currently nothing, which wouldn't mitigate the problem at core. We should at that call site panic but we don't. Those changes in https://github.com/census-instrumentation/opencensus-go/pull/979 and https://github.com/census-instrumentation/opencensus-go/pull/1037 enforce no negative bounds but really this code is about monotonic arrangement and sorting which is totally different from the assumption of this PR, thus my objection to not merge.

songy23:
AFAIS https://github.com/census-instrumentation/opencensus-go/issues/649 is a different issue where we didn't use cumulative counts for histogram in Prometheus. This PR doesn't change that part.

For the case when bucket bounds are out of order, we should fail fast when creating the distribution aggregation. This is the case in Java: https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/stats/BucketBoundaries.java#L52-L58. I'll add a PR in Go that also checks this.

songy23:
> If a user defines v := view.Distribution(0, 10, 9, 20, 40) what does OpenCensus-Go do?

This should return an error. This is already specified in the [proto](https://github.com/census-instrumentation/opencensus-proto/blob/c9cc5de7251f8a320dd6cb4f92e5698ba86f284c/src/opencensus/proto/metrics/v1/metrics.proto#L223) and is also a requirement for GA. /cc @mayurkale22

rghetia:
it sorts the buckets [here](https://github.com/census-instrumentation/opencensus-go/blob/2eaaf3ad5aeaf6d5f513040ace947fad79d9da63/stats/view/view.go#L98)

songy23:
> it sorts the buckets here

I see, then we can always guarantee the bucket bounds are ordered. I think merging this PR won't cause any problem then.

rghetia:
> > it sorts the buckets here
> 
> I see, then we can always guarantee the bucket bounds are ordered. I think merging this PR won't cause any problem then.

I don't think it should create any problem.

odeke-em:
Alright, cool then. But let's add a test to ensure that it doesn't trip things out. I ask because the PR https://github.com/census-instrumentation/opencensus-go/pull/979 that adding sorting (4 months ago) didn't add a test for that check hence the assumptions that it won't break.

bogdandrutu:
+1 for having a test

songy23:
Added a test in https://github.com/census-instrumentation/opencensus-go/pull/1061/commits/1f913338f90ff426e3889abba78bd9ba4785f4c1, PTAL.

exporter/prometheus/prometheus_test.go
```diff
@@ -343,6 +343,93 @@ func TestCumulativenessFromHistograms(t *testing.T) {
...
+	defer view.Unregister(v)
+
+	// Give the reporter ample time to process registration
+	<-time.After(10 * reportPeriod)
+
+	values := []float64{0.25, 245.67, 12, 1.45, 199.9, 7.69, 187.12}
+	// We want the results that look like this:
+	// 1:   [0.25]      		| 1 + prev(i) = 1 + 0 = 1
+	// 5:   [1.45]			| 1 + prev(i) = 1 + 1 = 2
+	// 10:	[]			| 1 + prev(i) = 1 + 2 = 3
```
odeke-em: There is a problem here, 7.69 has to be in 10 bucket. Or is that just a bad copy+paste?

exporter/prometheus/prometheus_test.go
```diff
@@ -343,6 +343,93 @@ func TestCumulativenessFromHistograms(t *testing.T) {
...
+
+	m := stats.Float64("tests/bills", "payments by denomination", stats.UnitDimensionless)
+	v := &view.View{
+		Name:        "cash/register",
+		Description: "this is a test",
+		Measure:     m,
+
+		// Intentionally used unordered elements in the ascending distribution
+		// to ensure unordered bucket bounds are handled.
+		Aggregation: view.Distribution(10, 5, 1, 50, 20, 100, 250),
```
odeke-em: Also make sure that we duplicate a bucket boundary in there as the previous code also deduplicated boundaries :)

exporter/prometheus/prometheus_test.go
songy23: I copied and pasted it from https://github.com/census-instrumentation/opencensus-go/blob/ed9f155754b50f84f15997a4550ebd54667078e7/exporter/prometheus/prometheus_test.go#L288-L295. Looks like the previous comments are wrong. Updated.

exporter/prometheus/prometheus_test.go
songy23: SG, done.

exporter/prometheus/prometheus_test.go
odeke-em: I think they got stale during the removal of 0 as the first bucket bound, but now this is cool. Thanks for fixing them!

PR #1068, opened 2019-03-15
Remove dependency from ourself, this happens because bundler in google.golang.org/api.

bogdandrutu:
I have no idea if this is the right thing to do, but we kind of have a module cycle dependency but not a package cycle dependency. If anyone knows if I am doing the right thing please confirm :)


bogdandrutu:
@dmitshur any idea about what if I am doing here is correct?

* go.opencensus.io dependend on google.golang.org/api just for the "support" package.
* google.golang.org/api depends in other packages on go.opencensus.io
* I added the replace with the idea that we have always a newer version.

dmitshur:
I'm not sure that this will actually work, for the following reason.

Replace directives of a go.mod file only apply when that module is the main module. When this module is required elsewhere, its replace directive has no effect.

Since the change here still has a requirement on `google.golang.org/api@v0.2.0`, which in turn requires `go.opencensus.io@v0.19.1`, and you're back to having a problem.

See my comment at https://github.com/googleapis/google-api-go-client/issues/342#issuecomment-473139030 for a possible way to get out of this situation.

songy23:
This has been fixed in https://github.com/census-instrumentation/opencensus-go/releases/tag/v0.20.1-alpha.

PR #1069, opened 2019-03-18
ochttp plugin: makes Body a transparent wrapper

GreenHorse:
With Go 1.12 the body of http.Response can implement `io.Writer`.
https://tip.golang.org/pkg/net/http/#Response.Body

Upgrading to websocket for example will result with a body implementing `io.Writer`.
https://golang.org/src/net/http/httputil/reverseproxy.go#L521

I followed the logic of wrappedResponseWriter, it might be overkill as Response.Body has only one optional interface for now.

Both `ochttp.tracker` and `ochttp.bodyTracker` are affected.

### DIFF
--- a/plugin/ochttp/client_stats.go
+++ b/plugin/ochttp/client_stats.go
@@ -68,7 +68,7 @@ func (t statsTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 			track.end()
 		} else {
 			track.body = resp.Body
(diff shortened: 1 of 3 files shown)


rghetia:
LGTM.
@GreenHorse can you please close and reopen so that appveyor build gets triggered again?

GreenHorse:
@rghetia checks have passed

PR #1071, opened 2019-03-19
Exemplar: Restore previous package to avoid breaking downstream.

songy23:
Only restored minimal files and marked them as deprecated.

### DIFF
--- a/exemplar/exemplar.go
+++ b/exemplar/exemplar.go
@@ -1,0 +1,83 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


exemplar/exemplar.go
```diff
@@ -0,0 +1,83 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package exemplar implements support for exemplars. Exemplars are additional
+// data associated with each measurement.
+//
+// Their purpose it to provide an example of the kind of thing
+// (request, RPC, trace span, etc.) that resulted in that measurement.
+//
+// Deprecated: Use go.opencensus.io/metric/metricdata instead.
```
songy23: Please import `go.opencensus.io/metric/metricdata` for migration.

rghetia:
Is the goal here to simply not break the downstream builds? None of the content is actually used in the core library.

songy23:
> Is the goal here to simply not break the downstream builds? None of the content is actually used in the core library.

Yes, if downstream always depends on mainline and imports "go.opencensus.io/exemplar", the previous PR will break their build. Per https://github.com/census-instrumentation/opencensus-go#deprecation-policy we should leave time for downstream to update their dependencies.

Issue #1058, opened 2019-03-12
Refactor Exemplar APIs.


songy23:
Reported offline by @bogdandrutu. Currently in Go `Exemplar` is a first-class package and it has direct dependency on Tracing APIs, which is not ideal. `Exemplar` should be under metrics data model instead and the generic interface shouldn't depend on Tracing.

songy23:
All PRs have been merged.

PR #1075, opened 2019-03-20
 Exemplar: Add new record APIs that take exemplar attachments and SpanContext key.

songy23:
Updates https://github.com/census-instrumentation/opencensus-go/issues/1058.

Existing record APIs will continue to work.

### DIFF
--- a/metric/metricdata/exemplar.go
+++ b/metric/metricdata/exemplar.go
@@ -18,6 +18,11 @@ import (
 	"time"
 )
 
+// Exemplars keys.
+const (
+	AttachmentKeySpanContext = "SpanContext"
+)
+
 // Exemplar is an example data point associated with each bucket of a
 // distribution type aggregation.
 //
--- a/stats/record.go
+++ b/stats/record.go
@@ -18,6 +18,7 @@ package stats
 import (
 	"context"
 
+	"go.opencensus.io/metric/metricdata"
 	"go.opencensus.io/stats/internal"
 	"go.opencensus.io/tag"
 )
@@ -38,7 +39,7 @@ func Record(ctx context.Context, ms ...Measurement) {
 
 // RecordWithAttachments records measurements and the given exemplar attachments against context.
 // If there are any tags in the context, measurements will be tagged with them.
-func RecordWithAttachments(ctx context.Context, attachments map[string]interface{}, ms ...Measurement) {
+func RecordWithAttachments(ctx context.Context, attachments metricdata.Attachments, ms ...Measurement) {
 	recorder := internal.DefaultRecorder
 	if recorder == nil {
 		return
@@ -73,7 +74,7 @@ func RecordWithTags(ctx context.Context, mutators []tag.Mutator, ms ...Measureme
 // Measurements will be tagged with the tags in the context mutated by the mutators.
 // RecordWithTags is useful if you want to record with tag mutations but don't want
 // to propagate the mutations in the context.
-func RecordWithTagsAndAttachments(ctx context.Context, mutators []tag.Mutator, attachments map[string]interface{}, ms ...Measurement) error {
+func RecordWithTagsAndAttachments(ctx context.Context, mutators []tag.Mutator, attachments metricdata.Attachments, ms ...Measurement) error {
 	ctx, err := tag.New(ctx, mutators...)
 	if err != nil {
 		return err
--- a/stats/record.go
+++ b/stats/record.go
@@ -34,7 +34,7 @@ func init() {
 // Record records one or multiple measurements with the same context at once.
 // If there are any tags in the context, measurements will be tagged with them.
 func Record(ctx context.Context, ms ...Measurement) {
-	RecordWithAttachments(ctx, map[string]interface{}{}, ms...)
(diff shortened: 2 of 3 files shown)


metric/metricdata/exemplar.go
```diff
@@ -18,6 +18,11 @@ import (
 	"time"
 )
 
+// Exemplars keys.
+const (
+	AttachmentKeySpanContext = "SpanContext"
```
rakyll: Is there a reason why this is public?

songy23: This key will be used when 1. record with span context attachment and 2. exporters extract span context from exemplar. If it's not exposed, users may record span context attachment with a random key that exporters don't expect.

In Java it's also public: https://github.com/census-instrumentation/opencensus-java/blob/70f9701a32dee5d2cd07c7d4a1eb82cb2c26b863/contrib/exemplar_util/src/main/java/io/opencensus/contrib/exemplar/util/ExemplarUtils.java#L36.

stats/record.go
```diff
@@ -33,6 +33,12 @@ func init() {
 // Record records one or multiple measurements with the same context at once.
 // If there are any tags in the context, measurements will be tagged with them.
 func Record(ctx context.Context, ms ...Measurement) {
+	RecordWithAttachments(ctx, map[string]interface{}{}, ms...)
+}
+
+// RecordWithAttachments records measurements and the given exemplar attachments against context.
+// If there are any tags in the context, measurements will be tagged with them.
+func RecordWithAttachments(ctx context.Context, attachments map[string]interface{}, ms ...Measurement) {
```
bogdandrutu: I would do something like

```go
type Attachments map[string]interface{}
```

Not sure if interface{} is the right type here, we need something that converts (allows user to convert to String).

songy23: Updated the API to take `metricdata.Attachments`. I think we can do type assertions with interface{} for the common types that we already knew, for example:
```go
for k, v := range(attachments) {
  if spanctx, ok := v.(SpanContext); ok {
    //...
  } else if str, ok := v.(string); ok {
    // ...
  } else {
    str := fmt.Sprintf("%v", v)
    // ...
  }
}
```
WDYT?

songy23: Chatted offline, we can instead make the exemplar attachment similar to [span attributes](https://github.com/census-instrumentation/opencensus-go/blob/master/trace/basetypes.go#L45-L70). Like:
```go


type Attachment {
	key   string
	value interface{}
}

func SpanContextAttachment(key string, value SpanContext) Attachment {
	return Attachment{key: key, value: value}
}

func StringAttachment(key string, value string) Attachment {
	return Attachment{key: key, value: value}
}
```

In this way we restricted the type of attachment we want. However at the same time we'll introduce tracing dependency to metrics. I'm not sure if this is the best practice in Go either. @rakyll any suggestions here?

rghetia: for non-string attachement is the expectation that they provide string form of the attachment?
If so then you could have
```go
type GetValue interface {
       ToString() string 
}

type Attachment struct {
	key   string
	value interface{}
}

func NonStringAttachment(key string, value GetValue) *Attachment {
	return &Attachment{key: key, value: value}
}

func StringAttachment(key, value string) *Attachment {
	return &Attachment{key: key, value: value}
}

func (a *Attachment) encodeAttachement() {
	switch v := a.value.(type) {
	case GetValue:
    	fmt.Printf("key: %s, value %s\n", a.key, v.ToString()); 
	case string:
		fmt.Printf("key: %s, value %s\n", a.key, v); 
	default:
	    panic("unsupported type")
	}
}
```

stats/record.go
```diff
@@ -33,6 +34,12 @@ func init() {
 // Record records one or multiple measurements with the same context at once.
 // If there are any tags in the context, measurements will be tagged with them.
 func Record(ctx context.Context, ms ...Measurement) {
+	RecordWithAttachments(ctx, map[string]interface{}{}, ms...)
```
bogdandrutu: why not nil?

stats/record.go
songy23: > for non-string attachement is the expectation that they provide string form of the attachment?

IMO no. In the exemplar attachment API we want to keep the value structs as-is, and only do the interpretation/formatting when exporting. For example we want to keep the whole `SpanContext` struct as the attachment value instead of only keeping a string representation of it. Only when we're about to export to Stackdriver, we get the trace and span id from span context and convert them into strings.

stats/record.go
songy23: Good catch, fixed.

stats/record.go
songy23: > Chatted offline, we can instead make the exemplar attachment similar to span attributes.
> In this way we restricted the type of attachment we want. 

Did some experiments on this, actually we won't be able to achieve this goal, because the key and value of exemplar attachment has to be public in order for other packages to access them. https://github.com/census-instrumentation/opencensus-go/pull/1085 implemented this idea, though I'm more inclined to keep the current API as-is.

songy23:
Going to create a new PR to the `dev` branch.

PR #1076, opened 2019-03-21
Exemplar: Record with sampled SpanContext in gRPC plugin.

### LINKED ISSUE #1058: Refactor Exemplar APIs.

songy23:
Fixes #1058.

Blocked by https://github.com/census-instrumentation/opencensus-go/pull/1075.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,21 +1,26 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-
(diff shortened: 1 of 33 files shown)


songy23:
Superseded by https://github.com/census-instrumentation/opencensus-go/pull/1127.

PR #1078, opened 2019-03-22
stats worker as metric producer.

rghetia:
This change provides stats data (view data model) as metric data. Stats worker registers as a metric producer with global producer manager. Read() function is then invoked by Readers.

### DIFF
--- a/stats/view/aggregation_data.go
+++ b/stats/view/aggregation_data.go
@@ -30,7 +30,7 @@ type AggregationData interface {
 	addSample(v float64, attachments map[string]interface{}, t time.Time)
 	clone() AggregationData
 	equal(other AggregationData) bool
-	toPoint(time time.Time) metricdata.Point
+	toPoint(t metricdata.Type, time time.Time) metricdata.Point
 }
 
 const epsilon = 1e-9
@@ -62,8 +62,15 @@ func (a *CountData) equal(other AggregationData) bool {
 	return a.Value == a2.Value
 }
 
-func (a *CountData) toPoint(t time.Time) metricdata.Point {
-	return metricdata.NewInt64Point(t, a.Value)
+func (a *CountData) toPoint(metricType metricdata.Type, t time.Time) metricdata.Point {
+	switch metricType {
+	case metricdata.TypeCumulativeInt64:
+		return metricdata.NewInt64Point(t, a.Value)
+	case metricdata.TypeCumulativeFloat64:
+		return metricdata.NewFloat64Point(t, float64(a.Value))
+	default:
+		panic("unsupported metricdata.Type")
+	}
 }
 
 // SumData is the aggregated data for the Sum aggregation.
@@ -92,8 +99,15 @@ func (a *SumData) equal(other AggregationData) bool {
 	return math.Pow(a.Value-a2.Value, 2) < epsilon
 }
 
-func (a *SumData) toPoint(t time.Time) metricdata.Point {
-	return metricdata.NewFloat64Point(t, a.Value)
+func (a *SumData) toPoint(metricType metricdata.Type, t time.Time) metricdata.Point {
+	switch metricType {
+	case metricdata.TypeCumulativeInt64:
+		return metricdata.NewInt64Point(t, int64(a.Value))
+	case metricdata.TypeCumulativeFloat64:
+		return metricdata.NewFloat64Point(t, a.Value)
+	default:
+		panic("unsupported metricdata.Type")
+	}
 }
 
 // DistributionData is the aggregated data for the
@@ -217,7 +231,7 @@ func (a *DistributionData) equal(other AggregationData) bool {
 	return a.Count == a2.Count && a.Min == a2.Min && a.Max == a2.Max && math.Pow(a.Mean-a2.Mean, 2) < epsilon && math.Pow(a.variance()-a2.variance(), 2) < epsilon
 }
 
-func (a *DistributionData) toPoint(t time.Time) metricdata.Point {
+func (a *DistributionData) toPoint(metricType metricdata.Type, t time.Time) metricdata.Point {
 	buckets := []metricdata.Bucket{}
 	for i := 0; i < len(a.CountPerBucket); i++ {
 		buckets = append(buckets, metricdata.Bucket{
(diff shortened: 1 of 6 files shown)


stats/view/view_to_metric.go
```diff
@@ -0,0 +1,131 @@
...
+	labelValues := []metricdata.LabelValue{}
+	for _, tag := range row.Tags {
+		labelValues = append(labelValues, metricdata.NewLabelValue(tag.Value))
+	}
+	return labelValues
+}
+
+func rowToTimeseries(row *Row, now time.Time, startTime time.Time) *metricdata.TimeSeries {
+	return &metricdata.TimeSeries{
+		Points:      []metricdata.Point{row.Data.toPoint(now)},
```
songy23: You need to convert metric types here. For example `SumData` is always converted to `Float64` point: https://github.com/census-instrumentation/opencensus-go/pull/1078/files#diff-a282032fe900298505848fd4d08399a2R96. But when measure is of type `Int64Measure`, metric type will be `CumulativeInt64`: https://github.com/census-instrumentation/opencensus-go/pull/1078/files#diff-0ac84b9d221fc10ebe99482beb04a4fbR45. This will be an error since the type of point doesn't match with metric type.

rghetia: good catch.

rghetia: fixed. 
comparing  serialized version made the test pass incorrectly. Removed the serialized comparision.

rghetia:
@songy23 PTAL.

stats/view/aggregation_data.go
```diff
@@ -61,6 +62,17 @@ func (a *CountData) equal(other AggregationData) bool {
 	return a.Value == a2.Value
 }
 
+func (a *CountData) toPoint(metricType metricdata.Type, t time.Time) metricdata.Point {
+	switch metricType {
+	case metricdata.TypeCumulativeInt64:
+		return metricdata.NewInt64Point(t, a.Value)
+	case metricdata.TypeCumulativeFloat64:
+		return metricdata.NewFloat64Point(t, float64(a.Value))
```
songy23: `Count` should always be `Int64`.

stats/view/view_to_metric.go
```diff
@@ -0,0 +1,131 @@
...
+func getUnit(unit string) metricdata.Unit {
+	switch unit {
+	case "1":
+		return metricdata.UnitDimensionless
+	case "ms":
+		return metricdata.UnitMilliseconds
+	case "By":
+		return metricdata.UnitBytes
+	}
+	return metricdata.UnitDimensionless
```
songy23: Why convert all other units to "1"?

stats/view/view_to_metric.go
```diff
@@ -0,0 +1,131 @@
...
+		switch m.(type) {
+		case *stats.Int64Measure:
+			return metricdata.TypeGaugeInt64
+		case *stats.Float64Measure:
+			return metricdata.TypeGaugeFloat64
+		default:
+			panic("unexpected measure type")
+		}
+	case AggTypeCount:
+		switch m.(type) {
```
songy23: Same here, `count` should always be `Int64`.

stats/view/view_to_metric.go
```diff
@@ -0,0 +1,131 @@
...
+		Unit:        getUnit(v.Measure.Unit()),
+		Type:        getType(v),
+		LabelKeys:   getLableKeys(v),
+	}
+}
+
+func toLabelValues(row *Row) []metricdata.LabelValue {
+	labelValues := []metricdata.LabelValue{}
+	for _, tag := range row.Tags {
+		labelValues = append(labelValues, metricdata.NewLabelValue(tag.Value))
```
songy23: Can `tag.Value` be `nil`? In Java it's possible, I cannot remember for Go. Maybe check for `nil` here.

stats/view/view_to_metric.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return labelValues
+}
+
+func rowToTimeseries(v *viewInternal, row *Row, now time.Time, startTime time.Time) *metricdata.TimeSeries {
+	return &metricdata.TimeSeries{
+		Points:      []metricdata.Point{row.Data.toPoint(v.metricDescriptor.Type, now)},
+		LabelValues: toLabelValues(row),
+		StartTime:   startTime,
```
songy23: For `Gauge`s this field doesn't need to be set.

stats/view/view_to_metric.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func viewToMetric(v *viewInternal, now time.Time, startTime time.Time) *metricdata.Metric {
+	if v.metricDescriptor.Type == metricdata.TypeGaugeInt64 ||
+		v.metricDescriptor.Type == metricdata.TypeGaugeFloat64 {
+		startTime = time.Time{}
```
songy23: Maybe use `nil` here and don't set `nil` `startTime`.

stats/view/view_to_metric.go
rghetia: I'll open a new PR to fix this.

stats/view/view_to_metric.go
rghetia: while LabelValue has a concept of
1. non-empty string
2. empty string (go nil)
3. not set (equivalent of java nil)

Tag itself doesn't have that concept. So it is impossible to determine if it #2 or #3 when tag.Value is empty string.

stats/view/aggregation_data.go
rghetia: fixed.

stats/view/view_to_metric.go
rghetia: fixed.

stats/view/view_to_metric.go
rghetia: not setting startime is same as time.Time{}.  StackdriverExporter sets it to nil in its conversion for gauges.

stats/view/view_to_metric.go
rghetia: see above.

rghetia:
@songy23 PTAL

stats/view/aggregation_data.go
```diff
@@ -208,6 +229,26 @@ func (a *DistributionData) equal(other AggregationData) bool {
 	return a.Count == a2.Count && a.Min == a2.Min && a.Max == a2.Max && math.Pow(a.Mean-a2.Mean, 2) < epsilon && math.Pow(a.variance()-a2.variance(), 2) < epsilon
 }
 
+func (a *DistributionData) toPoint(metricType metricdata.Type, t time.Time) metricdata.Point {
+	buckets := []metricdata.Bucket{}
```
songy23: nit: you can also check `metricType` here (expect to be cumulative distribution).

stats/view/view_to_metric_test.go
```diff
@@ -0,0 +1,424 @@
...
+				v = float64(i)
+			default:
+				t.Errorf("unexpected value type %v", r.tags)
+			}
+			tc.vi.addSample(tag.FromContext(ctx), v, nil, now)
+		}
+
+		gotMetric := viewToMetric(tc.vi, now, startTime)
+		if !reflect.DeepEqual(gotMetric, tc.wantMetric) {
+			// JSON format is strictly for checking the content when test fails. Do not use JSON
```
songy23: Consider using https://github.com/google/go-cmp.

PR #1085, opened 2019-03-29
[DO NOT MERGE] Exemplar: Refactor Attachment to be similar to Attribute.

songy23:
Created this PR for API discussion first.

Updated the exemplar APIs per https://github.com/census-instrumentation/opencensus-go/pull/1075#discussion_r268331566. Make the exemplar attachment more similar to trace attribute.

### DIFF
--- a/metric/metricdata/exemplar.go
+++ b/metric/metricdata/exemplar.go
@@ -16,6 +16,13 @@ package metricdata
 
 import (
 	"time"
+
+	"go.opencensus.io/trace"
+)
+
+// Exemplars keys.
+const (
+	AttachmentKeySpanContext = "SpanContext"
 )
 
 // Exemplar is an example data point associated with each bucket of a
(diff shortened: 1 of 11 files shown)


metric/metricdata/exemplar.go
```diff
@@ -24,10 +31,23 @@ import (
...
-	Timestamp   time.Time   // the time the value was recorded
-	Attachments Attachments // attachments (if any)
+	Value       float64                // the value that was recorded
+	Timestamp   time.Time              // the time the value was recorded
+	Attachments map[string]interface{} // attachments (if any)
+}
+
+// Attachment is a key-value pair associated with a recorded example data point.
+type Attachment struct {
+	Key   string
```
songy23: `Key` and `Value` have to be exposed because other packages (stats, exporters) need to access them.

PR #1086, opened 2019-03-29
Refactor gauge api with options.

rghetia:


metric/registry.go
```diff
@@ -37,33 +37,63 @@ const (
 	derivedGaugeFloat64
 )
 
+type metricOptions struct {
```
songy23: Please leave a TODO here for adding constant labels once the spec is approved.

rghetia: sure.

PR #1088, opened 2019-04-01
Bump up the version to 0.21.0

rghetia:


jadekler:
This does not appear to have resultedin a tag. Is that intentional?

songy23:
This tag should actually be "0.21.0-SNAPSHOT" or "Pre-0.21.0". We only push a new tag when releasing (https://github.com/census-instrumentation/opencensus-go/releases).

PR #1090, opened 2019-04-03
Add support for cumulative.

rghetia:


metric/registry.go
```diff
@@ -134,11 +134,79 @@ func bmTypeToMetricType(bm *baseMetric) metricdata.Type {
...
+	}
+	_, err := r.initBaseMetric(&i.g, name, mos...)
+	if err != nil {
+		return nil, err
+	}
+	return i, nil
+}
+
+// AddFloat64DerivedCumulative creates and adds a new derived float64-valued gauge to this registry.
+// A derived gauge is convenient form of gauge where the object associated with the gauge
```
rghetia: s/gauge/cumulative/

metric/common.go
```diff
@@ -54,9 +54,26 @@ type baseEntry interface {
 	read(t time.Time) metricdata.Point
 }
 
+func (bm *baseMetric) startTime() time.Time {
+	switch bm.bmType {
+	case cumulativeInt64:
```
songy23: Nit: all four cases can be combined to one line `case cumulativeInt64, cumulativeFloat64...`

metric/common.go
```diff
@@ -54,9 +54,26 @@ type baseEntry interface {
...
+		return bm.start
+	case cumulativeFloat64:
+		return bm.start
+	case derivedCumulativeInt64:
+		return bm.start
+	case derivedCumulativeFloat64:
+		return bm.start
+	default:
+		// gauges don't have start time.
+		return time.Now()
```
songy23: Better to return nil here, and check for nil in `read()`.

metric/common.go
```diff
@@ -54,9 +54,26 @@ type baseEntry interface {
...
+	}
+}
+
 // Read returns the current values of the baseMetric as a metric for export.
 func (bm *baseMetric) read() *metricdata.Metric {
 	now := time.Now()
+	startTime := bm.startTime()
```
songy23: `if startTime == nil { startTime = now }` so that we can guarantee startTime and now are the same.

metric/cumulative.go
```diff
@@ -0,0 +1,224 @@
...
+	"math"
+	"sync/atomic"
+	"time"
+
+	"go.opencensus.io/metric/metricdata"
+)
+
+// Float64Cumulative represents a float64 value that can only go up.
+//
+// Float64Cumulative maintains a float64 value for each combination of of label values
```
songy23: of of -> of

metric/cumulative.go
```diff
@@ -0,0 +1,224 @@
...
+			equalOrLess = true
+		}
+	}
+}
+
+// Inc increments the cumulative entry value by val. It returns without incrementing if the val
+// is negative.
+func (e *Float64CumulativeEntry) Inc(val float64) {
+	var swapped bool
+	if val < 0.0 {
```
songy23: Nit: if val == 0.0 you can also return.

metric/cumulative.go
```diff
@@ -0,0 +1,224 @@
...
+		oldVal := atomic.LoadUint64(&e.val)
+		newVal := math.Float64bits(math.Float64frombits(oldVal) + val)
+		swapped = atomic.CompareAndSwapUint64(&e.val, oldVal, newVal)
+	}
+}
+
+// Int64Cumulative represents a int64 cumulative value that can only go up.
+//
+// Int64Cumulative maintains an int64 value for each combination of label values passed to the
+// Set or Add methods.
```
songy23: s/Add/Inc

metric/cumulative.go
```diff
@@ -0,0 +1,224 @@
...
+		} else {
+			equalOrLess = true
+		}
+	}
+}
+
+// Inc increments the current cumulative entry value by val. It returns without incrementing if
+// the val is negative.
+func (e *Int64CumulativeEntry) Inc(val int64) {
+	if val < 0 {
```
songy23: Same here.

metric/common.go
rghetia: thx. I was looking for that syntax.

metric/common.go
rghetia: done.

metric/common.go
rghetia: done.

metric/cumulative.go
rghetia: done.

metric/cumulative.go
rghetia: done.

metric/cumulative.go
rghetia: done.

PR #1091, opened 2019-04-03
Bump up the version to 0.19.3

songy23:
v0.19.3 contains all commits from v0.20.0 except https://github.com/census-instrumentation/opencensus-go/commit/3b8e2721f2c3c01fa1bf4a2e455874e7b8319cd7.

Reason for making this patch release: https://github.com/census-instrumentation/opencensus-go/pull/1052#issuecomment-479621304.

### DIFF
--- a/opencensus.go
+++ b/opencensus.go
@@ -17,5 +17,5 @@ package opencensus // import "go.opencensus.io"
 
 // Version is the current release version of OpenCensus in use.
 func Version() string {
-	return "0.20.0"
+	return "0.19.3"
 }


songy23:
Suggested by @odeke-em. The projects that have dependencies on Jaeger can upgrade to v0.19.3. (Upgrading to v0.20.0 will break their build.)

songy23:
Build failure is:
```bash
exporter.Version is out of date with Git tags. Got 0.19.3; want something greater than 0.20.0
```
We're trying to make a patch release with a previous version so IMO we don't need to worry about this warning.

songy23:
The reason for making a 0.19.3 release instead of 0.20.1 is if users specify "latest" for opencensus-go:
`require(go.opencensus.io latest)` they'll automatically get the latest version. If we release 0.20.1 they'll get 0.20.1 by default, which may cause other problems. With 0.19.3 users won't get this version unless explicitly specified.

PR #1105, opened 2019-04-12
Add support for metrics in prometheus exporter

### LINKED ISSUE #1040: add support for metrics in prometheus exporter
Similar to work done for stackdriver export in [metrics.go](https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/metrics.go)

rghetia:
fixes #1040

### DIFF
--- a/exporter/prometheus/prometheus.go
+++ b/exporter/prometheus/prometheus.go
@@ -29,6 +29,9 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"go.opencensus.io/metric/metricexport"
+	"go.opencensus.io/metric/metricdata"
+	"context"
 )
 
 // Exporter exports stats to Prometheus, users need
@@ -61,6 +64,8 @@ func NewExporter(o Options) (*Exporter, error) {
 		c:       collector,
 		handler: promhttp.HandlerFor(o.Registry, promhttp.HandlerOpts{}),
 	}
+	collector.ensureRegisteredOnce()
+
 	return e, nil
 }
 
@@ -154,6 +159,9 @@ type collector struct {
 	registeredViewsMu sync.Mutex
 	// registeredViews maps a view to a prometheus desc.
 	registeredViews map[string]*prometheus.Desc
+
+	// reader reads metrics from all registered producers.
+	reader  *metricexport.Reader
 }
 
 func (c *collector) addViewData(vd *view.Data) {
@@ -166,16 +174,7 @@ func (c *collector) addViewData(vd *view.Data) {
 }
 
 func (c *collector) Describe(ch chan<- *prometheus.Desc) {
-	c.registeredViewsMu.Lock()
-	registered := make(map[string]*prometheus.Desc)
-	for k, desc := range c.registeredViews {
-		registered[k] = desc
-	}
-	c.registeredViewsMu.Unlock()
-
-	for _, desc := range registered {
-		ch <- desc
-	}
+	c.readDesc(ch)
 }
 
 // Collect fetches the statistics from OpenCensus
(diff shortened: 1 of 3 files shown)


exporter/prometheus/prometheus_test.go
```diff
@@ -187,7 +55,7 @@ func TestMetricsEndpointOutput(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to create prometheus exporter: %v", err)
 	}
-	view.RegisterExporter(exporter)
+	//view.RegisterExporter(exporter)
```
songy23: Nit: if these are not needed any more then just remove them.

exporter/prometheus/prometheus.go
```diff
@@ -123,11 +96,8 @@ func (o *Options) onError(err error) {
 // corresponding Prometheus Metric: SumData will be converted
 // to Untyped Metric, CountData will be a Counter Metric,
 // DistributionData will be a Histogram Metric.
+// Deprecated in lieu of metricexport.Reader interface.
 func (e *Exporter) ExportView(vd *view.Data) {
```
songy23: Not sure if it's the right way to go to simply remove the support for view data. From what I saw Stackdriver exporter still supports both: https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/stats.go#L173.

rghetia: For existing user of SD, they have to explicitly add a call to StartMetricExporter and StopMetricExporter. So to avoid breaking, support for ExportView was left as is. In case of prometheus, there is no need to for ExportView anymore because Gatherer can always read using new interface. There is no additional code change required to enable collecting metrics using reader interface.

exporter/prometheus/prometheus_test.go
rghetia: fixed.

exporter/prometheus/prometheus.go
```diff
@@ -145,151 +115,163 @@ type collector struct {
...
+	c        *collector
+	metricCh chan<- prometheus.Metric
+}
 
-	default:
-		return nil, fmt.Errorf("aggregation %T is not yet supported", v.Aggregation)
+// ExportMetrics exports to the Prometheus.
+// Each OpenCensus Metric will be converted to
+// corresponding Prometheus Metric:
+// SumData will be converted to Untyped Metric,
```
songy23: These comments no longer apply here.

exporter/prometheus/prometheus.go
```diff
@@ -145,151 +115,163 @@ type collector struct {
...
+			// Histograms are cumulative in Prometheus.
+			// Get cumulative bucket counts.
+			cumCount := uint64(0)
+			for i, b := range v.BucketOptions.Bounds {
+				cumCount += uint64(v.Buckets[i].Count)
+				points[b] = cumCount
+			}
+			return prometheus.NewConstHistogram(desc, uint64(v.Count), v.Sum, points, labelValues...)
+		default:
+			return nil, pointTypeError(point)
```
songy23: Nit: leave a TODO to support Summary metrics.

exporter/prometheus/prometheus_test.go
```diff
@@ -212,9 +129,9 @@ func TestCumulativenessFromHistograms(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to create prometheus exporter: %v", err)
 	}
-	view.RegisterExporter(exporter)
-	reportPeriod := time.Millisecond
-	view.SetReportingPeriod(reportPeriod)
+	//view.RegisterExporter(exporter)
```
songy23: There are a few more commented lines that can be removed.

exporter/prometheus/prometheus.go
rghetia: done.

exporter/prometheus/prometheus.go
rghetia: fixed. I have referred to Metric Types (like TypeCumulativeInt64)  in modified comment instead of stats aggregation type (like sum).

exporter/prometheus/prometheus_test.go
rghetia: removed all of them.

exporter/prometheus/prometheus.go
```diff
@@ -145,151 +115,164 @@ type collector struct {
...
 	}
-	return buf.String()
+	return values
 }
 
-func (c *collector) cloneViewData() map[string]*view.Data {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+func pointTypeError(point metricdata.Point) error {
+	return fmt.Errorf("point type %T is not yet supported", point)
```
songy23: Nit: the error message could be a bit confusing, consider "point type %T doesn't match with metric type".

rghetia: done.

PR #1107, opened 2019-04-15
Add gauges example.

### LINKED ISSUE #1035: Add gauge example
Add gauge example just like one in [java](https://www.google.com/url?q=https://github.com/census-instrumentation/opencensus-java/tree/master/examples/src/main/java/io/opencensus/examples/gauges).

- [x] gauge example. #1107 
- [x] derived gauge example. #1110

rghetia:
fixes #1035 partially

### DIFF
--- a/examples/gauges/README.md
+++ b/examples/gauges/README.md
@@ -59,7 +59,7 @@ Create a gauge metric. In this example we have two metrics.
 [embedmd]:# (gauge.go alloc)
 ```go
 allocGauge, err := r.AddInt64Gauge(
-	"process/heap/alloc",
+	"process_heap_alloc",
 	metric.WithDescription("Process heap allocation"),
 	metric.WithUnit(metricdata.UnitBytes))
 if err != nil {
@@ -72,7 +72,7 @@ if err != nil {
 [embedmd]:# (gauge.go idle)
 ```go
 ratioGauge, err := r.AddFloat64Gauge(
-	"process/heap/idle_to_alloc_ratio",
+	"process_heap_idle_to_alloc_ratio",
 	metric.WithDescription("process heap idle to allocate ratio"),
 	metric.WithUnit(metricdata.UnitDimensionless))
 if err != nil {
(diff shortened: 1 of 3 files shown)


examples/gauges/README.md
```diff
@@ -0,0 +1,289 @@
...
+}
+```
+
+
+### Set gauge values
+Use `Set` or `Add` function to update the value of gauge entries. You can call these methods anytime based on your metric and your application. In this example, `Set` is called periodically.
+
+[embedmd]:# (gauge.go record)
+```go
+			allocEntry.Set(int64(getAlloc()))     // int64 gauge
```
songy23: Nit: indentation seems to be off.

examples/gauges/README.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Create gauge metric
+Create a gauge metric. In this example we have two metrics.
+
+**process_heap_alloc**
+
+[embedmd]:# (gauge.go alloc)
+```go
+allocGauge, err := r.AddInt64Gauge(
+	"process/heap/alloc",
```
songy23: Nit: Prometheus doesn't allow '/', consider replacing with '_' instead.

examples/gauges/README.md
rghetia: it is auto generated.

examples/gauges/README.md
rghetia: prometheus exporter replaces '/' with '_'.   
Should we '/' or '_' be based on exporter? or based on what is more appropriate generically?

songy23: In this example I'm inclined to be consistent with the exporter, since it will be confusing if users see the '/' characters being swapped on Prometheus endpoint. In addition the readme suggests the metric names use '_' ("process_heap_alloc"): https://github.com/census-instrumentation/opencensus-go/pull/1107/files#diff-2b3521dd2010161fca962a345a4eaddeR21.

rghetia: SG

PR #1118, opened 2019-04-19
Move exporters out from core.

songy23:
Also run `$ GO111MODULE=on go mod tidy`.


rghetia:
@songy23 , let me remove prometheus from gauge examples.

rghetia:
> @songy23 , let me remove prometheus from gauge examples.

I mean let me make the example with main package so it doesn't create dependencies.

rghetia:
after vanity-url [PR](https://github.com/census-ecosystem/opencensus-go-vanityurls/pull/6) is merged I can replace with new dependencies in examples
Ideally, it would be nice to create a separate repo for example so we don't haver circular dependencies.

songy23:
> after vanity-url PR is merged I can replace with new dependencies in examples
Ideally, it would be nice to create a separate repo for example so we don't haver circular dependencies.

Agree, I'll hold off this PR until examples are updated. Or we can move the exporter-specific examples to exporter repos.

rghetia:
The problem with moving examples to exporter repo is that most of the example related to core functionality will move to prometheus just because we use prometheus in most (if not all) examples.

@bogdandrutu, should we create example repo and simply link it in readme.md in core library?

bogdandrutu:
Probably we can have in each exporter repo it's own example.

rghetia:
> Probably we can have in each exporter repo it's own example.

That would require that we repeat feature examples (like gauge) in all exporters.

songy23:
Rebased against the latest `dev` branch (included https://github.com/census-instrumentation/opencensus-go/pull/1128). Should be good to merge now.

hsanjuan:
If anyone lands here wondering where they were moved to: https://github.com/census-ecosystem (with imports as `contrib.go.opencensus.io/exporter/jaeger` etc.)

PR #1122, opened 2019-04-21
add constant labels to gauges and cumulative metrics

### LINKED ISSUE #1112: Add support for constant labels
Add support for constant labels. Similar to what is done in [java](https://github.com/census-instrumentation/opencensus-java/pull/1842)

paivagustavo:
This PR adds support for constant labels on gauges. 

The proposed syntax is:
```go
r := NewRegistry()

 	f, _ := r.AddFloat64Gauge("TestGaugeWithConstLabel",
		WithLabelKeys("k1"), WithConstLabel("const", "same"), WithConstLabel("const2", "same2"))
```

@rghetia this is my first contribution and as you were the one who created the issue, could you review this?

closes #1112

### DIFF
--- a/metric/cumulative_test.go
+++ b/metric/cumulative_test.go
@@ -90,8 +90,8 @@ func TestCumulativeConstLabel(t *testing.T) {
 	f, _ := r.AddFloat64Cumulative("TestCumulativeWithConstLabel",
 		WithLabelKeys("k1"),
 		WithConstLabel(map[metricdata.LabelKey]metricdata.LabelValue{
-			metricdata.LabelKey{Key: "const"}:  metricdata.NewLabelValue("same"),
-			metricdata.LabelKey{Key: "const2"}: metricdata.NewLabelValue("same2"),
+			{Key: "const"}:  metricdata.NewLabelValue("same"),
+			{Key: "const2"}: metricdata.NewLabelValue("same2"),
 		}))
 
 	e, _ := f.GetEntry(metricdata.LabelValue{})
--- a/metric/cumulative_test.go
+++ b/metric/cumulative_test.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+
 	"go.opencensus.io/metric/metricdata"
 )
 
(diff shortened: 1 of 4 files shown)


metric/registry.go
```diff
@@ -61,13 +62,21 @@ func WithUnit(unit metricdata.Unit) Options {
...
-// WithLabelKeys applies provided label.
+// WithLabelKeys applies provided label keys.
 func WithLabelKeys(labelKeys ...string) Options {
 	return func(mo *metricOptions) {
 		mo.labelkeys = labelKeys
 	}
 }
 
+// WithConstLabel applies provided constant label.
+func WithConstLabel(labelKey, labelValue string) Options {
```
rghetia: it might be better to take multiple pairs of constant key/value. Perhaps a map would be better.
```go
func WithConstLabel(constLabel map[LabelKey]LabelValue) Options {
}

Comment (paivagustavo) [reply metric/registry.go]: changed.

**Comment by paivagustavo** [comment]
@rghetia I've rebased to dev and change the target branch (I may have pick some commits from the master that was not on dev yet, is it a problem?)

While doing this, I noticed that the `canonicalize` was not deterministic and sometimes the `TestGauge` was failing, I have fixed this and made it more reliable while keeping the same purpose.

**Comment by rghetia** [comment]
> @rghetia I've rebased to dev and change the target branch (I may have pick some commits from the master that was not on dev yet, is it a problem?)

yes, that will be a problem. you can  do `git rebase --onto dev HEAD~2` since you two commits.

**Comment by paivagustavo** [comment]
Done, the PR now contains only the correct commits.

File: metric/cumulative_test.go [hunk 2]
```diff
@@ -83,6 +84,62 @@ func TestCumulative(t *testing.T) {
...
+		WithConstLabel(map[metricdata.LabelKey]metricdata.LabelValue{
+			{Key: "const"}:  metricdata.NewLabelValue("same"),
+			{Key: "const2"}: metricdata.NewLabelValue("same2"),
+		}))
+
+	e, _ := f.GetEntry(metricdata.LabelValue{})
+	e.Inc(5)
+	e, _ = f.GetEntry(metricdata.NewLabelValue("k1v1"))
+	e.Inc(1)
+	e, _ = f.GetEntry(metricdata.NewLabelValue("k1v1"))
```
rghetia: nit: e is not used again. Please remove it.

paivagustavo: remove :)

PR #1125, opened 2019-04-23
Add support for Tag metadata

rghetia:
Replaces #1032

### DIFF
--- a/tag/map.go
+++ b/tag/map.go
@@ -155,7 +155,7 @@ func createMetadatas(mds ...Metadata) metadatas {
 			}
 		}
 	} else {
-		TTLUnlimitedPropagation(&metas)
+		WithTTLUnlimitedPropagation()(&metas)
 	}
 	return metas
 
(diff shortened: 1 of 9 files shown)


tag/map.go
```diff
@@ -95,13 +100,14 @@ type Mutator interface {
 // Insert returns a mutator that inserts a
 // value associated with k. If k already exists in the tag map,
 // mutator doesn't update the value.
-func Insert(k Key, v string) Mutator {
+// Metadata applies metadata to the tag. It is optional.
```
songy23: What if there're more than one metadatas and they're conflicting? E.g `Insert(k, v, unlimitedPropagation, noPropagation)`?

rghetia: the last one will be effective.

songy23: Consider adding a note on the API document, and adding a few tests on this case.

rghetia: added a note in the apis and testcases.

Issue #1093, opened 2019-04-07
ochttp Go integration payload information


alexamies:
Please add payload information to the ochttp Go integration. This is one of the very helpful items for debugging, and present in the Java implementation and also in the Go gRPC integration. For example, reasoning about why one trace has a higher latency than another, the first thing to look at would be the payload size.

**What I found**
I tried a modified version of the [Go HTTP client tutorial](https://opencensus.io/guides/http/go/net_http/client/) using Zipkin and Stackdriver. They both give similar views omitting the payload size, like this screenshot of Zipkin
![ochttp_zipkin](https://user-images.githubusercontent.com/5554156/55688353-69cd3500-592c-11e9-9cf5-48f2e0d2677c.png) with no payload size. It is not nearly as useful as the Java equivalent, which shows the payload size in Stackdriver as events, screenshot:
![http_java_payload](https://user-images.githubusercontent.com/5554156/55688377-aef16700-592c-11e9-9ea0-798bd89cc844.png)

This could be implemented in the [ochttp Go plugin](https://github.com/census-instrumentation/opencensus-go/tree/master/plugin/ochttp) with payload events using code like the gRPC integration in [ocgrpc / trace_common.go](https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ocgrpc/trace_common.go#L92).

alexamies:
Server integration was completed with PR #1129. Client integration is still TODO.

PR #1129, opened 2019-04-24
Added payload size to http server integration

alexamies:
Added payload size to http server integration, logged in the LogExporter, and modified example to use LogExporter


alexamies:
Partially addresses issue
https://github.com/census-instrumentation/opencensus-go/issues/1093

plugin/ochttp/server.go
```diff
@@ -201,6 +207,9 @@ func (t *trackingResponseWriter) Header() http.Header {
 func (t *trackingResponseWriter) Write(data []byte) (int, error) {
 	n, err := t.writer.Write(data)
 	t.respSize += int64(n)
+	// Add message event for request bytes read.
```
rghetia: nit: s/read/sent/

alexamies: Corrected in latest commit

Issue #973, opened 2018-11-12
Replace "golang.org/x/net/context" with "context"


rakyll:
gRPC plugin was depending on golang.org/x/net/context due to the 1.6 support gRPC couldn't deprecate for a long time. With https://github.com/grpc/grpc-go/pull/2439 merged, we can replace "golang.org/x/net/context" with the standard library "context" package.

cc @ramonza @odeke-em

PR #983, opened 2018-11-21
Replace "x/net/context" import with "context"

rakyll:
We also need to pin to a gRPC version that includes
https://github.com/grpc/grpc-go/commit/04ea82009cdb9ecdefc6289f4c93ec919a10b3b6.

PR #1138, opened 2019-05-03
Replace "x/net/context" import with "context"

songy23:
Fixes https://github.com/census-instrumentation/opencensus-go/issues/973.

Supersedes https://github.com/census-instrumentation/opencensus-go/pull/983.


songy23:
Appveyor failure is not related to this change.

PR #1141, opened 2019-05-27
tag: add func MustNewKey to wrap NewKey with panic

dolmen:
For conveniency when creating `tag.Key`, provide a `Must` wrapper of `tag.NewKey` that panic if the key is invalid: `func MustNewKey(name string) Key`.

### DIFF
--- a/tag/key.go
+++ b/tag/key.go
@@ -29,6 +29,16 @@ func NewKey(name string) (Key, error) {
 	return Key{name: name}, nil
 }
 
+// MustNewKey creates or retrieves a string key identified by name.
+// An invalid key name raises a panic.
+func MustNewKey(name string) Key {
+	k, err := NewKey(name)
+	if err != nil {
+		panic(err)
+	}
+	return k
+}
+
 // Name returns the name of the key.
 func (k Key) Name() string {
 	return k.name


rghetia:
@dolmen this wrapper is not useful for most users of the library. Hence it should not be added.

dolmen:
As an OpenCensus beginner it made sense to me to declare tags as global (const-like) variables, and panicing on tag name validation error appeared to be future proof, especially as OpenCensus is not yet at v1.

I know see how `plugin/ochttp` [defines its tags](https://godoc.org/go.opencensus.io/plugin/ochttp#KeyClientMethod):
```go
var KeyClientMethod, _ = tag.NewKey("http_client_method")
```
Is it the idiom we are supposed to use in external code?

rghetia:
> As an OpenCensus beginner it made sense to me to declare tags as global (const-like) variables, and panicing on tag name validation error appeared to be future proof, especially as OpenCensus is not yet at v1.
> 
> I now see [how `plugin/ochttp` defines its tags](https://godoc.org/go.opencensus.io/plugin/ochttp#KeyClientMethod):
> 
> ```go
> var KeyClientMethod, _ = tag.NewKey("http_client_method")
> ```
> 
> Is it the idiom we are supposed to use in external code? This doesn't seem as safe as this:
> 
> ```go
> var KeyClientMethod = tag.MustNewKey("http_client_method")
> ```

I agree it is safe to use MustNewKey. In most cases tag Key will be created  during init and MustNewKey is more appropriate. However,  when tags are received over the wire from remote one should use NewKey as we don't want to raise panic for malformed tags. 

@rakyll do you have any comments?.

rakyll:
I'm in favor of adding a MustX API. We discussed this before but haven't done it. The current suggested usage where we ignore the error is not ideal.

PR #1150, opened 2019-07-06
fix span decoding for StackDriver / GoogleCloudFormat propagation

mhindery:
### What version of OpenCensus are you using?
- opencensus-python [0.6.0]
- opencensus-go 0.22.0; stackdriver-exporter 0.10.0

### What version of Go are you using?
go version go1.11.5 darwin/amd64

### What did you do?
The span ID is a 64-bit id (confusing sidenote, the [specs here](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Span.md#spanid) claim it to be 8 bit). Both the [Python](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/span_context.py#L151-L158) (16-char hex encoded string) and [Go](https://github.com/census-instrumentation/opencensus-go/blob/master/trace/trace.go#L576-L585) (8-byte array) implementation do this correctly.

However, there is an incompatibility in the propagation step. What the Python code does with the `GoogleCloudFormatPropagator ` is [put that 16-char hex string in the header](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/propagation/google_cloud_format.py#L108-L111). Then the header is something like

```X-Cloud-Trace-Context: 129b452d2ef458f798c10811e2c28cf6/14a650d3b8b5ddde;o=1```

Now what the Go [stackdriver exporter tries to do](https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/propagation/http.go#L65-L75) is parsing that hexadecimal string as an integer with

```go
sid, err := strconv.ParseUint(spanstr, 10, 64)
```

This throws an error, leading the trace propagation to fail and traces being separated between services.

The fix is simple, the base should be 16 instead of 10: https://goplay.space/#iD9_V7MH_Dd

### DIFF
--- a/exporter/stackdriver/propagation/http.go
+++ b/exporter/stackdriver/propagation/http.go
@@ -68,7 +68,7 @@ func (f *HTTPFormat) SpanContextFromRequest(req *http.Request) (sc trace.SpanCon
 	if semicolon != -1 {
 		spanstr, h = h[:semicolon], h[semicolon+1:]
 	}
-	sid, err := strconv.ParseUint(spanstr, 10, 64)
+	sid, err := strconv.ParseUint(spanstr, 16, 64)
 	if err != nil {
 		return trace.SpanContext{}, false
 	}


mhindery:
Going to close this as this code has been moved to a new repo, so reopening it there

Issue #742, opened 2018-05-07
HTTP views do not match spec


ramonza:
Measure, view and tag names of ochttp are all out of sync with the spec. We need to implement the views as per the spec.

johansja:
Just wondering if there is any update on this. `http_server_route` is one of the important tag that I am looking for in the server stats.

johansja:
I'm OK to work on a PR for this. However, I would also like to understand the direction of this project as I know OpenTelemetry is on its way.

johansja:
Submitted a PR for this: https://github.com/census-instrumentation/opencensus-go/pull/1159

johansja:
I'll purposely leave out backward compatibility in the PR. Let me know if it is a concern.

rghetia:
> I'm OK to work on a PR for this. However, I would also like to understand the direction of this project as I know OpenTelemetry is on its way.

@johansja It would be best to do this in opentelemetry.

johansja:
What is the status of OpenTelemetry at the moment? Is it on parity with this now? I am looking for traces and stats with Stackdriver exporter.

rghetia:
Beta version of OpenTelemetry should be available by Oct 1st.

johansja:
What are the things that will be available in beta?

PR #1159, opened 2019-09-03
Update ochttp stats according to spec.

johansja:
Related to this issue: https://github.com/census-instrumentation/opencensus-go/issues/742

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	golang.org/x/net v0.0.0-20190501004415-9ce7a6920f09
 	golang.org/x/sys v0.0.0-20190502145724-3ef323f4f1fd // indirect
 	golang.org/x/text v0.3.2 // indirect
+	google.golang.org/appengine v1.4.0 // indirect
 	google.golang.org/genproto v0.0.0-20190425155659-357c62f0e4bb // indirect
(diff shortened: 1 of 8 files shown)


johansja:
It seems like the failure was caused by something else. And when I ran the test locally, it works fine.

rghetia:
@johansja I think it would be best to make this change in opentelemetry given that it would create a breaking change.

johansja:
Thanks for looking into this. Would you consider adding some tag key to the view a breaking change?

The main thing that I am trying to achieve right now is to add route-specific metrics such as latency.

rghetia:
> Thanks for looking into this. Would you consider adding some tag key to the view a breaking change?
> 
> The main thing that I am trying to achieve right now is to add route-specific metrics such as latency.

It is not a breaking change but it adds another dimension which will increase number of time-series for everyone. That is not desirable. You could create your own view and have route tag in it.

bogdandrutu:
@johansja you can define your own views that include the route and keep the default without for the moment. This is just an idea.

johansja:
Hi all, thanks for the suggestion. I guess the custom view or forking might be the way to go temporarily while we wait for OpenTelemetry to be more mature. Hopefully, OpenTelemetry will have metrics per route by default.

PR #1160, opened 2019-09-03
Use time.Since rather than time.Now().Sub

bvwells:
Use time.Since rather than time.Now().Sub.


bvwells:
Thanks @rghetia for the super fast merge!

PR #1161, opened 2019-09-03
Fix bug in ocgrpc client logging

bvwells:
Fix bug in ocgrpc client logging. Info is nil in the logging branch so info.FullMethodName will cause a crash.


bvwells:
Thanks for the merge @rghetia!

PR #1164, opened 2019-09-04
Update LRU usage

joshuaseaton:
This change updates the LRU cache implementation used in the trace
package, replacing github.com/hashicorp/golang-lru with
github.com/golang/groupcache, while preserving the manner in which trace
abstractions make use of the data structure.

The reason for this change is that github.com/hashicorp/golang-lru has a
Mozilla license, which is different from that of
github.com/census-instrumentation/opencensus-go. The difference (and the
mere presence of the former) makes it more difficult for open-source
projects (e.g., https://fuchsia.dev) to vendor the latter.


joshuaseaton:
I'm seeing continuous-integration/appveyor/pr checks fail with Windows `runtime` package errors. Is this a known issue?

rghetia:
> I'm seeing continuous-integration/appveyor/pr checks fail with Windows `runtime` package errors. Is this a known issue?

yes, it is a known issue.

opencensus.go
```diff
@@ -17,5 +17,5 @@ package opencensus // import "go.opencensus.io"
 
 // Version is the current release version of OpenCensus in use.
 func Version() string {
-	return "0.23.0"
+	return "0.23.1"
```
rghetia: this is changed after minor release. please remove it.

trace/lrumap.go
```diff
@@ -15,23 +15,47 @@
...
+		lm.droppedCount++
+	}
 	return lm
 }
 
-func (lm *lruMap) add(key, value interface{}) {
-	evicted := lm.simpleLruMap.Add(key, value)
-	if evicted {
-		lm.droppedCount++
+func (lm lruMap) Len() int {
```
rghetia: Len shouldn't be exported. Same with Keys, Add and Get.

joshuaseaton: I was treating the exported methods (of the unexported struct) as the API of the lruMap, which isn't explicitly contrary to best practices as far as I know. (Though if that is indeed the case, a style link would be appreciated.) But you of course set the convention here. Done!

PR #1166, opened 2019-09-10
Use stats.UnitDimensionless rather than deprecated stats.UnitNone

bvwells:
Update quickstart stats example to use stats.UnitDimensionless rather than deprecated stats.UnitNone.


bvwells:
Thanks @rghetia!

PR #1170, opened 2019-09-11
Remove Gopkg.toml and Gopkg.lock

### LINKED ISSUE #1048: Remove dep files
This is a reminder to remove Gopkg.lock and Gopkg.toml once the community transition into the Go modules. In the meanwhile, we are trying to keep both dep and modules files up-to-date. 

Please note that the version resolution method they use are different hence there might be cases of incompatibilities. IMHO, we should drop support for dep when it happens.

cc @rghetia @bogdandrutu

rghetia:
fixes #1048


bogdandrutu:
Do we need to remove some of the tools that generated these files?

rghetia:
> Do we need to remove some of the tools that generated these files?

dep tool is used to update these files but we don't install as part of our Makefile. We also don't use them in any CI scripts either.

bogdandrutu:
@rghetia consider to remove the support in the Makefile for that :) (if any)

PR #1174, opened 2019-09-20
Add AlwaysParentSampler

s4nji:
Adds a sampler that inherits sampling decision from parent span, and defaults to not sample if it has no parent.

### DIFF
--- a/trace/sampling.go
+++ b/trace/sampling.go
@@ -73,3 +73,11 @@ func NeverSample() Sampler {
 		return SamplingDecision{Sample: false}
 	}
 }
+
+// AlwaysParentSampler returns a Sampler that only samples spans,
+// and only if the parent span has already decided that it should be sampled.
+func AlwaysParentSampler() Sampler {
+	return func(p SamplingParameters) SamplingDecision {
+		return SamplingDecision{Sample: p.ParentContext.IsSampled()}
+	}
+}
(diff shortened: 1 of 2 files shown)


s4nji:
Test cases for `AlwaysParentSampler` added.  

However I'm now unsure if it should (not) sample new spans without parents.
Currently it doesn't sample spans without parents: this is exactly `ProbabilitySampler` with probability set to 0, and it feels redundant.

trace/sampling.go
```diff
@@ -73,3 +73,11 @@ func NeverSample() Sampler {
 		return SamplingDecision{Sample: false}
 	}
 }
+
+// AlwaysParentSampler returns a Sampler that only samples spans,
+// and only if the parent span has already decided that it should be sampled.
```
rghetia: nit: It may be better to reword the Doc to
```
AlwaysParentSampler returns a Sampler that samples a span if it's parent span is sampled. The sampler does not sample a root span (a span without a parent).
```

rghetia:
> Test cases for `AlwaysParentSampler` added.
> 
> However I'm now unsure if it should (not) sample new spans without parents.
> Currently it doesn't sample spans without parents: this is exactly `ProbabilitySampler` with probability set to 0, and it feels redundant.

I missed this comment. I agree that it is redundant.

s4nji:
Closing, as it is unclear and requires further specification.

PR #1176, opened 2019-09-25
trace: add starttime start option

gwik:
Adds a StartTime option when creating a new span.
In some cases, you are able to trace only after the operation
was made. for example in some post-operation hook/observer.

    func myHook(ctx context.Context, info queryInfo) {
       _, span := StartSpan("mydatabase", WithStartTime(time.Now().Sub(info.Duration()))
       span.End()
    }


trace/trace.go
```diff
@@ -161,6 +165,13 @@ func WithSampler(sampler Sampler) StartOption {
 	}
 }
 
+// WithStartTime sets the span start time.
```
rghetia: Perhaps a bit more description would be better.
```
WithStartTime sets the start time of the span to provided time t, when it is started.
In absence of this option, wall clock time is used as the start time.
This option is typically used when starting of a span is delayed.
```

Issue #1171, opened 2019-09-17
Allow customization of health endpoints in ochttp server


tjcelaya:
This is part feature request and part bug. The automatic tracing includes a check where both the ochttp client and server components inspect the URL path before attaching a trace to the request being made. Unfortunately those paths to skip tracing are hardcoded:

https://github.com/census-instrumentation/opencensus-go/blob/65310139a05de5c10077b75ac45eac743aa01214/plugin/ochttp/trace.go#L231-L238

called from the server here:

https://github.com/census-instrumentation/opencensus-go/blob/65310139a05de5c10077b75ac45eac743aa01214/plugin/ochttp/server.go#L89-L91

**Is your feature request related to a problem? Please describe.**
The `/health` and `/metrics` endpoints in our server end up being sampled alongside service endpoints and are generating excessive traces.

**Describe the solution you'd like**
Enhance `ochttp.Handler` to allow callers to supply a set of paths or path pattern to use for unconditionally skipping sampling. This could either be something like a `HealthEndpoints []string` field for directly specifying paths, a `HealthEndpointPattern *regex.Regex` for use with `Match`/`MatchString`. The default paths currently in use could be used as the default to provide backwards compatibility.

While a similar enhancement could also be introduced for the client this issue is focused on improving the server-side sampling logic.

**Describe alternatives you've considered**
Exposing health/metrics/etc endpoints through a different `http.Server` instance is possible and could be used to bypass ochttp entirely but would have to listen on a different port and introduce the possibility of unexpected or divergent behavior from the state of the service's primary `http.Server` (e.g. during server Shutdown or config reload).

**Additional context**
I could find time to work on a PR for this soon but would greatly appreciate if a maintainer could weigh in with a preferred approach and a general suggestion for what kinds of unit tests are expected. Perhaps a new unit test is desired or it might be sufficient to adds a few case to this test (or both):

https://github.com/census-instrumentation/opencensus-go/blob/3da91ae5e6af5a3e42dac7761866fa513a753d5e/plugin/ochttp/trace_test.go#L204-L210

PR #1177, opened 2019-09-25
Allow overriding health endpoint check in handler

tjcelaya:
Addresses #1171 by allowing users to supply a `func (string) bool` which can be used to determine if the path is a health-check, falling back to the existing solution if no func is supplied.

Test copied from `TestIgnoreHealthz` to supply the new parameter but I'd be happy to merge them or modify the cases.

### DIFF
--- a/plugin/ochttp/server.go
+++ b/plugin/ochttp/server.go
@@ -73,7 +73,8 @@ type Handler struct {
 
 	// IsHealthEndpoint holds the function to use for determining if the
 	// incoming HTTP request should be considered a health check. If set and
-	// returns true, no trace will be recorded.
+	// returns true no trace will be recorded, otherwise the handler will
+	// use the private isHealthEndpoint func in this package.
 	IsHealthEndpoint func(string) bool
 }
 
--- a/plugin/ochttp/server.go
+++ b/plugin/ochttp/server.go
@@ -75,7 +75,7 @@ type Handler struct {
 	// incoming HTTP request should be considered a health check. This is in
 	// addition to the private isHealthEndpoint func which may also indicate
 	// tracing should be skipped.
-	IsHealthEndpoint func(string) bool
+	IsHealthEndpoint func(*http.Request) bool
 }
 
 func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


plugin/ochttp/server.go
```diff
@@ -70,6 +70,12 @@ type Handler struct {
 	// from the information found in the incoming HTTP Request. By default the
 	// name equals the URL Path.
 	FormatSpanName func(*http.Request) string
+
+	// IsHealthEndpoint holds the function to use for determining if the
+	// incoming HTTP request should be considered a health check. If set and
+	// returns true, no trace will be recorded, otherwise the private
+	// isHealthEndpoint function will be used.
```
rghetia: The IsHealthEndpoint is an 'Additional' check on top of default isHealthEndpoint check. I think the doc should reflect that. Currently it seems that it will use User supplied over the default.

plugin/ochttp/server_test.go
```diff
@@ -595,3 +595,63 @@ func TestIgnoreHealthz(t *testing.T) {
...
+
+	client := &http.Client{}
+	tests := []struct {
+		path               string
+		healthEndpointFunc func(string) bool
+	}{
+		{"/healthz", nil},
+		{"/_ah/health", nil},
+		{"/health", testHealthEndpointSkipArray},
+		{"/metrics", testHealthEndpointSkipArray},
```
rghetia: Need a test case for 
```
		{"/healthz", testHealthEndpointSkipArray},
		{"/_ah/health", testHealthEndpointSkipArray},
```

Also, TestIgnoreHealthz is not required as it is covered under this test. May be make this a generic name as it covers all cases.

plugin/ochttp/server.go
tjcelaya: This is somewhat ambiguous since the current order of the conditional could be confusing if users expect the func to have complete control over skipping tracing. If instead we flipped the order around to the following it would act as an override instead of an additional check:

`h.IsHealthEndpoint == nil && isHealthEndpoint(r.URL.Path) || h.IsHealthEndpoint(r.URL.Path)`

But then users would have to remember to include `/healthz` and `/_ah/health` when supplying a func for `IsHealthEndpoint`.

Which is preferred: the 'Additional' check currently implemented in the PR or giving the user complete control when they supply a func?

tjcelaya: I agree that the override/additional behavior is somewhat unclear so the comment has been to indicate that either one of the user-supplied func or the private default could cancel tracing.

plugin/ochttp/server.go
```diff
@@ -70,6 +70,12 @@ type Handler struct {
 	// from the information found in the incoming HTTP Request. By default the
 	// name equals the URL Path.
 	FormatSpanName func(*http.Request) string
+
+	// IsHealthEndpoint holds the function to use for determining if the
+	// incoming HTTP request should be considered a health check. This is in
+	// addition to the private isHealthEndpoint func which may also indicate
+	// tracing should be skipped.
+	IsHealthEndpoint func(string) bool
```
rghetia: Description looks good now.

It might be better to pass http.Request rather than just Path. For example, decision could be made on a header.

tjcelaya: Makes perfect sense since the other funcs also accept the `http.Request`, I feel silly for not having considered it sooner. Thanks for the suggestion!

Issue #1181, opened 2019-10-30
DistributionData and view.Rows in general do not get reset between ReportingIntervals


marwan-at-work:
I have a program that records the latency of a request using the OCHTTP server middleware. 

In my program, I am using the [DataDog OpenCensus Exporter](https://github.com/DataDog/opencensus-go-exporter-datadog) to report this metric. 

Therefore, I wanted to simulate a latency spike by having my handler do the following:

1. The first 4 minutes, the HTTP Handler takes between 200-400ms to respond.

2. The second 6 minutes, the HTTP handler takes between 1-2seconds (latency spike) 

3. The third 4 minutes, the HTTP handler takes between 200-400ms (latency goes back to normal) 

**The latency gets reported every 10 seconds**

In other words, OpenCensus calls DataDog's ExportView every 10 seconds.

OpenCensus collects the Min/Max latencies and passes them as DistributionData to the DataDog exporter. 

**My expectation is that for every 10 seconds, OpenCensus calculates the Min/Max/Mean/etc and passes them to an Exporter, but then it should reset those numbers so that they don't obscure the next bucket of time. Is this assumption incorrect?**

Even if this assumption is wrong, let's continue with how this behavior has obscured the latency spike significantly:

The fact that OpenCensus does not "reset" the Min every 10 seconds, means that the Min stays the same across the entire lifetime of the Go process. 

Therefore, if we had a dashboard of the "min" value only, you'll see that the min never went up, it stays as ~200ms for the entirety of the 14 minutes. Even though for 6 minutes, the Min value was at least 1000ms. 

On the other hand, if you have a "max" metric, you'll see that it went up to ~2s, but then it never went back down even though the traffic spike ended after 10 minutes. 

Predictably, if I "kill" the process and restart it, suddenly the dashboard gets fixed and that's because we no longer hold the "min/max" that was previously preventing the dashboard to be accurate. See the following screen shot for demonstration:

<img width="1247" alt="Screen Shot 2019-10-30 at 2 27 26 AM" src="https://user-images.githubusercontent.com/16294261/67836920-3d455a00-fac4-11e9-9516-bcf8d2438603.png">

What you see here is that the Purple line is the Max metric while the Blue line is the Min metric. 

When 4 minutes passed, the Purple line went up, but the Blue line stayed down. This is because even though new incoming requests were clocking between 1-2 seconds, OpenCensus kept comparing those numbers to the Min that was initially recorded in the first 4 minutes. 

[...]

PR #1182, opened 2019-10-30
clearRows on every ExportView

marwan-at-work:
Fixes https://github.com/census-instrumentation/opencensus-go/issues/1181 

Or at least I think so :) 

I am not familiar with the code that much so would love some close eyes on this. I am not sure what kind of impact this one line of code would have on performance. However, it does seem to fix the behavior described in the issue above and all the unit tests passed.

### DIFF
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -233,6 +233,7 @@ func (w *worker) reportView(v *viewInternal, now time.Time) {
 		e.ExportView(viewData)
 	}
 	exportersMu.Unlock()
+	v.clearRows()
 }
 
 func (w *worker) reportUsage(now time.Time) {
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,6 @@ require (
 	golang.org/x/net v0.0.0-20190620200207-3b0461eec859
 	golang.org/x/sys v0.0.0-20190502145724-3ef323f4f1fd // indirect
 	golang.org/x/text v0.3.2 // indirect
-	google.golang.org/appengine v1.4.0 // indirect
 	google.golang.org/genproto v0.0.0-20190425155659-357c62f0e4bb // indirect
 	google.golang.org/grpc v1.20.1
 )
--- a/go.sum
+++ b/go.sum
@@ -67,6 +67,7 @@ google.golang.org/grpc v1.19.0/go.mod h1:mqu4LbDTu4XGKhr4mRzUsmM4RtVoemTSY81AxZi
 google.golang.org/grpc v1.19.0/go.mod h1:mqu4LbDTu4XGKhr4mRzUsmM4RtVoemTSY81AxZiDr8c=
 google.golang.org/grpc v1.20.1 h1:Hz2g2wirWK7H0qIIhGIqRGTuMwTE8HEKFnDZZ7lm9NU=
 google.golang.org/grpc v1.20.1/go.mod h1:10oTOabMzJvdu6/UiuZezV6QK5dSlG84ov/aaiqXj38=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+SDo693bJlVdllGtEeKM=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/yaml.v2 v2.2.2 h1:ZCJp+EgiOT7lHqUV2J862kp8Qj64Jo6az82+3Td9dZw=
 gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=


stats/view/worker.go
```diff
@@ -233,6 +233,7 @@ func (w *worker) reportView(v *viewInternal, now time.Time) {
 		e.ExportView(viewData)
 	}
 	exportersMu.Unlock()
+	v.clearRows()
```
rghetia: This will affect sum/count and bucket count in Distribution. It will also affect count aggregation which is expected to be cumulative.

I understand the problem of min and max that you are trying to solve. Reseting just min and max might be the solution.

marwan-at-work: @rghetia Thanks for the review :) 

So I don't think it's just the min/max. I'm not sure about the sum/count in Distribution just yet, but for the Count-Aggregation, it should probably also be reset and let the Exporter be responsible for the cumulativeness side of things. 

The reason I say that is because if the server restarts for whatever reason, then the count is completely gone. Furthermore, if the Count is displayed on a timeseries graph, then a lot of these values will be falsely duplicated. 

For example, imagine a server receives 10 requests per second..after 10 seconds, the graph will say there are 100 requests in the last *second* (which is wrong, there were only 10) and if the serve restarts you will see a sharp drop from 100 back to 10. 

If we collect the "count per interval", then that should solve the issue. 

Would also love to know more about which data should never be reset, *even* if the process restarts and the data gets "forced" to be reset.

rghetia: Distributions and Counters have Start-time. Start-time lets backend detect the restart. 
It also lets it calculate the rate correctly. 
Example:
```
At time T0: Application starts 
T1: First request served. Start-time is set to T1
T2: 100 request served. 
T3: 200 request served. Export enabled.
T4: 300 request served. First export happens, Count = 300, Start-time = T1. Rate=300/(T4-T1)
T5: 400 request served.
T6: Application Restarts
T7: First request served after restart, Start-time= T7
T8: 100 request served. Export count = 100, Start-time = T7, Rate = 100/(T8-T7)
```
Resetting all metrics would be a behavior change and would have an adverse consequence on some backends.

marwan-at-work: @rghetia awesome. However, from what I can see there's no way for a backend to determine that: 

The CountData only has access to `Value` and on idea of start time: https://github.com/census-instrumentation/opencensus-go/blob/master/stats/view/aggregation_data.go#L42

In your example, the Exporter does not have access to `T1` (does it?), it can have access to `T0` by just putting `func init() { t1 = time.Now() }` I imagine but that's not accurate. 

Furthermore, dividing Count by (Tnow-T1) will just make an even distribution across time, and not the _real_ distribution. For example, if the first minute  we got 300 requests, but the next minute we got 10 requests, it will not show that accurately no? 

Finally, do you have an example of an exporter doing this time calculation? 

Thanks again :)

rghetia: startTime per view is stored [here](https://github.com/census-instrumentation/opencensus-go/blob/df6e2001952312404b06f5f6f03fcb4aec1648e5/stats/view/worker.go#L44)

There are two interface for exporters to get the data
1. [ExportView(vd *ViewData)](https://github.com/census-instrumentation/opencensus-go/blob/3b5a343282fe4b4fccdb0f24cbd1d7169d20858a/stats/view/export.go#L35)
starttime for ViewData is set [here](https://github.com/census-instrumentation/opencensus-go/blob/df6e2001952312404b06f5f6f03fcb4aec1648e5/stats/view/worker.go#L227)
2. [ExportMetric(context.Context, []*metricdata.Metric)](https://github.com/census-instrumentation/opencensus-go/blob/3b5a343282fe4b4fccdb0f24cbd1d7169d20858a/metric/metricexport/export.go#L25)
startTime for Metric is retrieved [here](https://github.com/census-instrumentation/opencensus-go/blob/3b5a343282fe4b4fccdb0f24cbd1d7169d20858a/stats/view/worker.go#L261)
and set [here](https://github.com/census-instrumentation/opencensus-go/blob/df6e2001952312404b06f5f6f03fcb4aec1648e5/stats/view/view_to_metric.go#L115)

Regarding even distribution - If backend has individual data points then it can correctly represent the rate for each time window. For example
T0: Count = 300
T1: Count = 400
T2: Count = 410
Rate for the window T1-T0 = (400-300)/(T1-T0)
Rate for the window T2-T1 = (410-400)/(T2-T1)

ViewData and Metric both have end time as well. So the backend can compute the rate accordingly. 

I'll post a screenshot shortly.

rghetia: image

marwan-at-work: @rghetia 

>If backend has individual data points then it can correctly represent the rate for each time window

What do you mean by individual data points? Does the Exporter have to keep track of "previous ExportData" metrics? 

If that's true, then an Exporter wouldn't even need to care about the time, it can just take the "last count value" and subtract that from the current value and that's the right number for every ExportView invocation, right? 

For example, 

T0: Process starts
T1: Count = 100 -- Real Count is  100 because (100-0 = 100)
T2: Count = 150 -- Real Count is 150-100 because (150-100 = 50)

Therefore, at T1 count is 100, and at T2 count is 50. 

Please let me know if my understanding is correct. 

Lastly, can you tell me which of the data here needs to be "reset" and which must stay the same? https://github.com/census-instrumentation/opencensus-go/blob/master/stats/view/aggregation_data.go#L118:L129

Thanks again 🎉

rghetia: > What do you mean by individual data points? Does the Exporter have to keep track of "previous ExportData" metrics?
> 
> If that's true, then an Exporter wouldn't even need to care about the time, it can just take the "last count value" and subtract that from the current value and that's the right number for every ExportView invocation, right?

Exporter doesn't keep any state. It simply sends the data to the backend. 
The start-time and end-time are both important from the backend perspective.
Start-time is used to detect reset. It is also used to estimate if initial data points were missed.
End-time is important to accurately compute the rate between two data points.
Example:

| Time | Value | Start Time | Exported | Rate | Comments |
| ----- | ----- | ---------- | --------- | --- | ---------- |
| T0+60s | 600 | T0 | No |  |  |
| T0+120s | 1200 | T0 | No |  |  |
| T0+180s | 1800 | T0 | Yes | 10/s  | Rate is still computed even though the first two data points are missing. Without start-time it would not be able to compute the rate |
| T0+240s | 2100 | T0 | Yes | 5/s  | Actual rate from time T0+180 to T0+240s  |
| T0+560s | 600 | T0+500 | Yes | 10/s  | Reset Detected T0 != T0+500 |
| T0+620s | 1200 | T0+500 | No | 10/s  | Lost for whatever reason |
| T0+680s |1800 | T0+500 | Yes | 10/s  | Rate is computed properly even though previous point was missing. Without the end-time it would be able to calculate the rate. Alternatively backend could record time on receiving data point but that would not be as accurate as the time at the observation point |

> 
> Lastly, can you tell me which of the data here needs to be "reset" and which must stay the same? https://github.com/census-instrumentation/opencensus-go/blob/master/stats/view/aggregation_data.go#L118:L129
> 
> I imagine the min/max definitely need to be reset, probably the Mean as well? But I'm not sure about the rest of the struct fields here.

you can only reset min/max.

marwan-at-work: > Reset Detected T0 != T0+500

That's interesting. However, from all the exporters I'm looking at, *none* of them even mention the StartTime let alone pass it back to their respective backends...Are all of these backends incorrect? More likely that I'm missing something here, but the ones I'm looking at are [Prometheus](https://github.com/census-ecosystem/opencensus-go-exporter-prometheus) and [Datadog](https://github.com/DataDog/opencensus-go-exporter-datadog). 

I'm also looking at the getting started guide [here](https://opencensus.io/quickstart/go/metrics/) and you can see that all the data goes up cumulatively, but if I reset the process, then the data drops and start over, so it's not automatically doing any time calculation the way you explained it above. Maybe I'm missing a configuration or a knob somewhere? 

On the other hand, I think we're talking about two different models: the OpenCensus pull model that you very kindly explained. But on the other hand, the model DataDog operates on is the [StatsD Model](https://github.com/b/statsd_spec) which from what I gather is quite trivial in that it only takes one number at a specific time  and that's the end of the story. Therefore, the only way to solve the problem is to actually make the Exporter keep **state**. At my company we have a contact with DataDog so I'll try to reach out to them about this. 

In any case, thank you @rghetia  for the thorough explanation 💯 💯

marwan-at-work: @rghetia I'll send a fix for min/max but before that I'd love to confirm that the exporters I mentioned above are indeed not working as intended. Thanks!

rghetia: yes prometheus doesn't handle the reset because it doesn't have start-time.
[stackdriver](https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver) does use the start-time and handles the reset.
datadog seems to convert everything to Gauge (except histograms). In Gauge there is no start-time.  
By definition Count() and Sum() aggregation are defined as cumulative so unfortunately it can't be reset. What you are asking for is a Delta() aggregator which provides delta between two reads.

marwan-at-work: > yes prometheus doesn't handle the reset because it doesn't have start-time.

This means that in Prometheus, the RATE is only accurate if the process never restarts. 

Furthermore, the COUNT is never accurate since I can't imagine prometheus doing the more complex Delta calculations on their end that we both described in the comments above (re-iterated here: checking if the previous value is smaller or bigger to determine if the process falsely restarted in order to aggregate the count accurately) 

> By definition Count() and Sum() aggregation are defined as cumulative

This is where there's a misunderstanding between specs and libraries. According to the DataDog [docs](https://docs.datadoghq.com/developers/metrics/metrics_type/?tab=count#metric-type-definition):

> ***The COUNT metric submission type represents the number of events that occur in a defined time interval, otherwise known as a flush interval. This number of events can accumulate or decrease over time – it is not monotonically increasing.***

That said, you did mention the following: 

> datadog seems to convert everything to Gauge (except histograms)

In fact, even histograms (DistributionData) are converted to a Gauge: https://github.com/DataDog/opencensus-go-exporter-datadog/blob/master/stats.go#L77:L105

I understand that the exporter is treating the OpenCensus Count() type more of a DataDog GAUGE type than a DataDog COUNT type. But it's still incorrect because it's always gonna be the number since the process started, and it gets even worse when you are aggregating over multiple tags where each one has a different start time.


### MEAN
if Count() and Sum() are cumulative, what about the `DistributionData.Mean` field? Right now the Mean() is being calculated over the entire process, and similar to Min/Max, there's absolutely no way to calculate/fix process-restarts...right? Hopefully I'm wrong.

### DELTA 

> What you are asking for is a Delta() aggregator which provides delta between two reads.

To be exact, what I'm asking for is a Delta() aggregator which provides delta between two reads *as long as the process does not restart*, and if the process restarts then it needs to apply a more complex calculation to get the right delta (detailed above). 

By the way, I'm not really asking for the above, it sounds **very** cumbersome. Nonetheless, I am just trying to make OpenCensus work with 2 out of the 3 most commonly used stats exporters which both seem to be completely broken given our conversation so far. 

From an emotional perspective, I'm frustrated that this disconnect has been sitting there this entire time without anyone noticing. Like, did anyone test this at all? OpenCensus has docs and tutorials on both Prometheus and DataDog which is a clear endorsement of these exporters yet both seem quite broken to me. 

@rghetia thanks again for keeping up with all of this. If you'd like to discuss this over a video call for more real time discussion, I'm happy to do so. Otherwise, this thread has been very helpful as well. 

From a practical side of things, I'd like to solve this issue and move on. I'm currently using a fork of OpenCensus that is pinned at the commit of this PR but that's just a temporary workaround. 

### Moving Forward: 

So, let's see where we can go from here: 

We currently have 3 problems: 

1. Min/Max: solved since you're okay with merging this type of fix. 
2. Count(): we have two possible solutions (feel free to suggest others): 
2A. Create a new Delta type in OpenCensus 
2B. Make the exporters keep state and get the delta. 
3. Mean(): Right now, I'm not sure how I can get the mean of a histogram say between 12pm to 1pm (uninfluenced by data before 12pm and also unadulterated by process restarts between 12pm to 1pm)

rghetia: google hangout or zoom works. I am on PST. Ping me on gitter with time and a link.

marwan-at-work: @rghetia I've updated the PR to reset min, max, mean (and also sqauredSumDev) as all of these values share the same problem we've discussed. 

Let me know if this looks good. 

Thanks again!

marwan-at-work: @rghetia now that I think about, resetting Mean doesn't work without resetting all the buckets. Right? 

I have another idea, why don't I roll back to my original PR, but instead of forcing the restart, we just have it as a dynamic configuration? Something like `view.RestartAfterFlush(true)` that will set a config in the worker.. 

This way, StackDriver and exporters relying on Monotonicity won't break, but DataDog and other StatsD based backends will work as expected. 

Let me know!

rghetia: Mean and squaredSumDev are used in stackdriver (at least they are  exported to the backend). 
view.RestartAfterFlush(true) (not sure about the name though) is better but it has following caveats
- It will not work with [Derived metrics](https://github.com/census-instrumentation/opencensus-go/blob/aad2c527c5defcf89b5afab7f37274304195a6b2/metric/registry.go#L200). Derived metrics are simply read when metrics are exported. 
- There can only be one Exporter if new interface (Reader or IntervalReader) is used.

keeping a state in the exporter would work for all cases.

marwan-at-work: >keeping a state in the exporter would work for all cases. 

Unfortunately, `DistributionData.Mean` cannot be reset without resetting `DistributionData.Count`, because the calculation looks like this in the code: `a.Mean = a.Mean + (v-a.Mean)/float64(a.Count)` -- in other words the count has to be reset or OpenCensus has to keep state of the previousCount itself. 

The only option I see that doesn't break StackDriver and others is to make the config an opt-in. The DataDog exporter doesn't use Derived metrics and the exporter can include in the documentation that it can only be used by itself...most people don't use multiple metrics exporters at the same time so that should be okay

 Thanks

rghetia: From the library perspective adding view.RestartAfterFlush seems like a hack. Are you only interested in reseting Distribution metrics after every export? If so, I am more inclined to add another aggregation type 'DistributionDelta' which will reset after every read. This would have to be added in the spec though and implemented across all languages.

I still think handling this in an exporter might be better at this point, especially if it is for distribution metrics only.

marwan-at-work: @rghetia personally I'm interested in both `view.DistributionData` as well as `view.CountData` -- but just as An FYI, `view.SumData` also falls under this category. 

The Delta for view.CountData can be maintained in the exporter, but the Delta for `view.DistributionData.Mean` (as well as squaredSumDev) cannot be maintained without resetting view.DistributionData.Count. 

>I still think handling this in an exporter might be better at this point, especially if it is for distribution metrics only.

Can you confirm that that's impossible without resetting the DistributionData.Count? 

If so, we only have two options: 

1. Create Delta types for both Count and DistributionData and implement them everywhere
2. Add `view.DisableMonotonicity(true)` option to the configuration with the 2 caveats documented for it. This might be fine especially if OpenTelemetry will resolve the Delta issue natively?

rghetia: > The Delta for view.CountData can be maintained in the exporter, but the Delta for `view.DistributionData.Mean` (as well as squaredSumDev) cannot be maintained without resetting view.DistributionData.Count.

The way to do it is maintain the prev state for entire distribution. So calculation would be
```
exportDist.Count = currDist.Count - prevDistCount
exportDist.Mean = (currDist.Count * currDist.Mean - prevDist.Count * prevDist.Mean)/exportDist.Count
exportDist.Bucket[0].Count = currDist.Bucket[0].Count - prevDist.Bucket[0].Count
```
we do similar thing (maintaining state) in prometheus receiver ([challenge #5](https://github.com/open-telemetry/opentelemetry-collector/tree/master/receiver/prometheusreceiver#challenges-and-solutions))

marwan-at-work: @rghetia ah, that sounds good. 

This is more of a question: how does the exporter keep track of previous data? Can it do it by each tag key-value individually? Or by the combination of each `*view.Row.Tags`?? 

I notice that the number of rows that are given to ExportView can vary for different reasons. 

Thank you!

rghetia: by combination of each view.Row.Tags. In prometheus receiver it does it by metric-name+sorted label-values(tag-values).

marwan-at-work: @rghetia 

1. I'd love to see an example of that, I'll send you what I have soon as well. Thank you :pray: 
2. Is the ExportView expected to be called synchronously or concurrently? Just wondering if I should be guarding against checking-updating previous count/distribution data. 

Thanks again

marwan-at-work: Also, I just reset the PR back to only resetting min/max :)

marwan-at-work: Lastly, I'm not sure how we can calculate the difference in `SumOfSquaredDev` 🤔

rghetia: Example:

| metric | Labels | Count | Sum | Start-time| Delta Count | Delta Sum | comment |
|-------|--------|-------|------|-------------|------------|----------|-----------|
| M1 | K1=V11, K2=V21 | 10 | 100 | T0 | 0 | 0 | first report ignore |
| M1 | K1=V11, K2=V22 | 20 | 150 | T0 | 0 | 0 | first report ignore |
| M1 | K1=V11, K2=V11 | 20 | 200 | T0 | 10 | 100 | second report take delta |
| M1 | K1=V11, K2=V22 | 35 | 300 | T0 | 15 | 150 | first report take delta |
| M1 | K1=V11, K2=V11 | 10 | 210 | T1 | 10 | 210 | Start-time changed, so either ignore it or report the value as a delta  |
| M1 | K1=V11, K2=V22 | 15 | 150 | T1 | 15 | 150 | Start-time changed, so either ignore it or report the value as a delta |

SumOfSquareDev cannot be calculated without the individual updates known.

marwan-at-work: @rghetia thank you! 

> SumOfSquareDev cannot be calculated without the individual updates known

So in this case, min/max are restarted, count/sum/avg are restarted, and SumOfSquaredDev is an unsolvable bug. This means Exporters should probably be aware of those details to export the metrics accurately. 

Would you prefer the above over `view.DisableMonotonicity(true)` configuration? 

I'm happy either way. 

Thanks!

rghetia: I prefer 
> So in this case, **min/max** are _restarted_, **count/sum/avg** are _diffed_, and SumOfSquaredDev is an unsolvable bug. This means Exporters should probably be aware of those details to export the metrics accurately.

over
> Would you prefer the above over view.DisableMonotonicity(true) configuration?

because, the later changes the meaning of aggregation as currently defined. It would be sort of a hack.

marwan-at-work: @rghetia sounds good, then feel free to review/merge this pr :+1:

stats/view/worker.go
```diff
@@ -233,6 +233,7 @@ func (w *worker) reportView(v *viewInternal, now time.Time) {
 		e.ExportView(viewData)
 	}
 	exportersMu.Unlock()
+	v.resetValues()
```
rghetia: 1. This should be also called from w.Read() method. This is for new mechanism (ExportMetrics)
2. Apart from that reportUsage() should not go through all the views unless there is an exporter registered (used for old mechanism ExportView())
If user configures both ways or multiple exporter using new mechanism then values of min/max is unpredictable.
3. Min/Max doc should be updated to reflect the behavior.
4. Add tests for min/max.

stats/view/worker.go
marwan-at-work: @rghetia there's a potential new issue: how can we ensure that ExportView does _not_ get called if a `stats.Record()` was never called after the last interval? 

AFAIK, in the StatsD model, you only report to the backend when there's data. But OpenCensus will infinitely call ExportData even if the code hasn't called stats.Record since the last ReportingPeriod...right?

stats/view/worker.go
marwan-at-work: @rghetia done, let me know if the last commit is what you were looking for :+1:

stats/view/worker.go
marwan-at-work: In the original solution, `clearRows` makes sure that there are no rows, and therefore no new data is reported.

stats/view/worker_test.go
```diff
@@ -494,7 +567,11 @@ func (e *countExporter) ExportView(vd *Data) {
 	if len(vd.Rows) == 0 {
 		return
 	}
-	d := vd.Rows[0].Data.(*CountData)
+	d, ok := vd.Rows[0].Data.(*CountData)
+	if !ok {
+		debug.PrintStack()
+		panic("BYE")
```
rghetia: Is this necessary?

stats/view/worker.go
rghetia: test is good.
Item 2 above is not taken care of.

There should be new option in ReadAndExport method. If the option to reset min/max is specified then only do the reset.

Similarly, for old approach (export viewdata) there should be a new method view.ResetMinMaxOnExport() - It is rather a long name but I don't have a good suggestion. Alternatively, provide reset functionality only using new approach with ReadAndExport().

stats/view/worker.go
AlexandreYang: > 1. Create Delta types for both Count [and Sum?] and DistributionData and implement them everywhere 

I'm not sure if that option has been discarded or not. Wouldn't this solve all issues if implemented ?

PR #1185, opened 2019-12-03
Added test exporter for use in unit tests.

jkohen:
With this exporter one can write unit tests to verify that the instrumentation is working. See the included code example.

### DIFF
--- a/metric/test/exporter.go
+++ b/metric/test/exporter.go
@@ -56,6 +56,7 @@ func (e *Exporter) ReadAndExport() {
 	e.metricReader.ReadAndExport(e)
 }
 
+// String defines the ``native'' format for the exporter.
 func (e *Exporter) String() string {
 	return fmt.Sprintf("points{%v}", e.points)
 }
@@ -79,6 +80,9 @@ func labelMapToKeyValue(labels map[string]string) []keyValue {
 }
 
 func labelObjectsToKeyValue(keys []metricdata.LabelKey, values []metricdata.LabelValue) []keyValue {
+	if len(keys) != len(values) {
+		panic("keys and values must have the same length")
+	}
 	kv := make([]keyValue, 0, len(values))
 	for i := range keys {
 		if values[i].Present {
--- a/metric/test/exporter.go
+++ b/metric/test/exporter.go
@@ -1,0 +1,98 @@
+package test
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"go.opencensus.io/metric/metricdata"
+	"go.opencensus.io/metric/metricexport"
+	"go.opencensus.io/stats/view"
+)
+
+// Exporter keeps exported metric data in memory to aid in testing the instrumentation.
+//
+// Metrics can be retrieved with `GetPoint()`. In order to deterministically retrieve the most recent values, you must first invoke `ReadAndExport()`.
+type Exporter struct {
+	// points is a map from a label signature to the latest value for the time series represented by the signature.
+	// Use function `labelSignature` to get a signature from a `metricdata.Metric`.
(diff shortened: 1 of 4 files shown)


metric/test/exporter.go
```diff
@@ -0,0 +1,100 @@
...
+func (e *Exporter) ReadAndExport() {
+	// The next line forces the view worker to process all stats.Record* calls that
+	// happened within Store() before the call to ReadAndExport below. This abuses the
+	// worker implementation to work around lack of synchronization.
+	// TODO(jkohen,rghetia): figure out a clean way to make this deterministic.
+	view.SetReportingPeriod(time.Minute)
+	e.metricReader.ReadAndExport(e)
+}
+
+func (e *Exporter) String() string {
```
rghetia: missing doc.

metric/test/exporter_test.go
```diff
@@ -0,0 +1,64 @@
...
+		// The code under test begins here.
+		stats.RecordWithTags(context.Background(),
+			[]tag.Mutator{tag.Upsert(myTag, "label1")},
+			myMetric.M(int64(i)))
+		// The code under test ends here.
+
+		metrics.ReadAndExport()
+		metricValue := getCounter(metrics, myMetric.Name(), newMetricKey("label1"))
+		fmt.Printf("increased by %d\n", metricValue-metricBase)
+	}
```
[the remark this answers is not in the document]
jkohen: `metricBase` is needed to isolate test runs. The metric state is global, so if there are two tests that refer to the same metric, a test that looks for absolute values will fail. I confirmed that by removing `metricBase`, then creating a copy of the example. The output was:
```
 % go test -v go.opencensus.io/metric/test
=== RUN   ExampleExporter
--- PASS: ExampleExporter (0.00s)
=== RUN   ExampleExporter_2
--- FAIL: ExampleExporter_2 (0.00s)
got:
increased by 7
increased by 9
increased by 12
want:
increased by 1
increased by 3
increased by 6
FAIL
FAIL    go.opencensus.io/metric/test    0.005s
```

metric/test/exporter.go
jkohen: Done.

metric/test/exporter.go
```diff
@@ -0,0 +1,100 @@
...
+		kv = append(kv, keyValue{Key: k, Value: v})
+	}
+	sortKeyValue(kv)
+	return kv
+}
+
+func labelObjectsToKeyValue(keys []metricdata.LabelKey, values []metricdata.LabelValue) []keyValue {
+	kv := make([]keyValue, 0, len(values))
+	for i := range keys {
+		if values[i].Present {
```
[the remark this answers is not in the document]
jkohen: Done. Is `panic` OK?

metric/test/exporter_test.go
```diff
@@ -50,6 +52,35 @@ func ExampleExporter() {
...
+
+func ExampleExporter_metric() {
+	metricReader := metricexport.NewReader()
+	metrics := NewExporter(metricReader)
+	r := metric.NewRegistry()
+	g, _ := r.AddInt64DerivedCumulative("derived", metric.WithLabelKeys(myTag.Name()))
+	for i := 1; i <= 3; i++ {
+		// The code under test begins here.
+		m := derivedMetric{int64(i)}
+		g.UpsertEntry(m.ToInt64, metricdata.NewLabelValue("l1"))
```
rghetia: proper example of derived metrics would be move line 70 and 71 outside of the for loop and then updating `m.i = i` in the for loop.

jkohen: Thanks for the tip, I like that! I suspected that I wasn't using the API as intended. I'm glad to have this as a reference of how this is intended to be used :)

PR #1196, opened 2020-02-02
Allow creating additional View universes.

### LINKED ISSUE #1195: Support exporting different stats to different exporters.
I'm writing a multi-tenant backend component in Go using the opencensus libraries. I'd like to be able to have separate exporters for the metrics for each resource managed by the multi-tenant service, so that I can (for example) export the stats via OpenCensus with different [Resource](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/agent/metrics/v1/metrics_service.proto#L54) attributes for each managed resource.

Right now, all the stats aggregation code in `stats/view` tunnels through a set of singleton resources (`defaultWorker` and `exporters`, mostly), which means that statistics are always aggregated only by `tags.Tag` values, and not segregated by Resource.

For advanced use cases, I'd like a mode where I can maintain multiple Views, one for each Resource, with separate aggregation.

We could make this separation by Resource the default, and extend `stats.Record` and `view` to also separate by Resource fields as well as by tag. This seems like a more complex change to the library, so I'm inclined to defer it to the OpenTelemetry library, and leave exporting for multiple resources as an advanced use case.

cc @anniefu

evankanderson:
Fixes #1195

I wanted to be able to export statistics for multiple Resources, but I discovered that all the stats aggregation was bundled into a single "universe" in the stats/view `defaultWorker`. This PR moves all the singleton state into a `Worker` interface, and maintains the simple singleton interface for the common case, but allows creating additional `Workers` in complex cases.

### DIFF
--- a/stats/record.go
+++ b/stats/record.go
@@ -31,19 +31,17 @@ func init() {
 	}
 }
 
-// ResolvedOptions can be used to extract the current tags and measurements
-// from context and stats arguments when using custom workers to export stats
-// to a separate exporter.
-type ResolvedOptions struct {
-	Attachments metricdata.Attachments
-	Tags        *tag.Map
-	Measures    []Measurement
+// Recorder is a subset of the view.Meter interface which only includes
+// the Record method, to avoid circular imports between stats and view.
+type Recorder interface {
+	Record(*tag.Map, interface{}, map[string]interface{})
 }
 
 type recordOptions struct {
 	attachments  metricdata.Attachments
 	mutators     []tag.Mutator
 	measurements []Measurement
+	recorder     Recorder
 }
 
 // WithAttachments applies provided exemplar attachments.
@@ -64,6 +62,14 @@ func WithTags(mutators ...tag.Mutator) Options {
 func WithMeasurements(measurements ...Measurement) Options {
 	return func(ro *recordOptions) {
 		ro.measurements = measurements
+	}
+}
+
+// WithMeter records the measurements to the specified `view.Meter`, rather
+// than to the global metrics recorder.
+func WithMeter(meter Recorder) Options {
+	return func(ro *recordOptions) {
+		ro.recorder = meter
 	}
 }
 
@@ -93,23 +99,6 @@ func RecordWithTags(ctx context.Context, mutators []tag.Mutator, ms ...Measureme
 	return RecordWithOptions(ctx, WithTags(mutators...), WithMeasurements(ms...))
 }
 
-// ResolveOptions determines the full set of Tags, Measurements, etc from the
-// provided Options and context.Context. This is mostly useful when using
-// multiple exporters.
-func ResolveOptions(ctx context.Context, ros ...Options) (*ResolvedOptions, error) {
-	o := createRecordOption(ros...)
-
-	if len(o.mutators) > 0 {
-		var err error
-		if ctx, err = tag.New(ctx, o.mutators...); err != nil {
-			return nil, err
-		}
-	}
-	return &ResolvedOptions{Tags: tag.FromContext(ctx),
-		Measures:    o.measurements,
-		Attachments: o.attachments}, nil
-}
-
 // RecordWithOptions records measurements from the given options (if any) against context
 // and tags and attachments in the options (if any).
 // If there are any tags in the context, measurements will be tagged with them.
@@ -118,9 +107,10 @@ func RecordWithOptions(ctx context.Context, ros ...Options) error {
 	if len(o.measurements) == 0 {
 		return nil
 	}
-	// This could use ResolveOptions, but it does additional work to
-	// short-circuit if there are no metrics that need to be exported.
 	recorder := internal.DefaultRecorder
+	if o.recorder != nil {
+		recorder = o.recorder.Record
+	}
 	if recorder == nil {
 		return nil
 	}
--- a/stats/record.go
+++ b/stats/record.go
@@ -31,6 +31,15 @@ func init() {
 	}
 }
 
+// ResolvedOptions can be used to extract the current tags and measurements
+// from context and stats arguments when using custom workers to export stats
+// to a separate exporter.
+type ResolvedOptions struct {
+	Attachments metricdata.Attachments
+	Tags        *tag.Map
+	Measures    []Measurement
+}
+
 type recordOptions struct {
 	attachments  metricdata.Attachments
 	mutators     []tag.Mutator
@@ -84,6 +93,23 @@ func RecordWithTags(ctx context.Context, mutators []tag.Mutator, ms ...Measureme
 	return RecordWithOptions(ctx, WithTags(mutators...), WithMeasurements(ms...))
 }
 
+// ResolveOptions determines the full set of Tags, Measurements, etc from the
+// provided Options and context.Context. This is mostly useful when using
+// multiple exporters.
+func ResolveOptions(ctx context.Context, ros ...Options) (*ResolvedOptions, error) {
+	o := createRecordOption(ros...)
+
+	if len(o.mutators) > 0 {
+		var err error
+		if ctx, err = tag.New(ctx, o.mutators...); err != nil {
+			return nil, err
+		}
+	}
+	return &ResolvedOptions{Tags: tag.FromContext(ctx),
+		Measures:    o.measurements,
+		Attachments: o.attachments}, nil
+}
+
 // RecordWithOptions records measurements from the given options (if any) against context
 // and tags and attachments in the options (if any).
 // If there are any tags in the context, measurements will be tagged with them.
@@ -92,6 +118,8 @@ func RecordWithOptions(ctx context.Context, ros ...Options) error {
 	if len(o.measurements) == 0 {
 		return nil
 	}
+	// This could use ResolveOptions, but it does additional work to
+	// short-circuit if there are no metrics that need to be exported.
 	recorder := internal.DefaultRecorder
 	if recorder == nil {
 		return nil
--- a/stats/record.go
+++ b/stats/record.go
@@ -31,9 +31,11 @@ func init() {
 	}
 }
 
-// Recorder is a subset of the view.Meter interface which only includes
-// the Record method, to avoid circular imports between stats and view.
+// Recorder provides an interface for exporting measurement information from
+// the static Record method by using the WithRecorder option.
 type Recorder interface {
+	// Record records a set of measurements associated with the given tags and attachments.
+	// The second argument is a `[]Measurement`.
 	Record(*tag.Map, interface{}, map[string]interface{})
 }
 
@@ -65,9 +67,9 @@ func WithMeasurements(measurements ...Measurement) Options {
 	}
 }
 
-// WithMeter records the measurements to the specified `view.Meter`, rather
+// WithRecorder records the measurements to the specified `Recorder`, rather
 // than to the global metrics recorder.
-func WithMeter(meter Recorder) Options {
+func WithRecorder(meter Recorder) Options {
 	return func(ro *recordOptions) {
 		ro.recorder = meter
 	}
--- a/stats/record_test.go
+++ b/stats/record_test.go
@@ -95,8 +95,8 @@ func cmpExemplar(got, want *metricdata.Exemplar) string {
 	return cmp.Diff(got, want, cmpopts.IgnoreFields(metricdata.Exemplar{}, "Timestamp"), cmpopts.IgnoreUnexported(metricdata.Exemplar{}))
 }
 
-func TestResolveOptions(t *testing.T) {
-	meter := view.NewWorker()
+func TestRecordWithMeter(t *testing.T) {
+	meter := view.NewMeter()
 	meter.Start()
 	k1 := tag.MustNewKey("k1")
 	k2 := tag.MustNewKey("k2")
@@ -128,7 +128,7 @@ func TestResolveOptions(t *testing.T) {
 		stats.WithTags(tag.Upsert(k1, "bar"), tag.Insert(k2, "bar")),
 		stats.WithAttachments(attachments),
 		stats.WithMeasurements(m1.M(12), m1.M(6), m2.M(5)),
-		stats.WithMeter(meter))
+		stats.WithRecorder(meter))
 	if err != nil {
 		t.Fatalf("Failed to resolve data point: %v", err)
 	}
--- a/stats/record_test.go
+++ b/stats/record_test.go
@@ -98,6 +98,7 @@ func cmpExemplar(got, want *metricdata.Exemplar) string {
 func TestRecordWithMeter(t *testing.T) {
 	meter := view.NewMeter()
 	meter.Start()
+	defer meter.Stop()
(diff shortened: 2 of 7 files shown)


evankanderson:
CC @anniefu

evankanderson:
I added an additional interface for extracting the `stats.Option` settings for passing to the `view.Record()` method. This seemed the simplest way to extract the logic in `stats.RecordWithOptions`, which currently consumes the context and performs some short-circuiting logic.

stats/view/worker.go
```diff
@@ -47,32 +47,71 @@ type worker struct {
...
+
+// Worker defines an interface which allows a single process to maintain
+// multiple sets of metrics exports (intended for the advanced case where a
+// single process wants to report metrics about multiple objects, such as
+// multiple databases or HTTP services).
+//
+// Note that this is an advanced use case, and the static functions in this
+// module should cover the common use cases.
+type Worker interface {
+	Record(tags *tag.Map, ms []stats.Measurement, attachments map[string]interface{})
```
rghetia: Add doc for each method.

stats/view/worker.go
```diff
@@ -143,9 +213,15 @@ func newWorker() *worker {
 		c:          make(chan command, 1024),
 		quit:       make(chan bool),
 		done:       make(chan bool),
+
+		exporters: make(map[Exporter]struct{}),
 	}
 }
 
+func (w *worker) Start() {
+	go w.start()
```
rghetia: There is another way to export data instead of exporting ViewData. This is [exporting metrics](https://github.com/census-instrumentation/opencensus-go/tree/master/metric/metricexport)  using Reader and IntervalReader. The Reader/IntervalReader reads metrics from all producer. Worker here is a producer. The existing w.start() method adds the worker (w) as producer. This means all workers will be producer and metrics from each producer will be available to the exporter using [ExportMetrics](https://github.com/census-instrumentation/opencensus-go/blob/d835ff86be02193d324330acdb7d65546b05f814/metric/metricexport/export.go#L25) interface. You are probably not using that interface. Otherwise this will not work.
Alternatively, you could register a wrapper exporter that creates SD exporter with specific resource. The wrapper exporter filters the time-series based on labels. It may create extra processing overhead but will limit the code change in knative.

stats/record.go
```diff
@@ -84,6 +93,23 @@ func RecordWithTags(ctx context.Context, mutators []tag.Mutator, ms ...Measureme
 	return RecordWithOptions(ctx, WithTags(mutators...), WithMeasurements(ms...))
 }
 
+// ResolveOptions determines the full set of Tags, Measurements, etc from the
+// provided Options and context.Context. This is mostly useful when using
+// multiple exporters.
+func ResolveOptions(ctx context.Context, ros ...Options) (*ResolvedOptions, error) {
```
rghetia: Not sure how this will be used. Can you provide an example?

stats/view/worker.go
evankanderson: It looks like using ExportMetrics requires tracking and aggregating all the metrics in something (which would basically be `views`).

Another option here would be to [add an additional internal key](https://github.com/census-instrumentation/opencensus-go/blob/master/stats/view/worker_commands.go#L164) to `worker.measure` that keyed off the Resource. Unfortunately, that looks to require some additional surgery to `encodeWithKeys` and other parts of the general code path that I was hoping to avoid.

stats/view/worker.go
```diff
@@ -47,32 +47,71 @@ type worker struct {
...
+}
+
+// Worker defines an interface which allows a single process to maintain
+// multiple sets of metrics exports (intended for the advanced case where a
+// single process wants to report metrics about multiple objects, such as
+// multiple databases or HTTP services).
+//
+// Note that this is an advanced use case, and the static functions in this
+// module should cover the common use cases.
+type Worker interface {
```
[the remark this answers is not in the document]
evankanderson: Sounds good, happy to do so.

stats/view/worker.go
```diff
@@ -89,31 +136,42 @@ func Unregister(views ...*View) {
...
 
 // RetrieveData gets a snapshot of the data collected for the the view registered
 // with the given name. It is intended for testing only.
 func RetrieveData(viewName string) ([]*Row, error) {
+	return defaultWorker.RetrieveData(viewName)
+}
+
+// RetrieveData gets a snapshot of the data collected for the the view registered
+// with the given name. It is intended for testing only.
+func (w *worker) RetrieveData(viewName string) ([]*Row, error) {
```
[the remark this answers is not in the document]
evankanderson: Done.

stats/view/worker.go
evankanderson: Done.

stats/record.go
evankanderson: I want to be able to dispatch to different Meter interfaces based on the underlying resource. Unfortunately, the mechanism for this today is:

    stats.RecordWithOptions(ctx, stats.WithMeasurements(...), stats.WithTags(...), ...)

It's hard to extract the tags from the `stats.Options` because the mutation functions act on a `*createRecordOption`, which is non-exported and doesn't have an interface. Additionally, Meter is in the `view` package, while RecordWithOptions is in the `stats` package, so there's no good way to smuggle the data across the package boundary.

rghetia: This will make the Tags readable, which defeats the original purpose of it being only readable by the library.
Let me understand this. The flow is 
- Call the ResolveOptions()
- Determine resource from the Tags. Assuming here, that you need all tags (tags passed to ResolveOptions and tags present in the context prior to calling ResolveOptions).
- Determine the Meter/Exporter based on the resource.
- Record the value against the Meter.

evankanderson: I'm not sure I understand -- is this about making the `stats.Options` not readable outside the library?

The other option (which may be what you're suggesting) is to either add a `WithMeter` option or a `RecordToMeter` method here which would use the logic internally.

Would you like me to take the `WithMeter` stats option for a spin? I hadn't considered it until you pushed a bit -- thanks!

rghetia: > Would you like me to take the WithMeter stats option for a spin? I hadn't considered it until you pushed a bit -- thanks!

Sure.

evankanderson: Done. I think I like this better, but I had to re-export `RetrieveData` from the `Meter` interface for testing.

stats/record_test.go
```diff
@@ -93,3 +94,107 @@ func TestRecordWithAttachments(t *testing.T) {
 func cmpExemplar(got, want *metricdata.Exemplar) string {
 	return cmp.Diff(got, want, cmpopts.IgnoreFields(metricdata.Exemplar{}, "Timestamp"), cmpopts.IgnoreUnexported(metricdata.Exemplar{}))
 }
+
+func TestResolveOptions(t *testing.T) {
```
rghetia: please rename the function to TestRecordWithOptions

stats/view/worker.go
```diff
@@ -47,32 +47,101 @@ type worker struct {
...
+// Meter defines an interface which allows a single process to maintain
+// multiple sets of metrics exports (intended for the advanced case where a
+// single process wants to report metrics about multiple objects, such as
+// multiple databases or HTTP services).
+//
+// Note that this is an advanced use case, and the static functions in this
+// module should cover the common use cases.
+type Meter interface {
+	// Record records a set of measurements ms associated with the given tags and attachments.
+	Record(tags *tag.Map, ms interface{}, attachments map[string]interface{})
```
rghetia:
```suggestion
	stats.Recorder
```

stats/record.go
```diff
@@ -31,10 +31,17 @@ func init() {
 	}
 }
 
+// Recorder is a subset of the view.Meter interface which only includes
+// the Record method, to avoid circular imports between stats and view.
```
rghetia: keep this definition and simply include this in the Meter.

stats/record.go
```diff
@@ -58,6 +65,14 @@ func WithMeasurements(measurements ...Measurement) Options {
 	}
 }
 
+// WithMeter records the measurements to the specified `view.Meter`, rather
+// than to the global metrics recorder.
+func WithMeter(meter Recorder) Options {
```
rghetia: s/WithMeter/WithRecorder/

stats/view/worker.go
```diff
@@ -124,17 +212,29 @@ func record(tags *tag.Map, ms interface{}, attachments map[string]interface{}) {
...
 		c: make(chan bool),
 	}
-	defaultWorker.c <- req
+	w.c <- req
 	<-req.c // don't return until the timer is set to the new duration.
 }
 
-func newWorker() *worker {
+// NewWorker constructs a
+func NewWorker() Meter {
```
rghetia: s/NewWorker/NewMeter/

stats/record.go
rghetia: Performance of RecordWithOptions WithMeter is 2/3rd of RecordWithOptions without meter
```
BenchmarkRecord8WithMeter-8      	  241356	       270 ns/op	     400 B/op	       4 allocs/op
BenchmarkRecord8WithoutMeter-8   	  335601	       197 ns/op	     368 B/op	       3 allocs/op
```

evankanderson: Using the following, I get 231 ns/op, rather than 270 ns/op:

```go
func BenchmarkRecord8_WithRecorder(b *testing.B) {
	ctx := context.Background()
	meter := view.NewMeter()
	meter.Start()
	defer meter.Stop()
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		stats.RecordWithOptions(ctx, stats.WithRecorder(meter), stats.WithMeasurements(m.M(1), m.M(1), m.M(1), m.M(1), m.M(1), m.M(1), m.M(1), m.M(1)))
	}

	b.StopTimer()
}
```

For the _Parallel case, I get 117 ns/op vs 108 ns/op:

```
goos: windows
goarch: amd64
pkg: go.opencensus.io/stats
BenchmarkRecord0-16                             12499986                98.0 ns/op
BenchmarkRecord1-16                              8571868               139 ns/op
BenchmarkRecord8-16                              6417102               188 ns/op
BenchmarkRecord8_WithRecorder-16                 5194611               231 ns/op
BenchmarkRecord8_WithRecorder_Parallel-16       10085448               117 ns/op
BenchmarkRecord8_Parallel-16                    11215130               108 ns/op
BenchmarkRecord8_8Tags-16                        5940622               198 ns/op
```

I can avoid the extra allocation and make the time between BenchmarkRecord8_WithRecorder and BenchmarkRecord8 match by pre-creating the `stats.WithRecorder(meter)` object like so:

```go
rec := stats.WithRecorder(meter)

for (....) {
  stats.RecordWithOptions(ctx, rec, stats.WithMeasurements(....)
}
```

I note that the current benchmark doesn't perform any aggregation, so it's actually testing the shortcut path AFAICT. Adding a view which is tracking a simple Sum (no tags) increases the benchmark time like so:

```
BenchmarkRecord8_Aggregated-16                   1000000              1017 ns/op
```

Note that using a view here also affects any subsequent benchmarks for the same measurement, because the `m.desc.subscribed()` check around line 120 of record.go will always return true once it has been registered once.

stats/record_test.go
evankanderson: Done.

stats/record.go
evankanderson: Done.

stats/record.go
evankanderson: Done. I updated the comment, too.

What do you think of changing the second argument to Record to be `[]stats.Measurement` rather than the current `interface{}`?

stats/view/worker.go
evankanderson: Done, thanks.

stats/view/worker.go
evankanderson: Done, thanks!

stats/record.go
evankanderson: Thanks, done.

stats/record.go
evankanderson: Done.

stats/record.go
rghetia: > Done. I updated the comment, too.
> 
> What do you think of changing the second argument to Record to be `[]stats.Measurement` rather than the current `interface{}`?



> Done. I updated the comment, too.
> 
> What do you think of changing the second argument to Record to be `[]stats.Measurement` rather than the current `interface{}`?

that is better.

stats/record.go
rghetia: BenchmarkRecord8_WithRecorder and BenchmarkRecord8 match by pre-creating the `stats.WithRecorder(meter)` object like so:
> 
> ```go
> rec := stats.WithRecorder(meter)
> 
> for (....) {
>   stats.RecordWithOptions(ctx, rec, stats.WithMeasurements(....)
> }
> ```
> 

+1.

stats/record.go
evankanderson: It turns out that this introduces an import cycle:

    import cycle not allowed
     package go.opencensus.io/stats/view
    	imports go.opencensus.io/stats
    	imports go.opencensus.io/stats/internal
    	imports go.opencensus.io/stats

Would you prefer that I move internal.DefaultRecorder out of `internal`, or do something like create an InternalMeasurement struct or interface and embed it in stats.Measurement?

rghetia: I would say leave it as interface.

evankanderson: It turns out that `stats.Measurement.Measure()` returns a stats.Measure, which seemed like a bridge too far in terms of putting things into Internal.

Take a look at 6003f86 for what that change would look like. I'm slightly concerned about moving DefaultRecorder from the internal package (and making it more public/obvious).

What do you think?

evankanderson: Reverted this to stick to interfaces.

evankanderson:
(Let me know if you want me to squash the commits or if there are other changes I need to make.)

rghetia:
> (Let me know if you want me to squash the commits or if there are other changes I need to make.)

@evankanderson 
Leave it as is. When I merge, I will squash it.

@songy23 would have time to review this? I would like to have one more reviewer.

evankanderson:
I have code like that, too.

Worse than that, I have code which auto-detects which type of resource it is, and has a fixed schema for each Resource of which tags should be extracted and removed from the tag map.

I suppose that I could still go through tags for aggregation and add a "RegisterType" method into my exports which would allow dynamically registering the schemas, but misdirecting all the aggregation through multiple layers of Tags and view aggregations feels wrong to me.

rghetia:
> The high level idea makes sense to me. The only alternative I can think of is to hard-code resources as metric labels, export to Collector, then have some sort of metric processor on Collector to reconstruct resources and strip those labels. For example, in addition to normal metric labels, add labels ["container_name": "some container", "pod_name": "pod"], then on the Collector remove those labels and attach a new k8s resource with these values.
> 
> We have internal use cases like this, but I'm don't think that's ideal since it requires lots of hard code and also mixes resources and metrics in a confusing way.

I agree that it would be confusing. Even though OC will be deprecated once Otel is released, this PR probably makes more sense than the alternate solution. It is also inline with concept of Meter in Otel. Do you agree?

PR #1204, opened 2020-03-30
Reduce allocations

boekkooi-fresh:
Hello,
First of thank you for taking the time to view this PR and your work on opencensus!

While trying to find a performance issue on a service I noticed that `Span.lruAttributesToAttributeMap` was allocating a bit more often then I was expecting. 
To reduce the amount of allocation I now provide the length of the source slice/map as the capacity as the new map.slice.

### DIFF
--- a/trace/lrumap.go
+++ b/trace/lrumap.go
@@ -44,7 +44,7 @@ func (lm lruMap) len() int {
 }
 
 func (lm lruMap) keys() []interface{} {
-	keys := []interface{}{}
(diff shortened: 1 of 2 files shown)


boekkooi-fresh:
@googlebot I signed it!

rghetia:
@boekkooi-fresh thank you for finding the issue and fixing it.

boekkooi-fresh:
No problem :smile: and thanks for the quick merge.

Issue #1188, opened 2019-12-13
Runmetrics: Expand to Include Goroutine & GC Stats


dm03514:
**Is your feature request related to a problem? Please describe.**
Greetings! I just started using opencensus in go.  The quality and framework was super easy to use, the primitives are intuitive and there are great docs! One thing I noticed was that runtime (`runmetrics`) metrics didn't include goroutine information or GC information.  I'm wondering if `runmetrics` are a good place to include these metrics? 

**Describe the solution you'd like**
If `runmetrics` is a good place to include these metrics, I was hoping to have [go routine and gc stats](https://github.com/prometheus/client_golang/blob/master/prometheus/go_collector.go#L315) included in the runtime metrics exported.

**Describe alternatives you've considered**
I could pretty easily instrument these in my own application, even package it as a library so all my go apps could easily get these metrics. 

**Additional context**

This request is based on the [runtime metrics that prometheus go client ships with](https://github.com/prometheus/client_golang/blob/master/prometheus/go_collector.go#L315).

Can see the prometheus metrics in action on their grafana dashboard:

![go_processes](https://user-images.githubusercontent.com/321963/70834608-c7bdf080-1dc8-11ea-9be4-455d0c0571f2.png)

-----

Thank you!!!

dm03514:
If you're open to this, I'd be happy to contribute it

rghetia:
> If you're open to this, I'd be happy to contribute it (will try and tackle it next week)

sure.

hougaardj:
Any idea when/if the pull request will be merged? The GC metrics are very useful to have.

punya:
@hougaardj I just merged the PR. Are you able to test it out from the master branch, or do you need a tagged RC?

hougaardj:
I have tested it from master branch and i got the additional metrics. I do need a tagged release to start using it proper.

hougaardj:
@punya Can we create a release?

punya:
@hougaardj we just published [v0.23.0](https://pkg.go.dev/go.opencensus.io@v0.23.0) ([release notes](https://github.com/census-instrumentation/opencensus-go/releases/tag/v0.23.0)).

PR #1207, opened 2020-04-07
Adding GC stats to runmetrics plugin

TrevinTeacutter:
Solves (hopefully): https://github.com/census-instrumentation/opencensus-go/issues/1188

### DIFF
--- a/plugin/runmetrics/producer.go
+++ b/plugin/runmetrics/producer.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"runtime"
 	"sync"
+	"time"
 
 	"go.opencensus.io/metric"
 	"go.opencensus.io/metric/metricdata"
@@ -52,6 +53,15 @@ type (
 		stackMSpanSys    *metric.Int64GaugeEntry
 		stackMCacheInuse *metric.Int64GaugeEntry
 		stackMCacheSys   *metric.Int64GaugeEntry
+
+		otherSys      *metric.Int64GaugeEntry
+		gcSys         *metric.Int64GaugeEntry
+		numGC         *metric.Int64GaugeEntry
+		numForcedGC   *metric.Int64GaugeEntry
+		nextGC        *metric.Int64GaugeEntry
+		lastGC        *metric.Int64GaugeEntry
+		pauseTotalNs  *metric.Int64GaugeEntry
+		gcCPUFraction *metric.Float64Entry
 	}
 
 	cpuStats struct {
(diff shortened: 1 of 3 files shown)


TrevinTeacutter:
@googlebot I signed it!

plugin/runmetrics/producer.go
```diff
@@ -219,6 +229,52 @@ func newMemStats(producer *producer) (*memStats, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	memStats.nextGC, err = producer.createInt64GaugeEntry("process/next_gc", "Target heap size of the next GC cycle in bytes", metricdata.UnitBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	memStats.lastGC, err = producer.createInt64GaugeEntry("process/last_gc", "Time the last garbage collection finished, as milliseconds since 1970 (the UNIX epoch)", metricdata.UnitMilliseconds)
```
rghetia, line 258: next_gc and last_gc appears to be similar but quite different. Perhaps the could be renamed to

next_gc_heap_size
last_gc_finished_timestamp

plugin/runmetrics/producer.go
```diff
@@ -219,6 +229,52 @@ func newMemStats(producer *producer) (*memStats, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	memStats.pauseTotalNs, err = producer.createInt64GaugeEntry("process/pause_total", "Cumulative milliseconds spent in GC stop-the-world pauses", metricdata.UnitMilliseconds)
+	if err != nil {
+		return nil, err
+	}
+
+	memStats.pauseTotalNs, err = producer.createInt64GaugeEntry("process/pause_total", "Cumulative milliseconds spent in GC stop-the-world pauses", metricdata.UnitMilliseconds)
```
rghetia, line 268: duplicate. Same as line 263.

rghetia: All cumulative values can use [AddInt64Cumulative](https://github.com/census-instrumentation/opencensus-go/blob/df6e2001952312404b06f5f6f03fcb4aec1648e5/metric/registry.go#L184)

this allows to compute rate on certain backends.

TrevinTeacutter: The problem with AddInt64Cumulative is that it expects an incremental value, which is not how this metric can be recorded. These would be similar to how `memMalloc` and `memFrees` work in this producer in that the runtime is incrementing for you and all that you can record is the total value.

TrevinTeacutter:
@googlebot I fixed it

plugin/runmetrics/producer.go
rghetia: In that case you could use [AddInt64DerivedCumulative](https://github.com/census-instrumentation/opencensus-go/blob/643eada29081047b355cfaa1ceb9bc307a10423c/metric/test/exporter_test.go#L68).

TrevinTeacutter: I finally did this, I also did this retroactively for the metrics that I believe are also cumulative in this runmetrics package since it seems weird to do this only for net new metrics.

rghetia: Sorry for the long delay. The problem with retroactively changing the type for existing metrics is that it will fail for some backends. For example, stackdriver will reject the same metric if the type is different. But on the other hand it is the right thing to do. We'll have add release note as breaking change for this plugin.
Alternatively, it could support both options. Provide a flag to choose the option.

TrevinTeacutter: Makes sense, I'll add a flag for this to ensure people can migrate at their own pace.

TrevinTeacutter: However, should this flag control only the ones necessary to stay passive or all cumulative metrics? I lean towards the latter but I can understand why someone would want the former.

TrevinTeacutter: Not really happy with this but honestly rather than have a bunch of `if` statements running around it made sense to have separate structures.

punya (approved):

punya:
@TrevinTeacutter would you mind updating this branch one more time, so that Github Actions can run CI? Thanks again for your patience!

TrevinTeacutter:
@punya np and done!

PR #1211, opened 2020-06-01
Delete views from measure ref when unregistering

ian-mi:
Currently measure refs will accumulate views that have been unregistered which can skew benchmarks when tests frequently reregister their views.

### DIFF
--- a/stats/view/worker_commands.go
+++ b/stats/view/worker_commands.go
@@ -103,7 +103,7 @@ func (cmd *unregisterFromViewReq) handleCommand(w *worker) {
 			// The collected data can be cleared.
 			vi.clearRows()
 		}
-		w.unregisterView(name)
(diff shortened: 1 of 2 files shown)


stats/view/worker_commands.go
```diff
@@ -104,6 +104,9 @@ func (cmd *unregisterFromViewReq) handleCommand(w *worker) {
 			vi.clearRows()
 		}
 		w.unregisterView(name)
+		if measure := w.measures[vi.view.Measure.Name()]; measure != nil {
+			delete(measure.views, vi)
+		}
```
rghetia, line 109: Just like reference to view is added in measure ref during tryRegisterView, removal should also be in unregisterView func.

ian-mi: Done.

PR #1212, opened 2020-06-04
Allow custom view.Meters to export metrics for other Resources

evankanderson:
Working with #1196, I discovered that `metricexport.Reader` collects metrics from all registered Producers, but doesn't provide a way to distinguish between different Meters. (With `RegisterExporter`, I can hand off a different exporter for each Meter and curry the Resource that way.)

This fills in the existing `metricdata.Resource` field when available, which looks like it should already work with Stackdriver. Prometheus seems to ignore the Resource field; I can send a PR for that if desired.

### DIFF
--- a/stats/view/worker_test.go
+++ b/stats/view/worker_test.go
@@ -18,9 +18,12 @@ package view
 import (
 	"context"
 	"errors"
+	"sort"
 	"sync"
 	"testing"
 	"time"
+
+	"go.opencensus.io/resource"
 
 	"go.opencensus.io/metric/metricdata"
 	"go.opencensus.io/metric/metricexport"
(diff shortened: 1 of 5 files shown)


jjzeng-seattle:
We need a minor fix for stackdriver though.  `resource` is overwritten by `ResourceByDescriptor`.  https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/metrics.go#L173

evankanderson:
I'm wondering if we can leverage `resource.MultiDetector` for this:

https://github.com/census-instrumentation/opencensus-go/blob/master/resource/resource.go#L146

evankanderson:
Alternatively, we could do the conversion on the recording side for Stackdriver in Knative, and retire the use of `ResourceByDescriptor` from Stackdriver.

stats/view/worker_test.go
```diff
@@ -619,3 +641,20 @@ func restart() {
 	defaultWorker = NewMeter().(*worker)
 	go defaultWorker.start()
 }
+
+// byTag implements sort.Interface for *metricdata.TimeSeries by Labels.
```
rghetia, line 645: nit: s/byTag/byLabel/

evankanderson: Whoops, done.

PR #1215, opened 2020-06-17
Initialize View Start Time During View Registration

ian-mi:
Fixes https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/issues/266

### DIFF
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -324,6 +324,7 @@ func (w *worker) tryRegisterView(v *View) (*viewInternal, error) {
 		return x, nil
 	}
 	w.views[vi.view.Name] = vi
+	w.startTimes[vi] = time.Now()
 	ref := w.getMeasureRef(vi.view.Measure.Name())
 	ref.views[vi] = struct{}{}
 	return vi, nil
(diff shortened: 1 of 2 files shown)


ian-mi:
@nilebox

stats/view/worker.go
```diff
@@ -333,20 +334,17 @@ func (w *worker) unregisterView(v *viewInternal) {
...
 	}
 }
 
-func (w *worker) reportView(v *viewInternal, now time.Time) {
+func (w *worker) reportView(v *viewInternal) {
 	if !v.isSubscribed() {
 		return
 	}
 	rows := v.collectedRows()
-	_, ok := w.startTimes[v]
```
nilebox, line 346: There is another place with the same code which can be deleted: https://github.com/census-instrumentation/opencensus-go/blob/1901b56b9515b0c34f5d25a5bce982dfc543d64b/stats/view/worker.go#L376-L379

stats/view/worker.go
```diff
@@ -333,20 +334,17 @@ func (w *worker) unregisterView(v *viewInternal) {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 	delete(w.views, v.view.Name)
+	delete(w.startTimes, v)
```
nilebox, line 337: 👍

stats/view/worker.go
ian-mi: Good catch!

Issue #1214, opened 2020-06-16
GRPC Plugin Reports server/client CompletedRpcs to Latency Measure


smhendrickson:
This can be found in master:
https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ocgrpc/client_metrics.go#L64
https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ocgrpc/server_metrics.go#L67

Smells like a typo? If so, I'm happy to provide a PR.

AIUI the process should be just adding a new stats var, and fixup tests:
https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ocgrpc/server_metrics.go#L25-L31

james-bebbington:
Yes, this definitely looks like a typo. Please go ahead and create a PR if you need this to be fixed.

smhendrickson:
Thanks James, I've sent a PR. we can go over the details there.

PR #1216, opened 2020-06-29
Implement server and client `completed_rpcs` (#1214)

smhendrickson:
The previous implementation appears to be a copy-paste error that directed data into
the latency metric, but never recorded data preserving the latency measure
accuracy. This cl implements the `completed_rpcs` measure by
incrementing by 1 for each `handleRPCEnd` call.

The measure is broken down by method and status.


smhendrickson:
@rghetia agreed. I've just checked my metrics and can confirm `completed_rpcs` is exported with the intended breakdowns. Not sure why I missed it and went digging before.

With that in mind I can either reduce this pr to an informative comment, or keep it as is. I find the PR more straightforward but it really depends how efficient you want the plugin to be. Up to you all.

RE: CLA not sure why this isn't working. My @google email should get his to pass. Will keeping looking.

rghetia:
> With that in mind I can either reduce this pr to an informative comment, or keep it as is. I find the PR more straightforward but it really depends how efficient you want the plugin to be. Up to you all.
> 

I would go with an informative comment.

smhendrickson:
https://github.com/census-instrumentation/opencensus-go/pull/1217 created for the comment.

Unsure if it is correct form to create a new PR for something like that, but thats what happens when I don't use github (or git) for years 🤷‍♂️

PR #1220, opened 2020-07-10
Record a Start Time Per Time Series within a View

ian-mi:
Currently all time series within a view will use a single start time recorded when the view is registered. Consequently if a time series is first recorded long after the view is registered the first point will be misattributed to a potentially very long time interval. Instead record a separate start time per time series within AggregationData that is recorded when data for that time series is first recorded. This start time is used when converting rows to TimeSeries and may also be used directly by stats exporters.

### DIFF
--- a/plugin/ocgrpc/client_stats_handler_test.go
+++ b/plugin/ocgrpc/client_stats_handler_test.go
@@ -320,14 +320,7 @@ func TestClientDefaultCollections(t *testing.T) {
 				continue
 			}
 			for i := range gotRows {
-				switch data := gotRows[i].Data.(type) {
-				case *view.CountData:
-					data.Start = time.Time{}
-				case *view.SumData:
-					data.Start = time.Time{}
-				case *view.DistributionData:
-					data.Start = time.Time{}
-				}
+				clearStart(gotRows[i].Data)
 			}
 
 			for _, gotRow := range gotRows {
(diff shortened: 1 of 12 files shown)


ian-mi:
@nilebox

plugin/ocgrpc/client_stats_handler_test.go
```diff
@@ -318,6 +319,16 @@ func TestClientDefaultCollections(t *testing.T) {
...
 				continue
 			}
+			for i := range gotRows {
+				switch data := gotRows[i].Data.(type) {
+				case *view.CountData:
+					data.Start = time.Time{}
+				case *view.SumData:
+					data.Start = time.Time{}
+				case *view.DistributionData:
+					data.Start = time.Time{}
```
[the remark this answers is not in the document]
ian-mi: This won't work because go must resolve data to a concrete type in order to reference the Start field.

stats/view/worker.go
```diff
@@ -371,15 +371,7 @@ func (w *worker) toMetric(v *viewInternal, now time.Time) *metricdata.Metric {
 		return nil
 	}
 
-	var startTime time.Time
-	if v.metricDescriptor.Type == metricdata.TypeGaugeInt64 ||
-		v.metricDescriptor.Type == metricdata.TypeGaugeFloat64 {
-		startTime = time.Time{}
-	} else {
-		startTime = w.startTimes[v]
```
nilebox, line 379: The only place we read from this map now is https://github.com/census-instrumentation/opencensus-go/blob/5fa069b99bc903d713add0295c7e0a55d34ae573/stats/view/worker.go#L350
Maybe rename it to `viewStartTimes` to make it more clear?

plugin/ocgrpc/client_stats_handler_test.go
```diff
@@ -318,6 +319,16 @@ func TestClientDefaultCollections(t *testing.T) {
 				t.Errorf("%q: RetrieveData(%q) = %v", tc.label, wantData.v().Name, err)
 				continue
 			}
+			for i := range gotRows {
+				switch data := gotRows[i].Data.(type) {
+				case *view.CountData:
+					data.Start = time.Time{}
```
nilebox, line 325: Can you please leave a comment about why do we need to overwrite the `Start`?
I believe it's needed to be able to compare "want" and "got" (otherwise they will be considered non-equal).

Also, it seems that you have more than one place doing this, so maybe extract it to an utility function `func ResetStart(data *view.CountData)`?

ian-mi: I wasn't sure what package to add such a test helper to so I just added the helper to each package that needed it.

ian-mi: I went ahead and moved this to aggregation_data.go in the view package. Let me know if this is not the right place to put it.

PR #1227, opened 2020-09-07
Convert 500 http status code to trace.status

OmarElGabry:
Is there any reason why we aren't converting http status code to a trace.Status? The `http.StatusInternalServerError` has a corresponding trace status code `trace.StatusCodeInternal`


OmarElGabry:
@googlebot I signed it!

PR #1228, opened 2020-10-01
Defer IDGenerator initialization until first use

jlebon:
Initializing the `IDGenerator` from `init()` means that any downstream
project which transitively imports this package somewhere in its
dependency tree will incur `getrandom()` syscalls it has no control over
at startup.

This causes problems for us in
[Ignition](https://github.com/coreos/ignition), where we're transitively
pulling in this package via cloud.google.com/go/storage. Ignition runs
very early during the boot process, which means that even though this
isn't using `GRND_RANDOM`, the `getrandom` syscall can block until the
entropy pool is ready. This is a real problem when running in VMs on
systems which don't provide a virtualized RNG device (such as VMware) or
which lack RDRAND.

I can't find a good reference for this, but I think in general it should
be considered good practice to avoid I/O like this in `init()` in favour
of a more lazy approach (or an explicit `Initialize()` function for
clients to call).

Otherwise, *every* program which pulls in the package will pay for it,
whether or not they intend to actually make use of the functionality
those syscalls are priming. (While it's not relevant here, another
advantage of not using `init()` for this is that I/O is fallible, and
`init()` semantics means errors can't be handled sanely.)

Let's rework things here so that we don't actually initialize the
`IDGenerator` fields until the first time it's used.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -206,6 +206,10 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
 	span.spanContext = parent
 
 	cfg := config.Load().(*Config)
+	if gen, ok := cfg.IDGenerator.(*defaultIDGenerator); ok {
+		// lazy initialization
+		gen.init()
+	}
 
 	if !hasParent {
 		span.spanContext.TraceID = cfg.IDGenerator.NewTraceID()
@@ -534,20 +538,9 @@ func (s *Span) String() string {
 var config atomic.Value // access atomically
 
 func init() {
-	gen := &defaultIDGenerator{}
-	// initialize traceID and spanID generators.
-	var rngSeed int64
-	for _, p := range []interface{}{
-		&rngSeed, &gen.traceIDAdd, &gen.nextSpanID, &gen.spanIDInc,
-	} {
-		binary.Read(crand.Reader, binary.LittleEndian, p)
-	}
-	gen.traceIDRand = rand.New(rand.NewSource(rngSeed))
-	gen.spanIDInc |= 1
-
 	config.Store(&Config{
 		DefaultSampler:             ProbabilitySampler(defaultSamplingProbability),
-		IDGenerator:                gen,
+		IDGenerator:                &defaultIDGenerator{},
 		MaxAttributesPerSpan:       DefaultMaxAttributesPerSpan,
 		MaxAnnotationEventsPerSpan: DefaultMaxAnnotationEventsPerSpan,
 		MaxMessageEventsPerSpan:    DefaultMaxMessageEventsPerSpan,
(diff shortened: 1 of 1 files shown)


jlebon:
Apologies for the drive-by contribution. For full disclosure, I only have a superficial understanding of this codebase and I'll admit that I haven't tested a trace with this patch. Unit tests do pass locally though.

cgwalters (approved):
I don't know this codebase either, but the change here looks sane to me.

(Aside: in Rust there is no `init()` equivalent, people use `lazy_static!` instead which naturally avoids this type of problem)

JAORMX (approved):

nilebox:
Thanks @jlebon for writing up the details of the issue and its impact!
Please note that the OpenCensus project is in maintenance at the moment, so we are only considering changes which address critical issues / use cases, and based on the description your PR matches that criteria.
 
Let me carefully review this change and get back to you in the next few days.

trace/trace.go
```diff
@@ -206,6 +206,21 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
 	span.spanContext = parent
 
 	cfg := config.Load().(*Config)
+	gen := cfg.IDGenerator.(*defaultIDGenerator)
```
nilebox, line 209: Please note that the `IDGenerator` can be overridden with a custom implementation, in which case this line would panic.

The easiest fix would be to perform a safe cast:
```go
gen, ok := cfg.IDGenerator.(*defaultIDGenerator)
if ok && !gen.randomIDsGenerated {
  ...
}
```

trace/trace.go
```diff
@@ -206,6 +206,21 @@ func startSpanInternal(name string, hasParent bool, parent SpanContext, remotePa
...
+
+	if !gen.randomIDsGenerated {
```
nilebox, line 211: It is possible to start multiple spans in parallel (in separate goroutines), so I think we need to introduce a lock here to make sure we initialize the generator exactly once?

I also suggest to extract this code into a function belonging to `defaultIDGenerator` and just invoke it from here, e.g.

```go
gen, ok := cfg.IDGenerator.(*defaultIDGenerator)
if ok {
  gen.initialize()
}
```
```go

type defaultIDGenerator struct {
  ...
  initOnce sync.Once
  ...
}

// initOnce initializes the generator on the first call
func (gen *defaultIDGenerator) initialize() {
  gen.initOnce.Do(func(){
    // initialize traceID and spanID generators.
    ...
  })
}
```

nilebox (changes requested):
The intended change makes sense to me, though I left a few suggestions to make it cleaner and safer.

trace/trace.go
nilebox: See updated code example in the other comment.

trace/trace.go
```diff
@@ -571,6 +575,8 @@ type defaultIDGenerator struct {
 
 	traceIDAdd  [2]uint64
 	traceIDRand *rand.Rand
+
+	randomIDsGenerated bool
```
nilebox, line 579: this field is redundant if you introduce `initOnce sync.Once` instead like I suggested above.

trace/trace.go
jlebon: Thanks, `sync.Once` fits nicely here. Updated with that!

(I wanted to give you `Co-authored-by` on this, but couldn't find an email address to use. Feel free to edit the commit message and add yourself if you'd like.)

nilebox (approved):

nilebox:
@jlebon you may switch to a pinned commit for dependency or wait until we make a new release for opencensus-go.

nilebox:
Release published: https://github.com/census-instrumentation/opencensus-go/releases/tag/v0.22.5

Issue #1232, opened 2020-10-08
Span created on client stats handler is not accessible from client interceptor


jcchavezs:
**Is your feature request related to a problem? Please describe.**
In the GRPC client, I am trying to access the span created by the stats handler from the interceptor but it is currently not possible as the context does not carry the span.

**Describe the solution you'd like**
I'd like this to be consistent with GRPC server where you can access the span in the context from the interceptor.

**Describe alternatives you've considered**
None.

**Additional context**

PR #1233, opened 2020-10-08
demostration on how context is lost between the stats handler and unary interceptor.

jcchavezs:
This PR is meant to demonstrate the behaviour in #1232


go.mod
```diff
@@ -2,14 +2,14 @@ module go.opencensus.io
 
 require (
```
jcchavezs, line 3: ignore this changes

nilebox:
Hi @jcchavezs, thanks for creating this PR.
As you may know, the OpenCensus project is currently in the maintenance mode, and has been superseded by [OpenTelemetry](http://opentelemetry.io/). As such, we only accept critical bugfixes with backward compatibility.

If the fix for the issue you reported here is trivial, we may consider accepting it.
Feel free to create a PR with a bugfix if you think that criteria applies here, and ping me on that PR.

PR #1235, opened 2020-10-23
Revert "Allow replacing trace SDK (#1234)"

tbpg:
This reverts commit fc3822be251c0e5c0d1b92bd51fabe401fe3c0bf.

This is causing a break in Google Cloud Functions, which depends on master when running in GOPATH mode.

See https://github.com/GoogleCloudPlatform/functions-framework-go/issues/56.

Here is the error:
```
src/github.com/cloudevents/sdk-go/v2/extensions/distributed_tracing_extension.go:161:3: cannot use span (type trace.Span) as type *trace.Span in return argument:
	*trace.Span is pointer to interface, not interface; Error ID: 1093f764
```


tbpg:
cc @nilebox @dashpole

rghetia (approved):

nilebox:
@tbpg It seems that you're using the HEAD revision for some reason? The breaking change wasn't released yet AFAICT.
Can't you pin the dependency tag to `v0.22.5`?

@dashpole if we can come up with a backward compatible solution that would be preferred over the reverted change.

tbpg:
In GOPATH mode, you need some extra dependency/vendor management to depend on a specific version. We don't have that at the moment.

PR #1237, opened 2020-10-23
Set up Github Actions for pull request validation

jsuereth:
- Set up a github actions PR validation workflow.
- Does *not* remove Travis CI (yet).

Example of it working: https://github.com/jsuereth/opencensus-go/pull/1/checks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest 
+    runs-on: ubuntu-20.04
     env:
       GO11MODULE: 'on'
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,23 @@
+name: Build
+
+on: 
+  pull_request:
+      branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest 
+    env:
+      GO11MODULE: 'on'
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          fetch-depth: 0 # we want all tags for version check.
+          lfs: true
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.11.0'
+      - name: Build and test
+        run: make install-tools && make travis-ci && go run internal/check/version.go


nilebox:
What is the ultimate plan here? Migrate all projects from Travis to GitHub Actions?
FWIW we use CircleCI in `opentelemetry-operations-go` and other OpenTelemetry / Google Cloud repos, so it would be useful to have a unified CI approach across all OpenCensus / OpenTelemetry projects.

jsuereth:
> What is the ultimate plan here? Migrate all projects from Travis to GitHub Actions?
> FWIW we use CircleCI in `opentelemetry-operations-go` and other OpenTelemetry / Google Cloud repos, so it would be useful to have a unified CI approach across all OpenCensus / OpenTelemetry projects.

Interesting.  I choose GHA because (most of) OpenTelemetry is using it and because I agree we should have a unified approach.   Primary goal of this CL was to figure out how expensive it would be to migrate (not very), secondary goal was to unify the projects in terms of technology if we're going to maintain both for while.  Given that OC is directly moving to OT, I thought unifying with OSS OT community would be better.   It appears that within OT, (JS/Go APIs, Collector + Specification) use CircleCI and the spot check I did (swift, c++, java, .NET, erlang, Python, JavaScript) use GHA.   I'm curious why there's such divergence between the repositories, do you know why to choose CircleCI over GHA?

.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
+name: Build
+
+on: 
+  pull_request:
+      branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
aabmass, line 10: should we pin it? Latest is changing https://github.com/actions/virtual-environments/issues/1816

```suggestion
    runs-on: ubuntu-20.04
```

aabmass (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
...
+      GO11MODULE: 'on'
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          fetch-depth: 0 # we want all tags for version check.
+          lfs: true
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.11.0'
```
dashpole, line 21: Lets use something more recent.  Maybe 1.14 or 1.15.  I think GO11MODULE isn't needed after that.

jsuereth: This is matching what we had in Travis-CI.  I'm not going to bump to 1.14/1.15 unless we're officially dropping support for older Go version here, which we're not.  This is just a migration.

PR #1238, opened 2020-10-26
Allow replacing trace SDK; attempt 2

dashpole:
Second attempt at #1234.

This time, it is completely backwards-compatible--including with the handling of nil `*Span`s.

To accomplish this, we introduce a new `Span` struct, which wraps a `SpanInterface` implementation provided by the tracer.

@nilebox


trace/trace_test.go
```diff
@@ -65,24 +65,25 @@ func checkChild(p SpanContext, c *Span) error {
 	if c == nil {
 		return fmt.Errorf("got nil child span, want non-nil")
 	}
-	if got, want := c.spanContext.TraceID, p.TraceID; got != want {
+	if got, want := c.SpanContext().TraceID, p.TraceID; got != want {
```
nilebox, line 68: Isn't this a breaking change still?

dashpole: spanContext is a non-exported field, so the removal of it from the `Span` struct isn't breaking for anyone using it as a library.

nilebox (approved):

trace/trace.go
```diff
@@ -109,13 +113,13 @@ type SpanContext struct {
 type contextKey struct{}
 
 // FromContext returns the Span stored in a context, or nil if there isn't one.
-func FromContext(ctx context.Context) *Span {
```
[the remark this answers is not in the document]
nilebox: Oh sorry I see that you've already done that in `trace_api.go` 👍

trace/trace_test.go
```diff
@@ -579,10 +580,10 @@ func TestSetSpanName(t *testing.T) {
 
 func TestSetSpanNameUnsampledSpan(t *testing.T) {
 	var nilSpanData *SpanData
-	span := startSpan(StartOptions{Sampler: NeverSample()})
-	span.SetName("NoopName")
-
-	if want, got := nilSpanData, span.data; want != got {
+	s := startSpan(StartOptions{Sampler: NeverSample()})
+	s.SetName("NoopName")
+	sp, _ := s.internal.(*span)
```
nilebox, line 585: here and in other places with similar change: we should panic rather than ignore cast result I think?
```suggestion
	sp := s.internal.(*span)
```

dashpole: Done for test files.

Do you just want to do this in the test file, or for non-test files as well?

nilebox: We just shouldn't ignore errors. I think you handle that properly in non-test files, i.e. check for `if ok { ... }`.
If not, either add proper handling or let it panic.

nilebox (approved):

trace/trace_test.go
dashpole: Got it.  I do handle cast failures properly elsewhere.

dashpole:
This should be good to go.

PR #1240, opened 2020-10-29
Provide accessor to the span implementation

dashpole:
This is needed if the SDK needs to cast to its underlying type.

Part of https://github.com/open-telemetry/opentelemetry-go/issues/93

@nilebox

### DIFF
--- a/trace/trace_api.go
+++ b/trace/trace_api.go
@@ -148,6 +148,11 @@ type Span struct {
 	internal SpanInterface
 }
 
+// Internal returns the underlying implementation of the Span
+func (s *Span) Internal() SpanInterface {
+	return s.internal
+}
+
 // IsRecordingEvents returns true if events are being recorded for this span.
 // Use this check to avoid computing expensive annotations when they will never
 // be used.


trace/trace_api.go
```diff
@@ -148,6 +148,11 @@ type Span struct {
 	internal SpanInterface
 }
 
+// Internal returns the underlying implementation of the Span
+func (s *Span) Internal() SpanInterface {
```
nilebox, line 152: Can you please explain why is this necessary?

dashpole:
This is needed when attempting to add an OpenCensus span to the OpenTelemetry context using `otel.ContextWithSpan(context.Context, otel.Span)`  In order to do that, we need to be able to get the underlying Span implementation.  The full implemnetation would look something like:


```golang
// This is the OpenCensus FromContext function, implemented using opentelemetry's SpanFromContext
func (o *otelTracer) FromContext(ctx context.Context) *octrace.Span {
	otSpan := otel.SpanFromContext(ctx)
	return // convert the opentelemetry span into the OC span implementation
}

// This is the OpenCensus NewContext function, implemented using opentelemetry's ContextWithSpan
func (o *otelTracer) NewContext(parent context.Context, s *octrace.Span) context.Context {
        // Below, I need to call s.Internal() in order to have an interface I can cast to my internal type.
        // For example, my internal type could wrap an opentelemetry Span, which I could use in ContextWithSpan
	if myspan, ok := s.Internal().(*myspanimplementation); ok {
		return otel.ContextWithSpan(parent, // get an opentelemetry Span from myspan)
	}
	// The user must have created the octrace Span using a different tracer, and we don't know how to store it.
	return parent
}
```

trace/trace_api.go
dashpole: This is needed when attempting to add an OpenCensus span to the OpenTelemetry context using `otel.ContextWithSpan(context.Context, otel.Span)`  In order to do that, we need to be able to get the underlying Span implementation.  The full implemnetation would look something like:


```golang
// This is the OpenCensus FromContext function, implemented using opentelemetry's SpanFromContext
func (o *otelTracer) FromContext(ctx context.Context) *octrace.Span {
	otSpan := otel.SpanFromContext(ctx)
	return // convert the opentelemetry span into the OC span implementation
}

// This is the OpenCensus NewContext function, implemented using opentelemetry's ContextWithSpan
func (o *otelTracer) NewContext(parent context.Context, s *octrace.Span) context.Context {
        // Below, I need to call s.Internal() in order to have an interface I can cast to my internal type.
        // For example, my internal type could wrap an opentelemetry Span, which I could use in ContextWithSpan
	if myspan, ok := s.Internal().(*myspanimplementation); ok {
		return otel.ContextWithSpan(parent, // get an opentelemetry Span from myspan)
	}
	// The user must have created the octrace Span using a different tracer, and we don't know how to store it.
	return parent
}
```

nilebox: Oh I see that makes sense.

nilebox (approved):

PR #1248, opened 2020-12-28
Adds an exported function to flush interval reader

quentin-cha:
This is a quick fix (bit hacky) to get me closer to fixing https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/issues/43,
by providing an exported function on `IntervalReader` to flush its content.

It doesn't change any existing behavior of `IntervalReader`, it only adds a way for the owner of `IntervalReader` to flush its content when needed.

### DIFF
--- a/metric/metricexport/reader.go
+++ b/metric/metricexport/reader.go
@@ -130,7 +130,7 @@ func (ir *IntervalReader) Start() error {
 		reportingInterval = ir.ReportingInterval
 	}
 
-	if ir.done != nil {
+	if ir.quit != nil {
 		return errAlreadyStarted
 	}
 	ir.timer = time.NewTicker(reportingInterval)
@@ -170,6 +170,19 @@ func (ir *IntervalReader) Stop() {
 	close(ir.quit)
 	close(ir.done)
 	ir.quit = nil
+}
+
+// Flush flushes the metrics if IntervalReader is stopped, otherwise no-op.
+func (ir *IntervalReader) Flush() {
+	ir.mu.Lock()
+	defer ir.mu.Unlock()
+
+	// No-op if IntervalReader is not stopped
+	if ir.quit != nil {
+		return
+	}
+
+	ir.reader.ReadAndExport(ir.exporter)
 }
 
 // ReadAndExport reads metrics from all producer registered with
--- a/metric/metricexport/reader_test.go
+++ b/metric/metricexport/reader_test.go
@@ -117,6 +117,69 @@ func TestManualReadForIntervalReader(t *testing.T) {
 	resetExporter(exporter1)
 }
 
+func TestFlushNoOpForIntervalReader(t *testing.T) {
+	ir1 = createAndStart(exporter1, duration1, t)
+
+	gaugeEntry.Set(1)
+
+	// since IR is not stopped, flush does nothing
+	ir1.Flush()
+
+	// expect no data points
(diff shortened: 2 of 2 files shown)


metric/metricexport/reader.go
```diff
@@ -172,6 +172,12 @@ func (ir *IntervalReader) Stop() {
 	ir.quit = nil
 }
 
+// Flush flushes the metrics by invoking ReadAndExport immediately.
+// The reader itself is not stopped and the timer is not reset.
+func (ir *IntervalReader) Flush() {
+	ir.reader.ReadAndExport(ir.exporter)
```
dashpole, line 178: do you know if this is safe to call concurrently?

quentin-cha: Thank you for reviewing this!

My bad. I have added some unit tests. They are also helpful in demonstrating the behavior of the code.
It seems that multiple concurrent calls will just export the metric multiple times. 
I'm not sure multiple calls to `Flush()` concurrent or not is the intended use case. I think this should be called once before shutdown. If it helps, I can make this into a `sync.Once` call.

dashpole: It sounds like the correct usage would always be to call `Flush` just after `Stop` to ensure they don't race, and to ensure metrics get sent:
```golang
ir.Start()
// do work
ir.Stop()
ir.Flush()
```
That way, you ensure you aren't sending anything else, and you empty the queue before shutting down (although I think it can still race and miss something?).  Normally, I would say we should combine Flush and Stop, but as the repo is in maintenance mode, I don't want to do anything that could cause regressions.

I think we should be able to use the quit and done channels to verify that it is being called after Stop(), or that the exporter has not been `Start'ed

quentin-cha: Made it only flush once after IR is stopped. Please let me know if this is acceptable. Thanks!

dashpole: I don't think we need to ensure flush is only called once--just that it isn't called with a started interval reader.  In theory, it should be possible to:
```golang
ir.Start()
ir.Stop()
ir.Flush()
ir.Start()
// ...
```

quentin-cha: That is a very good point, thank you!
I had to change the `Start()` function, I think it's a bug to check `ir.done` rather than `ir.quit` to see if IR is started.

dashpole: It looks like this still limits this to being called once.  Can we remove that bit?

quentin-cha: My bad, I misunderstood you. I thought you meant multiple calls with IR restart. I think this is fixed now. Thanks again.

dashpole (approved):

metric/metricexport/reader.go
```diff
@@ -172,6 +172,20 @@ func (ir *IntervalReader) Stop() {
 	ir.quit = nil
 }
 
+// Flush flushes the metrics if IntervalReader is stopped, otherwise no-op.
+// Additional call to Flush are no-ops.
```
dashpole, line 176: remove this comment, as it no longer applies

dashpole:
whoops.  one more comment, then i'll merge

metric/metricexport/reader.go
quentin-cha: done

quentin-cha:
Thank you so much your help!

Issue #1265, opened 2021-09-14
Improve performance of stats recording


howardjohn:
**Is your feature request related to a problem? Please describe.**
Yes. Usage of opencensus-go to record metrics has a substantial overhead. In real world application, we have seen OC accounting for well over 10% of our memory allocations (and we generate GBs of protobufs per minute - so it should be negligible). This has led us to doing things we really shouldn't have to think about, like adding a caching layer on top of the library.

**Describe the solution you'd like**
Improve performance of the library; in particular memory allocations

**Describe alternatives you've considered**
Adding a caching layer above the library, using a different library.

**Additional context**

I wrote some benchmarks to compare to the prometheus client. There are two variants, one with a precompute label/tag and one computed in the loops:
```go
func BenchmarkMetrics(b *testing.B) {
	b.Run("oc", func(b *testing.B) {
		mLineLengths := stats.Float64("test", "my-benchmark", stats.UnitDimensionless)
		key := tag.MustNewKey("key")
		v := &view.View{
			Measure:     mLineLengths,
			TagKeys:     []tag.Key{key},
			Aggregation: view.Sum(),
		}
		if err := view.Register(v); err != nil {
			b.Fatal(err)
		}

		for n := 0; n < b.N; n++ {
			allTags := []tag.Mutator{tag.Upsert(key, "val")}
			if err := stats.RecordWithTags(context.Background(), allTags, mLineLengths.M(1)); err != nil {
				b.Fatal(err)
			}
		}
	})
	b.Run("oc-fixed", func(b *testing.B) {
		mLineLengths := stats.Float64("test", "my-benchmark", stats.UnitDimensionless)
		key := tag.MustNewKey("key")
		v := &view.View{
			Measure:     mLineLengths,
			TagKeys:     []tag.Key{key},
			Aggregation: view.Sum(),
		}
		if err := view.Register(v); err != nil {
			b.Fatal(err)
		}

		allTags := []tag.Mutator{tag.Upsert(key, "val")}
		for n := 0; n < b.N; n++ {
			if err := stats.RecordWithTags(context.Background(), allTags, mLineLengths.M(1)); err != nil {
				b.Fatal(err)
			}
		}
	})
	b.Run("prom", func(b *testing.B) {
		g := prometheus.NewGaugeVec(prometheus.GaugeOpts{
			Namespace: "tests",
			Name:      "test",
		}, []string{"key"})
		prometheus.Register(g)
[...]

PR #1266, opened 2021-09-14
Register metrics in benchmark

howardjohn:
Currently, the metric we record is not registered. This hits the
fast-path code of not actually recording the metric, so we miss out on
detecting any performance to that main code path.

This registers the metrics so we actually trigger `record`.

Related: https://github.com/census-instrumentation/opencensus-go/issues/1265


dashpole (approved):

dashpole:
I went ahead and merged https://github.com/census-instrumentation/opencensus-go/pull/1267, which includes this change

PR #1267, opened 2021-09-14
Optimize `Record()` to avoid extra allocations 

howardjohn:
Currently, `Record()` re-uses code with `RecordWithOptions`. This always creates
allocations for createRecordOption, which is not needed in this case - we only
have measurements and not generic options.

With a little code duplication, we can reduce these allocations.

```
name                    old time/op    new time/op    delta
Record0-6                 92.2ns ± 9%     1.7ns ± 4%   -98.11%  (p=0.008 n=5+5)
Record1-6                  665ns ± 5%     634ns ± 6%    -4.57%  (p=0.095 n=5+5)
Record8-6                 1.24µs ± 5%    1.21µs ± 5%    -2.18%  (p=0.206 n=5+5)
Record8_WithRecorder-6     796ns ± 5%     777ns ± 5%    -2.45%  (p=0.222 n=5+5)
Record8_Parallel-6        1.21µs ± 2%    1.26µs ±24%      ~     (p=0.690 n=5+5)
Record8_8Tags-6           1.23µs ± 4%    1.23µs ± 2%      ~     (p=0.968 n=5+5)

name                    old alloc/op   new alloc/op   delta
Record0-6                  80.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)
Record1-6                   200B ± 0%      120B ± 0%   -40.00%  (p=0.008 n=5+5)
Record8-6                   424B ± 0%      344B ± 0%   -18.87%  (p=0.008 n=5+5)
Record8_WithRecorder-6      424B ± 0%      424B ± 0%      ~     (all equal)
Record8_Parallel-6          424B ± 0%      344B ± 0%   -18.87%  (p=0.008 n=5+5)
Record8_8Tags-6             424B ± 0%      344B ± 0%   -18.87%  (p=0.008 n=5+5)

name                    old allocs/op  new allocs/op  delta
Record0-6                   1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
Record1-6                   4.00 ± 0%      3.00 ± 0%   -25.00%  (p=0.008 n=5+5)
Record8-6                   4.00 ± 0%      3.00 ± 0%   -25.00%  (p=0.008 n=5+5)
Record8_WithRecorder-6      4.00 ± 0%      4.00 ± 0%      ~     (all equal)
Record8_Parallel-6          4.00 ± 0%      3.00 ± 0%   -25.00%  (p=0.008 n=5+5)
[...]

PR #1268, opened 2021-09-14
 Remove `convTslice` calls in Record() 

howardjohn:
This is built upon #1267; that
one should likely merge first. I split this out as it has a small public
API change (to work around circular imports) to avoid issues on the
first PR.

Benchmark relative to #1267:
```
name                    old time/op    new time/op    delta
Record0-6                 1.74ns ± 4%    1.79ns ± 2%   +2.85%  (p=0.238 n=5+5)
Record1-6                  634ns ± 6%     542ns ± 9%  -14.55%  (p=0.008 n=5+5)
Record8-6                 1.21µs ± 5%    1.23µs ± 2%   +1.97%  (p=0.254 n=5+5)
Record8_WithRecorder-6     777ns ± 5%     792ns ± 5%   +1.97%  (p=0.421 n=5+5)
Record8_Parallel-6        1.26µs ±24%    1.22µs ± 2%     ~     (p=0.690 n=5+5)
Record8_8Tags-6           1.23µs ± 2%    1.25µs ± 3%     ~     (p=0.651 n=5+5)

name                    old alloc/op   new alloc/op   delta
Record0-6                  0.00B          0.00B          ~     (all equal)
Record1-6                   120B ± 0%       96B ± 0%  -20.00%  (p=0.008 n=5+5)
Record8-6                   344B ± 0%      320B ± 0%   -6.98%  (p=0.008 n=5+5)
Record8_WithRecorder-6      424B ± 0%      424B ± 0%     ~     (all equal)
Record8_Parallel-6          344B ± 0%      320B ± 0%   -6.98%  (p=0.008 n=5+5)
Record8_8Tags-6             344B ± 0%      320B ± 0%   -6.98%  (p=0.008 n=5+5)

name                    old allocs/op  new allocs/op  delta
Record0-6                   0.00           0.00          ~     (all equal)
Record1-6                   3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.008 n=5+5)
Record8-6                   3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.008 n=5+5)
Record8_WithRecorder-6      4.00 ± 0%      4.00 ± 0%     ~     (all equal)
Record8_Parallel-6          3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.008 n=5+5)
Record8_8Tags-6             3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.008 n=5+5)
```

For https://github.com/census-instrumentation/opencensus-go/issues/1265


stats/record.go
```diff
@@ -94,7 +97,7 @@ func Record(ctx context.Context, ms ...Measurement) {
 	if len(ms) == 0 {
 		return
 	}
-	recorder := internal.DefaultRecorder
+	recorder := InternalMeasurementRecorder
```
dashpole, line 100: Can you help me double-check my reasoning here?  In theory, this could break someone who was overriding internal.DefaultRecorder, but that isn't possible, since it is in an internal package.

As long as this is backwards-compatible, I think we can accept the change.

howardjohn: I agree with your assessment; its backwards compatible because its `internal` and thus cannot be used. However, it does add a new public field that a user could start messing with when they probably shouldn't. I could not find a way to avoid this without introducing an import loop unfortunately.

howardjohn: I refactored it to avoid leaking into the public API, and its still backwards compatible since they could not override `internal.DefaultRecorder` as mentioned above.

stats/internal/record.go
```diff
@@ -21,5 +21,11 @@ import (
 // DefaultRecorder will be called for each Record call.
 var DefaultRecorder func(tags *tag.Map, measurement interface{}, attachments map[string]interface{})
 
+// MeasurementRecorder will be called for each Record call. This is the same as DefaultRecorder but
+// avoids interface{} conversion.
+// This will be a func(tags *tag.Map, measurement []Measurement, attachments map[string]interface{}) type,
+// but is interface{} here to avoid import loops
+var MeasurementRecorder interface{}
```
dashpole, line 28: If someone overrode this with a `func(tags *tag.Map, measurement []Measurement, attachments map[string]interface{})`, would `recorder := internal.MeasurementRecorder.(measurementRecorder)` accept it?

At first glance, this seems unnecessary, since this is already in an internal package

howardjohn: Not 100% sure what you mean:

if you mean modified this L28 to use that instead of `interface{}` it would not compile (import loop)

If you mean set the value; we do (`internal.MeasurementRecorder = recordMeasurement`) but it can only be done in this package since its in `internal`, so a user couldn't do it?

dashpole: Oh, I see the import loop now.  This approach is probably the best we can do.  Are the benchmarks in the description updated for the latest attempt?

dashpole (approved):
I'm going to hold off merging this for a day or two to make sure others have a chance to comment.

stats/internal/record.go
howardjohn: I just updated it, but its the same (besides recording noise) as the initial iteration.

howardjohn:
Is this good to go?

dashpole:
cc @punya 
for a second opinion

PR #1270, opened 2021-09-14
Fix formatting to pass with go1.17

howardjohn:
The CI job currently uses an unpinned image. This was recently updated
to golang 1.17, causing all PRs to fail. This fixes those build failures


dashpole (approved):
Thanks.

Issue #1262, opened 2021-08-14
stats: defaultWorker can not be stopped


cornelk:
Currently there is no way to stop the `defaultWorker` goroutine in the `go.opencensus.io/stats/view/` package.
This breaks adding a Goroutine leak detector to unit tests in packages that import the view package as it will always fail:

```
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 21 in state select, with go.opencensus.io/stats/view.(*worker).start on top of the stack:
goroutine 21 [select]:
go.opencensus.io/stats/view.(*worker).start(0xc000272000)
	/home/user/go/pkg/mod/go.opencensus.io@v0.23.0/stats/view/worker.go:276 +0x157
created by go.opencensus.io/stats/view.init.0
	/home/user/go/pkg/mod/go.opencensus.io@v0.23.0/stats/view/worker.go:34 +0x87
```

markuspeloquin:
This issue has been closed even though the fix isn't tagged for any release.

And just FYI, this fix cannot work with VerifyTestMain since you cannot trigger a cleanup before checking for leaks. You must basically reimplement it:

```go
func TestMain(m *testing.M) {
    exitcode := 0
    defer func() { os.Exit(exitcode) }

    exitcode = m.Run()
    view.Stop()
    if err := goleak.Find(); err != nil {
        fmt.Fprintf(os.Stderr, "goleak: Errors on successful test run: %v\n", err)
        exitcode = 1
    }
}
```

IMO a package `init()` shouldn't be starting threads.

PR #1272, opened 2021-10-17
fix: Add function to stop the defaultWorker

kylecarbs:
Fixes https://github.com/census-instrumentation/opencensus-go/issues/1262.

### DIFF
--- a/stats/view/worker.go
+++ b/stats/view/worker.go
@@ -221,6 +221,11 @@ func SetReportingPeriod(d time.Duration) {
 	defaultWorker.SetReportingPeriod(d)
 }
 
+// Stop stops the default worker.
+func Stop() {
+	defaultWorker.Stop()
+}
+
 // SetReportingPeriod sets the interval between reporting aggregated views in
 // the program. If duration is less than or equal to zero, it enables the
 // default behavior.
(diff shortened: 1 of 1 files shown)


kylecarbs:
@googlebot I signed it!

kylecarbs:
@dashpole how can I get this merged?

stats/view/worker.go
```diff
@@ -290,8 +295,11 @@ func (w *worker) start() {
 func (w *worker) Stop() {
 	prodMgr := metricproducer.GlobalManager()
 	prodMgr.DeleteProducer(w)
-
-	w.quit <- true
+	select {
```
dashpole, line 298: Sorry, i've lost context on this.  Why are these (this block, and L289) changes required?  I believe this means you can't call start again after stop, since the channel is closed.

kylecarbs: That's correct. We can refactor to make it so, but the objective was to enable calls to `Stop()` multiple times.

dashpole: I think we will need to do that.  We can't make breaking changes at this point in the project.

kylecarbs: Actually, it already didn't work to `Start()` after `Stop()`. `w.c` was closed on quit.

markuspeloquin:
Can this fix get a release tag?

dashpole:
It is part of [v0.24.0](https://github.com/census-instrumentation/opencensus-go/releases/tag/v0.24.0)

PR #1273, opened 2021-11-16
fix: zpages default handler add trailing slash

actliboy:
the path.Join call path.Clean will clean trailing slash


punya:
Thanks for your contribution. OpenCensus is in maintenance mode so we are only prioritizing critical bugfixes. Please reopen if you disagree with our assessment of severity.

PR #1276, opened 2022-03-31
Passing capacity to make() in place of length. 

fayfaychan:
The second argument is length and not capacity.


punya (approved):

punya (changes requested):
Sorry, I didn't notice that this tried to change maps in addition to slices. Please fix the compilation errors.

Dombo:
I'm fairly confident this improves the memory footprint/fixes a memory leak happening to any consuming package relying on opencensus-go for generating traces.

PR #1279, opened 2022-07-27
Enforce tag equivalence in `view.same()`

rubensf:
Otherwise, the lack of error may lead to developers wasting a lot
of time trying to figure out why do their labels not show up.

### DIFF
--- a/stats/view/view.go
+++ b/stats/view/view.go
@@ -65,6 +65,7 @@ func (v *View) same(other *View) bool {
 		return false
 	}
 	return reflect.DeepEqual(v.Aggregation, other.Aggregation) &&
+		reflect.DeepEqual(v.TagKeys, other.Aggregation) &&
 		v.Measure.Name() == other.Measure.Name()
 }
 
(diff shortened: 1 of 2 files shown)


dashpole:
He @rubensf!  Right now we are only accepting security fixes or critical bug fixes.  Can you say more about the issue this fixes?

rubensf:
Ah, I didn't notice this repo was being phased out. The problem here is that the check for two views being the same, in my semantic understanding, incorrect - it was returning that two views are the same even though they had different keys. This is an issue if you eg use [`zpages`](https://github.com/census-instrumentation/opencensus-go/blob/master/zpages/zpages.go#L46) which registers some views at `init`. You can technically Unregister them and Register new ones, but if you're not very familiar with your codebase you may spend a few hours (as I did) trying to find out why your views with different tags are not showing up on your views... :)

That'd def be a behaviour change. I'm not sure if you'd count it as a critical bug fix though so feel free to close it.

dashpole:
Thanks for your understanding.  If others hit this issue we can reconsider it, but i'll close it for now.  Hopefully your workaround will save them some time.

Issue #1282, opened 2022-10-25
Testify needs updating due to DOS and Null ptr dereference - plz update testify to >= 1.7.5


rohanthewiz:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
latest (master as of March 2022)

### What version of Go are you using?
1.19.1

### What did you do?
If possible, provide a recipe for reproducing the error.
Issue identified by Snyk
[CWE-400](https://cwe.mitre.org/data/definitions/400.html)
[CWE-476](https://cwe.mitre.org/data/definitions/476.html)
> go.opencensus.io@v0.23.0 › github.com/stretchr/testify@v1.6.1 › gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c

### What did you expect to see?
No Denial of Service, no Null pointer deference

### What did you see instead?
CWE-400 and CWE-476

### Additional context
testify@v1.6.1 includes yaml.v3@3.0.0. Issue is fixed in testify@1.7.5 which includes yaml.v3@3.0.1

PR #1283, opened 2022-10-31
Add started RPC metric for client and server side

zasweq:

PR #1284, opened 2022-11-01
Fix CI, and update testify

dashpole:
Address https://github.com/census-instrumentation/opencensus-go/pull/1283#issuecomment-1298899577

Also, update testify, and run `go fmt`.

Fixes https://github.com/census-instrumentation/opencensus-go/issues/1282


punya:
The changes in this PR seem safe, but are we concerned that this might make it easy for us to break pre-1.18 users in future PRs?

dashpole:
> The changes in this PR seem safe, but are we concerned that this might make it easy for us to break pre-1.18 users in future PRs?

We don't currently test multiple versions, so we might be breaking any version other than the one we are testing.  1.18 seemed reasonable, since it is 1 behind the latest go release.

punya (dismissed):

punya:
OpenCensus is meant to be in maintenance mode, so I would expect the minimum required versions of its dependencies to remain unchanged unless there's a high priority bugfix or security vulnerability. This should apply to the Go version as well.

I would support
* updating the documentation to handle recent versions of Go correct (`go install` vs `go get`)
* narrowly updating the testify dependency to fix the problem

I would want OpenCensus-Go to keep building and testing against an *old* version of Go, and relying on the backwards compatibility of the Go language and stdlib for newer versions.

Happy to discuss in person if I'm misunderstanding the purpose of this PR.

dashpole:
@punya 
done

punya (approved):

Issue #1191, opened 2020-01-15
goroutine leak detected 


adamdecaf:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?

`v0.22.0`

### What version of Go are you using?

`go1.13.6`

### What did you do?

I was running the new [goleak](https://github.com/uber-go/goleak) tool on my codebase and ran into a leak coming from OpenCensus.

```
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 37 in state select, with go.opencensus.io/stats/view.(*worker).start on top of the stack:
goroutine 37 [select]:
go.opencensus.io/stats/view.(*worker).start(0xc000143310)
	/Users/adam/code/pkg/mod/go.opencensus.io@v0.22.0/stats/view/worker.go:154 +0x100
created by go.opencensus.io/stats/view.init.0
	/Users/adam/code/pkg/mod/go.opencensus.io@v0.22.0/stats/view/worker.go:32 +0x57
]
```

### What did you expect to see?

I expected no goroutine leaks in libraries I depend on. 

### What did you see instead?

A crash from goleaks. 

### Additional context

I'm pulling this library in via [gocloud.dev/secrets](https://godoc.org/gocloud.dev/secrets) `v0.17.0`, but that doesn't matter here since `v0.22.0` of OpenCensus is used in both. 

This appears to be coming from internal global state of a default recorder being set.

https://github.com/census-instrumentation/opencensus-go/blob/v0.22.0/stats/view/worker.go#L30-L34

Also, there doesn't seem to be a way to retrieve this reporter to shutdown on application shutdown.

https://github.com/census-instrumentation/opencensus-go/search?q=DefaultRecorder&unscoped_q=DefaultRecorder

regularcoder:
Here is a workaround that I used:
	ignoreOpenCensus := goleak.IgnoreTopFunction("go.opencensus.io/stats/view.(*worker).start")
	defer goleak.VerifyNone(suite.T(), ignoreOpenCensus)

brandur:
Just wanted to bump one and add a little context — a lot of us are running goleak in our tests because it's easy to make mistakes around leaking goroutines in Go, and once you have enough code written in a project, you probably are somewhere, and goleak is a great way to hedge against that and protect yourself against bad production problems before they happen.

[...]

PR #1287, opened 2022-12-11
Lazily initialize the `view` package's default worker

brandur:
See discussion in #1191 for impetus, but in general, it's not
particularly good practice to start up goroutines in package `init`
functions, with one reason being that it means that they're started for
projects that don't even make use of the package, and maybe just have it
as a transitive dependency.

So here, remove the `view` package's default worker in favor of a new
function that'll lazily initialize one when any package-level functions
are called that use the default worker. We use a `sync.Once` that has an
optimized short-circuit path in case the work's already been done, so
new overhead should be negligible, and registering/unregistering views
is probably not a particularly common operation anyway.

This change is backwards compatible, with use of the package's API
staying exactly the same. Test coverage seems to be pretty good, with a
function that previously reset the default worker between test cases
which I've modified to instead reset the default worker back to its
uninitialized state, and thereby verify that the test cases can
initialize it if necessary.

I thought about trying to deinitialize the default worker in case its
last view is unregistered, but decided against it because (1) it's more
invasive, (2) it's not likely to actually help anyone in practice as
views probably stay registered for a program's entire duration anwyay,
and (3) its interaction with the existing package-level `Stop` function
would take some thinking through.

### DIFF
--- a/stats/view/export.go
+++ b/stats/view/export.go
@@ -36,10 +36,12 @@ type Exporter interface {
 //
 // Binaries can register exporters, libraries shouldn't register exporters.
 func RegisterExporter(e Exporter) {
+	maybeInitDefaultWorker()
 	defaultWorker.RegisterExporter(e)
 }
 
 // UnregisterExporter unregisters an exporter.
 func UnregisterExporter(e Exporter) {
+	maybeInitDefaultWorker()
 	defaultWorker.UnregisterExporter(e)
 }
--- a/stats/view/export_test.go
+++ b/stats/view/export_test.go
@@ -1,0 +1,24 @@
+package view
+
+import "testing"
+
+// Exporting functionality is tested in `worker_test.go`. These are a trivial
+// set of tests to make sure that these package-level exports will correctly
+// initialize defaultWorker if necessary.
+func TestRegisterExporter(t *testing.T) {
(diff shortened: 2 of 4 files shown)


brandur:
I noticed there seems to be a pre-existing intermittent test problem in the package which can be repro'ed on `master` with something like:

```
$ go test ./stats/view -run Test_Worker_MultiExport -count=20
--- FAIL: Test_Worker_MultiExport (0.00s)
    worker_test.go:217: Mismatched value (want 2, got 1) for &{[{a true}] [{2022-12-11 13:06:06.331908 -0800 PST m=+0.003086001 1}] 2022-12-11 13:06:06.331901 -0800 PST m=+0.003079585} in "/VF1"
    worker_test.go:222: Mismatched value (want 7.500000, got 2.000000) for &{[] [{2022-12-11 13:06:06.331908 -0800 PST m=+0.003086001 2}] 2022-12-11 13:06:06.331901 -0800 PST m=+0.003079585} in "/VF2"
```

I didn't take a stab at it though because I feel like someone else would be able to fix this a lot quicker.

dashpole:
Fixing presubmit: https://github.com/census-instrumentation/opencensus-go/pull/1288

